PR #10, opened 2016-02-18
Better handling of .^ in ames and ++mink

cgyarvin:


juped:
thx 4 lost fix :+1:

Yours,
r

PR #14, opened 2016-02-20
scanf was dependant on (cook :_(foo |=(...)))

ohAitch:
Another victim rescued from the dicey pits of electroplating. @cgyarvin, is there a more canonical translation?

### DIFF
--- a/arvo/zuse.hoon
+++ b/arvo/zuse.hoon
@@ -1024,7 +1024,7 @@
   (scan a (parsf b))
 ++  parsf                                              ::  make parser from:
   |^  |*  a/(pole _;/(*{$^(rule tape)}))            ::  ;"chars{rule}chars"
-      %-  cook  :_  (bill (norm a))
+      =-  (cook - (bill (norm a)))
       |*  (list)
       ?~  +<  ~
       ?~  t  i


juped:
This is not a much better idiom, imo - both "terrible idioms" expose an error in the library. The real problem is that cook is backwards - just like hoon doesn't have (map f x) but (turn x f) because f is pretty much always a heavier twig, so should things like cook be (cook x f). (And the idiom, though not my favorite, should still work... there's always more than one way to do things)

ohAitch:
The problem with that, as with turn, is it's very much not always: I
frequently run into e.g. (turn (sort (~(tap by foobar)) lor) tail), and
similarly I think most uses of cook /are/ of the (cook sa (more com ;~(pfix
foo bar))) variety. And frequently the actual situation is "both twigs are
heavy", in which case you definitely want the the cookee to be the one
split off into another arm(which is in fact much done): it is
disproportionately likely to be "housekeeping", of less relevance to
understanding the code than the parsers themselves are. After all, if you
remove all cooks, you would still know /what/ was being parsed by the type
signature(which would no longer compile), and the /grammar/ it was being
parsed with, which would be unchanged.

PR #22, opened 2016-02-23
Change ~$(a ~) to ~$(%a ~).

cgyarvin:


ohAitch:
Alternately just stick an arbitrary twig in there, it can't _hurt_

juped:
# *REJECTED*
Per my new policy of loudly rejecting things instead of quietly correcting them, this is loudly rejected for violating public decency standards.
I'm morbidly curious about what exactly you *do* that generates this stuff. Will accept a patch or pull-request of the changes in commit 440ebc4 on its own.

ohAitch:
+1, I had like noticed that but didn't comment, probably wrongly so

ohAitch:
And by "can't hurt" I mean "I ended up using ~> %live.(rap 3 %name- data ~) in one place, which should probably just be expressible as ~$  (rap 3 %name- data ~)"
That would be a ++twig type change tho

juped:
Merged in d690fa4 with https://lwn.net/Articles/328438/ compliance

PR #25, opened 2016-02-24
improve ford performance

ohAitch:
- results of dynamically executed code now cached(renderers) like it used to be (hook files)
- renderers also no longer parsed per resource path, with `/: %/foo` syntax handled by passing through "at current path, /foo" as a literal to be bound to the resource-specific path at compilation.
  + this in turn means that their compilation is largely cached, as stack trace information is no longer (mistakenly) the resource path. Fixing urbit/urbit#697.
  + tree comments have been moved to a separate renderer, isolating their /: usage

Additionally,
- jet hints throughout ford for aid in profiling
- various tweaks to ++pi-tell to make the profiler output format more readable

### DIFF
--- a/arvo/ford.hoon
+++ b/arvo/ford.hoon
@@ -262,6 +262,7 @@
   ::
   ++  zo
     ~%  %ford-z  ..is  ~
+    :: =|  dyv/@
     |_  {num/@ud task}
     ++  abet  %_(..zo q.tad.bay (~(put by q.tad.bay) num +<+))
     ++  amok  
@@ -1137,9 +1138,9 @@
       :: =+  ^=  pre
       ::     ?+  -.kas  -.kas
       ::       ^  %cell
-      :: ::      %boil  [-.kas p.kas (tope q.kas)]
-      ::       %bake  [-.kas p.kas (tope q.kas)]
-      ::       %core  [-.kas (tope p.kas)]
+      :: ::      $boil  [-.kas p.kas (tope q.kas)]
+      ::       $bake  [-.kas p.kas (tope q.kas)]
+      ::       $core  [-.kas (tope p.kas)]
       ::     ==
       :: ~&  [dyv `term`(cat 3 %make (fil 3 dyv ' ')) pre]
       :: =-  ~&  [dyv `term`(cat 3 %made (fil 3 dyv ' ')) pre]  -
--- a/arvo/ford.hoon
+++ b/arvo/ford.hoon
@@ -55,7 +55,7 @@
       dep/(set beam)                                    ::  dependencies
   ==                                                    ::
 ++  calx                                                ::  concrete cache line
-  $%  {$hood p/calm q/(pair beam cage) r/hood}          ::  compile
+  $%  {$hood p/calm q/(trel beam beam cage) r/hood}     ::  compile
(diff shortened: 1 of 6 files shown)


ohAitch:
Not separating this into like 5 feature branches for each of the bullet points signed off by @galenwp; as is breaching this in tomorrow if necessary, though preliminary testing against a **master** pill says 2d0dd16 should take care of that.

juped:
I don't quite get it - is it done? Is it a breach? Is part of it a breach?

galenwp:
I think this is mergeable, but it's a little risky.  I say merge it and deploy it, and we'll figure it out later.  @ohAitch?

ohAitch:
Yeah basically. I would run basic tests wrt deploying on a fake master-network, checking that http://localhost:8443/static still works, etc. but it should be fine.

galenwp:
This should (hopefully) fix the talk reloading issue.

juped:
It's really much better overall to have the breaching parts be a separate branch and merge that at breach time, rather than stubbing it in as a dead code path and making it undead at breach time. Just for future consideration...

ohAitch:
There are no longer any breaching parts. The dead code path isn't even /stubbed/, and there might in fact blocked requests using the horn, which this until-breach version of `%ford` will execute correctly. (To the extent I've tested it, which is not comprehensive)

Type-safe state transitions etc.

juped:
>temporary rename of %see to %sei
>REVERT on breach

ohAitch:
Yes

After the next breach, the backwards-compatible type will not be necessary, and the old behavior should be removed(and the `%see` name reclaimed). Until then, the change works fine, it just has a slightly ugly `$%` tag.

ohAitch:
*should perhaps have named the commit "NO LONGER BREACHING rename %see to %sei"*

juped:
OK, if it *does* nothing but make the types nest, sure.

ohAitch:
You have to do ~something~ to old-typed data, and "recognize it as perfectly valid new-typed data" is the simplest option.

To expand:
Breaches are philosophically wrong. An urbit should never lose its mind, even when acquiring new/better functionality. If we were on a proper network, this would have been a `%zuse` kelvin bump, with ford getting a state-adapter that walked through all of the `horn`s it held and converted them to the new type; as stands, I'm fine with a simpler-but-less elegant approach which still perfectly preserves type safety.

juped:
I am aware that breaches are a terrible hack, which is why I only care about functional changes. If the only difference between this tree and a breached version of it is the names in the type, that's cool. If functionality would be added on the breach, that would be less cool.

juped:
(consider e.g. a simple bisect leading you to a commit that says "uncomment that stuff that was commented out")

ohAitch:
The only post-breach action this requires is a revert of the last commit, which purely renamed the new type and added the old one(/handling) back in. The pull as stands already has all the new functionality, promise :3

juped:
Sure. I'm just on guard against "this is something, therefore we must merge it". :)

ohAitch:
Well, specifically it is performance fixes to ford, without which e.g. webtalk is unusable. There is in fact a question of stability, but the "therefore we must merge it" is on fairly firm ground.

PR #29, opened 2016-02-25
Fix homepage and menu overlapping

galenwp:
tree@https://github.com/urbit/tree/commit/88abd896771460b698c905c36d8c2372e2090d4d

### DIFF
--- a/web.md
+++ b/web.md
@@ -4,16 +4,12 @@ anchor: none
 
 # Hello.
 
-<hr></hr>
-
 <h2 class="even">
     <a class="blue inverse block" href="/~~/talk">Talk</a>
 </h2>
 <h2 class="even">
     <a class="blue inverse block" href="/~~/dojo">Dojo</a>
 </h2>
-
-<hr></hr>
 
 <h2 class="even">
     <a class="blue inverse block" href="/static">Static files</a>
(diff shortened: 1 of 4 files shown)


galenwp:
Hm, that's nice.  Looks like my next commit will go in too?

ohAitch:
PRs track a branch mhm

On Wed, Feb 24, 2016 at 4:31 PM, Galen Wolfe-Pauly <notifications@github.com

juped:
Yeah, that's how Github works - it's not actually that nice (you can't ever review anything for sure) but in this case you have hit the benefits and not the drawbacks.

Just comment when it's really done :)

galenwp:
Heh, yes — do in fact think this is done.  But there will likely be a continued stream of 'various fixes and improvements' to the front end.  Is there any reason not to just keep them all on this branch?

juped:
You could certainly do that. I'd simply request that you not merge master into the branch. (It would have no benefits and some drawbacks.)

ohAitch:
I think srs projects have bots on top of the system that track the time at
which someone e.g. comments "r+", and only merge if no commits have been
pushed since.

juped:
(rebasing *the branch* onto *master* would be ok, but that's your call. Here's hoping there's no more off by ones in clay...)

PR #35, opened 2016-02-26
Fix ford's occasional truncation of hoon files

philipcmonk:
NB: again, not attaching a pill, but one should be built.


juped:
This one's not breaching, right? I got distracted by the other branch, which does breach...

philipcmonk:
Yes, not breaching.

ohAitch:
`{{Citation Needed}}`

(pretty sure #38 is caused by this changing the prepatch image of all hoon files)

PR #36, opened 2016-02-29
Routed .tree 'comments' input through talk

ohAitch:
Requires %eyre reload
Wants further testing before being announced / used in public-facing material


juped:
What does that mean? Keep it off the network, keep it off urbit.org?

ohAitch:
Keep md files with

---
comments: y
---

off urbit.org, as previously when this functionality lived purely in
/+write. @galenwp this should otherwise be deployed yes?

galenwp:
There’s no risk of deploying this to anything public accidentally.  This is good to go, as far as I know.

PR #48, opened 2016-03-03
/web/pages

galenwp:
`/web/app` didn't seem like the right place, so we're switching to `/web/pages`.

`urbit/examples` is being updated to reflect this change.  Should land today.  This should go out when examples is up to date.


galenwp:
Okay this is ugly, let me try this again.

PR #49, opened 2016-03-03
move to web/pages

galenwp:
Should go out when the examples repo is updated.  Later today, I expect.


juped:
What's the change this is waiting on, exactly?

galenwp:
We're just forward porting `urbit/examples` to match.  We can just merge though.

PR #51, opened 2016-03-03
moved js dependencies to be requested from cdn

ohAitch:


galenwp:
(Can we put this out on the network?  I can hardly wait)

juped:
Will arrive shortly!

galenwp:
Thank you, thank you.

PR #58, opened 2016-03-07
Fix cat

chc4:
Someone's merge got messed up, looks like.

### DIFF
--- a/gen/cat.hoon
+++ b/gen/cat.hoon
@@ -9,7 +9,7 @@
 ::::
   ::
 :-  %say
-|=  {^ {arg/(list path)} vane=?($c $g)}
+|=  {^ {arg/(list path)} vane/?($c $g)}
 =-  tang+(flop `tang`(zing -))
 %+  turn  arg
 |=  pax/path
@@ -17,7 +17,7 @@
 =+  ark=.^(arch (cat 3 vane %y) pax)
 ?^  fil.ark
   ?:  =(%sched -:(flop pax))
-    [> .^((map ,@da cord) (cat 3 vane %x) pax)<]~
+    [>.^((map @da cord) (cat 3 vane %x) pax)<]~
   [leaf+(spud pax) (pretty-file .^(noun (cat 3 vane %x) pax))]
 ?-     dir.ark                                          ::  handle ambiguity
     $~


philipcmonk:
That was likely my fault.  Thanks!

ohAitch:
Can confirm blame of @pcmonk, I hit this too :P

juped:
It was just incomplete 151ization. Happens to everyone.

ohAitch:
Sure, given my recent #ERROR I have no high ground here

philipcmonk:
Or rather, two independently completed 151izations incomp(lete/etent)ly
merged.

juped:
The history knows the truth...
image

juped:
Merge diffs take a bit of practice to learn to read, but gitk always colors the parents in order with the same colors: red, blue, green, purple, orangebrown, teal, magenta and after that I don't really have it memorized. This is a simple merge of 2 histories, so the "ours" side, some Philip branch, is red (as always) and the stuff being merged in, the tag `hoon-151`, is blue. It's pretty hard to figure out a diff-of-diffs by looking at the +s and -s, hence the bolding and colorcoding - we ended up with the bold stuff in the final tree. Normal-weight black lines with no +s or -s from the diff are just our diff context, but the bold black lines are things that existed in neither parent - i.e., our handcrafted conflict resolutions. We see in this diff how `vane=?($c $g)` and `(map ,@da cord)` got in via this resolution's incomplete 151ization.

philipcmonk:
People often ask me when colorblindness does or doesn't impact my life.  Here, I have no real idea of how many colors are in that diff, except that there's at least three, and if I had to guess before reading your comment, I would have said five or six.  You see, colorblind people use other hints, like brightness and texture, to distinguish colors.  When I see a bold black and a light black, and red and blue are already used, my mind decides the light black must be green.  When I see a light blue and a dark, bold blue, the dark blue must be purple.  Ditto for red/orange.

They say this skill is useful in seeing through camo, but that diff is absolutely incomprehensible to me.  Thank you for the explanation, and I accept responsibility for incompletely 151izing (excusable, mistakes happen) and testing (inexcusable).

cgyarvin:
Ha!

I wonder if colorblindness is covered by the ADA.  I suppose if you have to
ask... can we sue Linus?

juped:
You can always sue anyone for any reason, but you might annoy the judge. You can attempt to set up your gitk with colors you can disambiguate - it's not documented, but it's all in the default configuration file, which is probably ~/.config/git/gitk. If that's not possible, then all that's left is the +/- columns - a + means it's new compared to that side of the merge. So something with a + in every column is new to every side, and is therefore a handmade resolution. This takes even more practice to learn to read, alas. However, it has information content about the merge, unlike the diffs the incredibly rich company GitHub spent $a lot of money to provide you.

PR #70, opened 2016-03-14
should have been cleaned by factor

galenwp:
As discussed: pretty sure these can be taken out for now.  Right @cgyarvin?


ohAitch:
Is lunt the "local subs exercise terminal-permissions on their
co-hosted ships" vane? All of these should come out of the tree, though I'd
want to double check that there exist branches for their development.

cgyarvin:
Yes.

On Mon, Mar 14, 2016 at 1:05 PM, Galen Wolfe-Pauly <notifications@github.com

galenwp:
They certainly exist in the history.

galenwp:
Let the record show @ohAitch and @cgyarvin agree that this should be merged.

galenwp:
Although @ohAitch and I agree that having the drafts of these vanes exist at a tag would be convenient.

PR #75, opened 2016-03-16
urb

philipcmonk:
The is a breaching change.  Additionally, this should be merged at the same time as #723 of urbit/urbit, though neither branch should cause catastrophic failure if it's merged without the other.

This instruments eyre and the dojo for use with [urb](https://github.com/urbit/urb).


ohAitch:
856ce75, and 4e12cae in general(as merged to web-stream in 492718f), seems relevant, as it debreaches the changes(and matches the urbit/urbit PR). Does splitting `:pipe` from out in the middle make sense here?

juped:
Nonbreaching version to go in soon

Issue #80, opened 2016-03-21
arc should just be called sur in the dojo


galenwp:
If we call libraries `lib` then shouldn't we just call structures `sur`?  Not sure what an `arc` is, and don't really feel like I need to know.

galenwp:
Based on a quick test a few find / replaces should do the trick.  Now using a dojo with `sur` and it seems to work.

Will wait on @cgyarvin and @philipcmonk.

ohAitch:
+1

cgyarvin:
Yeah I was going to change these, but right now just make them consistent.

galenwp:
What would they change to?

galenwp:
@cgyarvin if you don't have something to change them to then they're just going to be `lib` and `sur` as per https://github.com/urbit/arvo/pull/83

Fang-:
#83, which fixes this, has been merged. This issue can be closed.

PR #83, opened 2016-03-22
moved arc to sur in dojo

galenwp:
I don't totally know what I'm doing here — but this seemed like a simple find and replace.


galenwp:
Fixes https://github.com/urbit/arvo/issues/80

PR #84, opened 2016-03-22
forwards

galenwp:
things are weird enough.  `+hello 'world'` should do what you expect.


ohAitch:
I think the flop is the problem there, not the trip. Does this version compile?

galenwp:
Ha, thanks.  I had to do this twice because I screwed up the first branch.  Will re-push.

PR #94, opened 2016-03-28
Return 31-bit value for pair mug

### LINKED ISSUE #93: mug hoon doesn't snip to 31 bits for pairs
Try running `(mug [0 151])` using the hoon code, not the jet, you should get a value with bit 32 set.

I think [this line](https://github.com/urbit/arvo/blob/a1df9b2a5991bd0c288ff7acfe92d883b6b35207/arvo/hoon.hoon#L1091) should be

    ?. =(0 d) d

instead.

rsaarelm:
Should fix #93


juped:
Thanks!

Issue #95, opened 2016-03-29
++mold $-(* *) is a bad example of a mold


ohAitch:
`++clam` used to be, and `++mold` should be, `_,*`. `(*$-(* *) a)` is 0 instead of a, making generic containers like `(list)` misnormalize their contents to 0, breaking constructions like `;;({(unit) (map)} foo)`.

galenwp:
What is the change that needs to be made?  Not sure I understand what you're saying.

ohAitch:
Change the `++mold` arm in `hoon.hoon` to be `_*` instead of `$~(* *)`, after checking with @cgyarvin that there isn't any more readable way to achieve the same goal.
(`_|=(* +<)` would be more _verbose_ but I don't think it actually represents what moods are very well)

cgyarvin:
This is the smallest nit I've ever seen in my life.  But fine, Anton, please patch it on master.  Better to just PR things this small rather than creating an issue to debate.

ohAitch:
The reason I brought this to your attention was that I'm not satisfied with `_*` being informative enough, especially since I believe you made the exact opposite edit as part of the Hoon 151 restructuring. I'll take this as go ahead — I think we may have discussed this I person since, in which case it's probably folded into one of my cc-release clean ups, but I'll double check.

belisarius222:
In current master, `++mold` is defined as `gate`, and `++gate` is defined as `$-(* *)`, so this has not been changed. Not sure if the issue is still relevant as of Hoon %143.

PR #100, opened 2016-04-02
Fix type-fail in ;;.

cgyarvin:


juped:
Is this #95's fix?

cgyarvin:
No.

PR #108, opened 2016-04-07
Asana driver and comments

ault011:
do you mind all of these commits combined into one?

### DIFF
--- a/sec/com/asana.hoon
+++ b/sec/com/asana.hoon
@@ -7,31 +7,33 @@
   ::
 =+  ^=  aut
     %+  oauth2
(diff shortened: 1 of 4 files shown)


ohAitch:
As long as the commits themselves don't get trollsquashed, I'm downstream of the middle of this

juped:
While I would have done separate drivers on separate branches I can't say it's a huge deal

juped:
I will say that the topology is mega weird
image

ohAitch:
Do those, like, converge at some point?

juped:
I don't know what that means. It's not really the worst thing in the world but I doubt it was intentionally done like that.

PR #110, opened 2016-04-12
renamed security driver arms

philipcmonk:
As discussed in https://github.com/urbit/urbit/issues/692, renamed security driver arms as follows:

```
++  out -> filter-request
++  res -> filter-response
++  in -> receive-auth-query-string
++  bak -> receive-auth-response
++  upd -> update
++  wyp -> discard-state
```

All the drivers have been updated, and they all compile, but I haven't done an end-to-end test of any of them except Github.  My testing of the others consisted of verifying that `+https://domain.com` produced an `%oauth-no-keys` error.  If @ault011 or @ohAitch want to test them, that'd be great.  Would like @ohAitch to review, in any case.

Currently this is a breaching change because for some reason wires are a part of eyre's state somewhere.  I'm not familiar enough with eyre to quickly write a state adapter.

### DIFF
--- a/arvo/eyre.hoon
+++ b/arvo/eyre.hoon
@@ -67,10 +67,10 @@
 ++  whir-of  {p/knot:ship q/term r/wire}                ::  path in dock
 ++  whir-se  ?($core vi-arm)                            ::  build/call
 ++  vi-arm
-  $?  $out                                              ::  ++out mod request
-      $res                                              ::  ++res use result
-      $bak                                              ::  ++bak auth response
-      $in                                               ::  ++in handle code 
+  $?  $filter-request                                   ::  ++out mod request
+      $filter-response                                  ::  ++res use result
+      $receive-auth-response                            ::  ++bak auth response
+      $receive-auth-query-string                        ::  ++in handle code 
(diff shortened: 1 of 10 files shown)


ohAitch:
You would definitely still need the old wire names around as long as you
might have any events coming back on them; I'm not sure how to address this
in a reasonably straightforward way.

philipcmonk:
Easiest would be to accept both versions of the names in wires until the next breach.

ohAitch:
You can just add the new names to whir-se, and switchboard logic to ++vi,
yeah.

ohAitch:
Also, thoughts on the ~(simple auth-whatever ...) arm names? The old
interface names made sense as prefixes; porting that forward directly would
get a bit unwieldy, though I suppose I'm not philosophically opposed to
"++receive-auth-query-string-exchange-code-for-token" as long as I never
ever have to debug it.

juped:
Is this updated to accept both versions of the names?

philipcmonk:
Not yet

juped:
This is in in a non-breaching way

PR #129, opened 2016-05-04
Better gh app, various

philipcmonk:
- rewritten %gh to provide real porcelain
- marked gall %x scry
- improvements to %gh through fuse
- cherry-picked #128 to allow testing of fuse for clay over the network

From inspection, I don't believe there are any breaching changes in here.


juped:
I'm still too dumb to figure this one out, I'm afraid. :( I see some stuff getting removed in the first patch there, but I can't make head or tail of it in general.

philipcmonk:
How do you mean, "figure out"?  I don't think there'll be any problem pushing this out.  The only thing that changes state is `%gh`, and I don't think anyone's started that app this era, so we should be good.

juped:
#135 not an issue?

philipcmonk:
Eh, it's still an issue in that the changes are still gone, but it doesn't need to block this.

PR #131, opened 2016-05-05
Tree performance improvements

ohAitch:
- concatenate urbit assets
- use minified resources
- preload more data
- meeker dependency handling


ohAitch:
kernel?

ohAitch:
oh right, there was an eyre embedded js tweak

PR #143, opened 2016-05-19
Take ownership of ~nym

max19:
Was thinking Douay-Rheims, but don't want to look like a fogey, so RSVCE I guess.


philipcmonk:
Our policies have changed, and we now accept only Da Hawai'i Pidgin Bible.

PR #146, opened 2016-05-24
Comments, talk feed, &c

galenwp:
General frontend improvements.  This has changes in `%e` and `%f`, but @ohAitch can advise as to how safe this is to deploy.


galenwp:
Oh, this can probably go out.  Right @ohAitch ?

juped:
I don't see why not.

juped:
Actually, I guess those are real type changes, not renames. I'll wait for him

ohAitch:
All the state adapters are there, and this has been run on every urbit.org ship recently, so it should be fine.

juped:
I'll believe you over Github!

PR #164, opened 2016-06-14
fixes run:in (map over a set, producing a set)

joemfb:
The current implementation produces an "improper list" (ie, non-terminated), and requires an unusual calling convention. This version just takes a gate and applies to every item in the set (from left to right), producing a new set containing the products of the gate.

```
=+  a=(sy ~[1 2 3])
(~(run in a) |=(a/@ (gth 2 a)))
```

produces `{%.y %.n}`

### DIFF
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -2452,10 +2452,13 @@
     $(a r.a, +<+.b $(a l.a, +<+.b (b n.a +<+.b)))
   ::
   +-  run                                               ::  apply gate to values
-    |*  {b/gate c/*}
-    |-
-    ?~  a  c
-    $(a r.a, c [(b n.a) $(a l.a)])
+    ~/  %run
+    |*  b/gate
+    =|  c/(set _?>(?=(^ a) (b n.a)))
+    |-  ?~  a  c
+    =.  c  (~(put in c) (b n.a))
+    =.  c  $(a l.a, c c)
+    $(a r.a, c c)
   ::
   +-  tap                                               ::  convert to list
     ~/  %tap


joemfb:
I should point out that I'm using a bit of a hack to get the product type - I call the gate on the first element of the set, and unbunt that to get a mold: `(set _?>(?=(^ a) (b n.a)))`. Is there way to get the span of a gate (meaning the span of the product of a gate) without calling the gate?

philipcmonk:
In general, you should be able to just pull `$` from the gate, leaving the default arguments.  The following three forms should be equivalent:  `(b)`, `$:b`, `*b`.

I feel like this may cause an issue or two if the gate is wet, but I'm not sure they'd work well in `run:in` anyway.

juped:
You want `^.`.

juped:
Example: `_?>(?=(^ a) ^.(b n.a))`. It's exactly what you asked for.

juped:
Actually, I think that only works on fixpoints... leaving Philip's `*b` the best option, I guess.

philipcmonk:
Honestly, while `*b` is idiomatic, it feels a little weird, since `b` isn't really a type.  I know, `*` of a gate just pulls `$`, but I feel like one of the other options would make more sense.

juped:
We do want (b n.a) specifically here. I don't *think* it's ever actually run (it *could* be optimized out) if that line's changed into a `=|` as it should be, so leaving it is probably best.

Style points:

- don't put `b/gate` in `{}`s
- have a `^-` or `^+` in there somewhere to give the output type
- or at least `=|` your accumulator
- put a `~/` in there in case someone ever decides to jet it (full hints should be present inside hoon.hoon, imo)

philipcmonk:
Why `(b n.a)` rather than `(b)`?

joemfb:
Couldn't the type of the sample determine the type the product?

juped:
In case it's wet.

philipcmonk:
What happens if you run `(~(run in ~) $~)`?

joemfb:
Produces the empty set

juped:
`set` bunts to `(set *)`.

joemfb:
I applied your style points, all except for an overall product cast. Is it worth binding `(set _?>(?=(^ a) (b n.a)))` to a name, and then doing `=|  c/my-type` and `^-  my-type`?

philipcmonk:
I don't think `set` gets bunted?  If it works, I suspect it's deciding that it's `(set _!!)` (in spirit, I'm not sure if that works directly).  In other words, a set of elements of type `%void`.  Which would be correct.

At any rate, this is the sort of situation where it'd be really nice to have the old vulcanization back, because wet gates with arity > 1 still aren't going to work.

joemfb:
Well, with `*=`, I was explicitly bunting the set. Not idiomatic, as compared to `=|`, but it would handle the empty case with more type info, right?

juped:
`=|` is just `=+` bunt, it's only a style thing in that having the mold of the product appear somewhere is helpful when reading something.

arvo/hoon.hoon
```diff
@@ -2452,10 +2452,15 @@
...
+    ~/  %run
+    |*  b/gate
+    =|  c/(set _?>(?=(^ a) (b n.a)))
+    |-  ?~  a  c
+    =+  ^=  c
+      (~(put in c) (b n.a))
+    =+  ^=  c
+      $(a l.a, c c)
+    $(a r.a, c c)
   ::
```
joemfb: While I'm golfing here, would you prefer to see the recursion happen on one line, rather than shadowing `c` twice?

```
$(a r.a, c $(a l.a, c (~(put in c) (b n.a))))
```

juped: What you're actually attempting to do is `=.  c`

joemfb: Thanks! I'd missed that one. Do you prefer `=.` with the recursions on separate lines, or everything nested on one line?

juped: Do what looks nicer to you, but whatever you do, avoid closing bracket pileups by using tall forms. (The type incantation in the `=|` is a piece of cosmic insight, so that should stay on one line to avoid taking up the attention of the uninitiated.)

PR #166, opened 2016-06-15
Add xhr timeout property and handler to urb.js

knubie:
The timeout is set to 1 minute, and passes an error to the callback. Without a timeout the request would pend forever, which creates two problems.

1. If the client loses connection, the request is never re-established when connection is restored, and the user is forced to refresh the browser.
2. The `urb.poll.dely` property never changes so urbit/talk is unable to detect when it's offline.

I've also made a `beat: true` response reset the `urb.poll.dely` property back to 250 so urbit/talk can go back 'online' again.

Fixes urbit/talk#21


juped:
Equivalent patch merged

PR #169, opened 2016-06-15
fix overload

ohAitch:


chc4:
Silly question, but what does |overload even *do*? I tried following it into kiln, but I don't know what it means.

philipcmonk:
It's cache invalidation for ford.  The best algorithm would probably be the "clock" algorithm.  `|overload` is what we call the "wall clock" algorithm of cache invalidation (clear the cache every `n` hours).

PR #172, opened 2016-06-17
updates +moon to print an error message instead of a stack trace

joemfb:
...when run on a non-planet ship, such as a comet.

Some of the tutorials recommend creating a moon, and just print the stack trace isn't beginner-friendly.

### DIFF
--- a/gen/moon.hoon
+++ b/gen/moon.hoon
@@ -10,8 +10,10 @@
         $~
         $~
     ==
-:-  %noun
-?>  =(1 (met 5 p.bec))
+:-  %tang  :_  ~  :-  %leaf
+=+  ran=(clan p.bec)
+?:  ?=({?($earl $pawn)} ran)
+  "can't create a moon from a {?:(?=($earl ran) "moon" "comet")}"
 =+  mon=(mix (lsh 5 1 (end 5 1 eny)) p.bec)
 =+  tic=.^(@ /a/(scot %p p.bec)/tick/(scot %da now)/(scot %p mon))
 "moon: {<`@p`mon>}; ticket: {<`@p`tic>}"


cgyarvin:
To say the least!

juped:
Better: instead of removing the `!:` and causing a crash anyway, don't crash and print an actual message
Best: rewrite the whole thing to use the `%tang` mark, which is our best way of just printing lines to the screen in a `%say`

joemfb:
What should the message be? "only planets can create a moon"?

re:better, should such a generator produce the message as a tape (or cord), or print the message with `~&` and produce `~`?

I'll look into %tang.

juped:
`~&` is an out of band print mostly meant for debugging.

Not only planets can create a moon, but things larger than planets can't. (The test should actually be `(lte (met 5 our) 1)`, because `met` measures zero as having width zero at any bloqsize, and `~zod` can ticket.)

joemfb:
Ok, now it produces instead of prints the failure message, includes the ship type in the message, and supports `~zod`.

joemfb:
The message was too long when it included the comet name, but it's a little stark now. Thoughts?

cgyarvin:
You're learning our style too well!

gen/moon.hoon
```diff
@@ -10,8 +10,11 @@
...
         $~
     ==
-:-  %noun
-?>  =(1 (met 5 p.bec))
+:-  %tang  :_  ~  :-  %leaf
+=+  len=(met 5 p.bec)
+?.  (lte len 1)
+  =+  typ=?:(=(2 len) "moon" "comet")
+  "can't create a moon from a {typ}"
 =+  mon=(mix (lsh 5 1 (end 5 1 eny)) p.bec)
```
joemfb: This might be golfing a little, but I've been reading src and came across `++clan` and `++rank`...

Would this be more idiomatic ?

```
=+  ran=(clan p.bec)
?:  ?=({?($earl $pawn)} ran)
  "can't create a moon from a {?:(?=($earl ran) "moon" "comet")}"
```

juped: Yes,

PR #176, opened 2016-06-21
spin and spun

chc4:
Add two new arms, ++spin and ++spun, for "turn with state". This allows you to, for example, increment a counter while mapping over a list.  ++spun uses the sample of the gate as a starting point instead of being explicit.

```
=+  t=(limo ~[4 5 6])
%+  spun
  (limo ~[1 2 3])
|=({n/@ a/@} [[n (snag a t)] +(a)])
```

```
%^  spin  (reap 10 0)
  |=({n/@ rng/_og} (rads:rng 500))
~(. og eny)
```

I tried adding an assert to ++spin that asserts the gate is typed correctly (since electroplating makes error messages weird), but the commented line triggers a panic :/


juped:
I think it's still typeable, but it probably involves building an example out of the whole sample. (Which would be ugly, because we lack a wet `|:`). I believe it'll still fail when it needs to fail, even with an opaque message.

PR #184, opened 2016-06-27
Show message loading indicator in /stream

ohAitch:
+ misc stuff accumulated from master, can be rebased out if necessary


juped:
Just ignore anything github says, it's always terminally confused.

ohAitch:
No, github is right, this is a PR against urbit-org not master

Tho like the change should also get into master eventually

juped:
As far as I'm concerned urbit-org doesn't exist, but shouldn't that ship get normal updates as well as its own junk?

ohAitch:
The "junk" is a fast-tracked talk UI change; I think it's unsubscribed from
updates for performance-predictability reasons. But yes, merging this is a
question for @philipcmonk

philipcmonk:
I don't think there's any problem with just merging this and putting it all
on the urbit.org ship.  The talk change is also PR'd to master somewhere?

juped:
This is why you should never attempt to do what you're doing with git. You
end up needing to write a bot...

Yours,
r

ohAitch:
The thing I'm attempting with git is organizational synchronization /
handoff of responsibility :P

In particular, the initial question "should I rebase all the master drift
out" is the reason I didn't just pull this in directly and push urbit-org
from prod afterwards

juped:
I thought you were mirroring urbit.org...

Yours,
r

ohAitch:
Mhm, this is why a pull request into the branch seemed like a good place to
discuss what should be on urbit.org

juped:
We've ascended past this over on urbit.org

PR #188, opened 2016-06-30
deduplicates dojo history (by skipping repeat entries)

joemfb:
`?=(^ old.hit)` is in there to support bootstrap, where history is presumably `~`

### DIFF
--- a/lib/drum.hoon
+++ b/lib/drum.hoon
@@ -25,7 +25,7 @@
       cas/case                                          ::  boot case
   ==                                                    ::
 ++  source                                              ::  input device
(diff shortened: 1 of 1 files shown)


joemfb:
Confirmed that this works for bootstrapping a new pier, and that it doesn't without the `?=(^` ... ;)

joemfb:
Also aligns a margin comment (but ignores the copious trailing whitespace ...)

joemfb:
Empty lines are currently added to the history; is that by design, or would it be fine to filter those as well?

cgyarvin:
It would be totally fine to filter those!

joemfb:
done

PR #189, opened 2016-06-30
move robots.txt out of %eyre, permit crawling

ohAitch:


philipcmonk:
Did we want to fast track this for urbit.org?

ohAitch:
Yes

ohAitch:
Hm, this causes problems for tree children/sibling lists. Should robots stay hard-coded in eyre, but with Dissallow: none? Should it just be left to 404, which afaict has the same semantics? Other options are taking out the txt->html translation, or blacklisting the file by name/metadata/a comment in it, but those seems hacky / undermining of tree as a generic file browser.

Issue #191, opened 2016-07-01
inconsistencies resolving .^ in an app vs a generator


joemfb:
Not sure if this is the right place for this ... 

I'm butting my head trying to resolve a `.^` inside an app, when the same twig works in a generator.

In `:dojo`:

```
> .^(json %cx /=home/348/web/easy/json)
>> [%camping %c %x [p=~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer q=%home r=[%ud p=348]] s=/json/easy/web]
... :: json object printed
```

When I evaluate that twig in a generator, I see the same `[%camping ...]` printf, but inside an app, it just prints the path and blocks:

```
/cx/~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/home/348/web/easy/json
```

(same behavior in the `++prep` and `++poke` arms).

While I have ~1000 other questions about `.^`, which is pretty likely documented, this is the obstacle I haven't been able to work around.

joemfb:
Seemingly relevant tidbit:

In `:dojo`:

```
> .^(hart %e /(scot %p our)/host/real)
[p=%.y q=~ r=[%.y p=<|org urbit wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer|>]]
```

Whereas in `:talk`, the same twig just produces the path:

```
/e/~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/host/real
```

cgyarvin:
Wise question, glasshoppa!

The answer: blocking .^ is a very non-fundamental behavior.  It is
obviously not part of Nock proper (it turns into Nock instruction 11, which
is only implemented by the virtual extended Nock ++mock).  It is not even
available to %gall apps, which can only do non-blocking .^.  For the
programmer, the first layer blocking .^ works in is anything run by :dojo
(generators, command line, etc).

Internally, what you're doing with .^ from the dojo, and what you're doing
from a %gall app, are different.  From a %gall app, .^ will just use the
Arvo kernel namespace function ++scry on your behalf.  ++scry obviously has
no way to block.  So in Unix terms, it's essentially producing EWOULDBLOCK.

From the dojo, .^ is actually invoking the build vane, %ford.  The
superpowers of %ford are tremendous (and not well documented).  You're
essentially saying: start this build and call me back when it's done.  If
%ford does not actually have to block, it will give a result in the same
[...]

PR #192, opened 2016-07-02
adds support for meta (alt) keyboard shortcuts

joemfb:
Also adds implementations of meta-. and meta-backspace;  they're working (ie, tested with other keys), but commented out. Wiring them up will take changes to `dill-belt`, and the code-paths that construct instances of it. I tried changing `mar/dill/belt/hoon`, but it didn't seem to make a difference ...

Questions:

- where should the utility functions go (and how should they be named)?
- is there a better pattern to use in `++alnum`?
- what do I need to change to get meta+ backspace or `.` combinations into the `++ta-met` codepath?

### DIFF
--- a/lib/drum.hoon
+++ b/lib/drum.hoon
@@ -698,7 +698,19 @@
     =.  +>  (ta-dog(say.inp (~(commit sole say.inp) ted)) ted)
     +>
   ::
-  ++  alnum
+  ++  lcas                                            ::  lowercase
+    |*  a/(list @)
+    ^+  a
+    %+  turn  a
+    |=(a/@ ?.(&((gte a 'A') (lte a 'Z')) a (add 32 a)))
+  ::
+  ++  ucas                                            ::  uppercase
+    |*  a/(list @)
+    ^+  a
+    %+  turn  a
+    |=(a/@ ?.(&((gte a 'a') (lte a 'z')) a (sub a 32)))
+  ::
+  ++  alnm
     |=  a/@  ^-  ?
     ?|  &((gte a '0') (lte a '9'))
         &((gte a 'A') (lte a 'Z'))
@@ -707,10 +719,11 @@
   ::
   ++  next-edge
     |=  buf/(list @c)
-    =+  [i=0 m=|]
-    |-  ^-  @ud
+    =|  i/@ud
+    =+  m=|
+    |-  ^+  i
     ?~  buf  i
-    =+  w=(alnum i.buf)
+    =+  w=(alnm i.buf)
     =.  m  |(m w)
     ?:  &(m !|(=(0 i) w))
       i
@@ -718,9 +731,9 @@
   ::
   ++  next-word
     |=  buf/(list @c)
-    =+  i=0
-    |-  ^-  @ud
-    ?:  |(?=($~ buf) (alnum i.buf))
+    =|  i/@ud
+    |-  ^+  i
+    ?:  |(?=($~ buf) (alnm i.buf))
       i
     $(i +(i), buf t.buf)
   ::
@@ -761,7 +774,7 @@
             :~  %mor
               [%del sop]
               :+  %ins  sop
-              (turf (cuss [(tuft (snag sop buf.say.inp))]~))
+              (head (ucas (limo [(snag sop buf.say.inp)]~)))
             ==
             ::
       $d    ?:  =(pos.inp (lent buf.say.inp))
@@ -793,13 +806,13 @@
       ?($u $l)
             ?:  =(pos.inp (lent buf.say.inp))
               ta-bel
-            =+  case=?:(?=($u key) cuss cass)
+            =+  case=?:(?=($u key) ucas lcas)
             =+  sop=(add pos.inp (next-word (slag pos.inp buf.say.inp)))
             =+  f=(jump-fwrd sop buf.say.inp)
             %-  ta-hom
             :~  %mor
               (ta-cut sop (sub f pos.inp))
-              (ta-cat sop (tuba (trip (case (tufa (slag sop (scag f buf.say.inp)))))))
+              (ta-cat sop (case (slag sop (scag f buf.say.inp))))
             ==
     ==
   ::
--- a/lib/drum.hoon
+++ b/lib/drum.hoon
@@ -710,42 +710,42 @@
     %+  turn  a
     |=(a/@ ?.(&((gte a 'a') (lte a 'z')) a (sub a 32)))
   ::
-  ++  alnm
+  ++  alnm                                            ::  alpha-numeric
     |=  a/@  ^-  ?
     ?|  &((gte a '0') (lte a '9'))
         &((gte a 'A') (lte a 'Z'))
         &((gte a 'a') (lte a 'z'))
     ==
   ::
-  ++  next-edge
-    |=  buf/(list @c)
+  ++  nedg                                            ::  next boundary offset
+    |=  a/(list @)
     =|  i/@ud
-    =+  m=|
+    =+  b=|
     |-  ^+  i
-    ?~  buf  i
-    =+  w=(alnm i.buf)
-    =.  m  |(m w)
-    ?:  &(m !|(=(0 i) w))
+    ?~  a  i
+    =+  c=(alnm i.a)
+    =.  b  |(b c)
+    ?:  &(b !|(=(0 i) c))
       i
-    $(i +(i), buf t.buf)
+    $(i +(i), a t.a)
   ::
-  ++  next-word
-    |=  buf/(list @c)
+  ++  nwrd                                            ::  word-offset
+    |=  a/(list @)
     =|  i/@ud
     |-  ^+  i
-    ?:  |(?=($~ buf) (alnm i.buf))
+    ?:  |(?=($~ a) (alnm i.a))
       i
-    $(i +(i), buf t.buf)
+    $(i +(i), a t.a)
   ::
-  ++  jump-bwrd
-    |=  {pos/@ud buf/(list @c)}
+  ++  bwrd                                            :: prev pos by offset
+    |=  {a/@ud b/(list @) c/$-((list @) @)}
     ^-  @ud
-    (sub pos (next-edge (flop (scag pos buf))))
+    (sub a (c (flop (scag a b))))
   ::
-  ++  jump-fwrd
-    |=  {pos/@ud buf/(list @c)}
+  ++  fwrd                                            :: next pos by offset
+    |=  {a/@ud b/(list @) c/$-((list @) @)}
     ^-  @ud
-    (add pos (next-edge (slag pos buf)))
+    (add a (c (slag a b)))
   ::
   ++  ta-met                                          ::  meta key
     |=  key/@ud
@@ -753,24 +753,24 @@
       $dot  ?.  &(?=(^ old.hit) ?=(^ -.old.hit))
               ta-bel
             =+  old=`(list @c)`-.old.hit
-            =+  b=(jump-bwrd (lent old) old)
+            =+  b=(bwrd (lent old) old nedg)
             %-  ta-hom(ris ~)
             (ta-cat pos.inp (slag b old))
             ::
       $bac  ?:  =(0 pos.inp)
               ta-bel
-            =+  b=(jump-bwrd pos.inp buf.say.inp)
+            =+  b=(bwrd pos.inp buf.say.inp nedg)
             %-  ta-hom(kil `(slag b (scag pos.inp buf.say.inp)), ris ~)
             (ta-cut b (sub pos.inp b))
             ::
       $b    ?:  =(0 pos.inp)
               ta-bel
-            +>(pos.inp (jump-bwrd pos.inp buf.say.inp))
+            +>(pos.inp (bwrd pos.inp buf.say.inp nedg))
             ::
       $c    ?:  =(pos.inp (lent buf.say.inp))
               ta-bel
-            =+  sop=(add pos.inp (next-word (slag pos.inp buf.say.inp)))
-            %-  ta-hom(pos.inp (jump-fwrd sop buf.say.inp))
+            =+  sop=(fwrd pos.inp buf.say.inp nwrd)
+            %-  ta-hom(pos.inp (fwrd sop buf.say.inp nedg))
             :~  %mor
               [%del sop]
               :+  %ins  sop
@@ -779,13 +779,13 @@
             ::
       $d    ?:  =(pos.inp (lent buf.say.inp))
               ta-bel
-            =+  f=(jump-fwrd pos.inp buf.say.inp)
+            =+  f=(fwrd pos.inp buf.say.inp nedg)
             %-  ta-hom(kil `(slag pos.inp (scag f buf.say.inp)), ris ~)
             (ta-cut pos.inp (sub f pos.inp))
             ::
       $f    ?:  =(pos.inp (lent buf.say.inp))
               ta-bel
-            +>(pos.inp (jump-fwrd pos.inp buf.say.inp))
+            +>(pos.inp (fwrd pos.inp buf.say.inp nedg))
             ::
       $r    %-  ta-hom(lay.hit (~(put by lay.hit) pos.hit ~))
             :~  %mor
@@ -795,12 +795,12 @@
               (snag (sub num.hit +(pos.hit)) old.hit)
             ==
             ::
-      $t    =+  a=(jump-fwrd pos.inp buf.say.inp)
-            =+  b=(jump-bwrd a buf.say.inp)
-            =+  c=(jump-bwrd b buf.say.inp)
+      $t    =+  a=(fwrd pos.inp buf.say.inp nedg)
+            =+  b=(bwrd a buf.say.inp nedg)
+            =+  c=(bwrd b buf.say.inp nedg)
             ?:  =(b c)
               ta-bel
-            =+  prev=`(pair @ud @ud)`[c (jump-fwrd c buf.say.inp)]
+            =+  prev=`(pair @ud @ud)`[c (fwrd c buf.say.inp nedg)]
             =+  next=`(pair @ud @ud)`[b a]
             %-  ta-hom(pos.inp q.next)
             :~  %mor
@@ -814,8 +814,8 @@
             ?:  =(pos.inp (lent buf.say.inp))
               ta-bel
             =+  case=?:(?=($u key) ucas lcas)
-            =+  sop=(add pos.inp (next-word (slag pos.inp buf.say.inp)))
-            =+  f=(jump-fwrd sop buf.say.inp)
+            =+  sop=(fwrd pos.inp buf.say.inp nwrd)
+            =+  f=(fwrd sop buf.say.inp nedg)
             %-  ta-hom
             :~  %mor
               (ta-cut sop (sub f pos.inp))
--- a/lib/drum.hoon
+++ b/lib/drum.hoon
@@ -698,10 +698,66 @@
     =.  +>  (ta-dog(say.inp (~(commit sole say.inp) ted)) ted)
     +>
   ::
+  ++  alnum
+    |=  a/@c  ^-  ?
+    ?|  &((gte a 48) (lte a 57))                      ::  0-9
+        &((gte a 65) (lte a 90))                      ::  A-Z
+        &((gte a 97) (lte a 122))                     ::  a-z
+    ==
+  ::
+  ++  next-word-pos
+    |=  buf/(list @c)
+    =+  i=0
+    |-  ^-  @ud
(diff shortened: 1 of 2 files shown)


ohAitch:
Ooh! Any chance of alt-arrows, if alt-backspace can work?

Dill-belt is in arvo, specifically zuse and dill(and perhaps the c code
backing dill). I'm not sure why only letters are coming through.

ohAitch:
Ooh! Any chance of alt-arrows, if alt-backspace can work?

Dill-belt is in arvo, specifically zuse and dill(and perhaps the c
interface driving it). Not sure why only letters would be coming through.

joemfb:
Woooah, I just tried and they already work! Something's already normalizing the input ... That's convenient, and maybe a clue for meta-. and meta-backspace

ohAitch:
I've seen `%met` stub messages when I reach for Alt keys, yeah

cgyarvin:
This is because you're resurrecting a feature that used to exist :-)

joemfb:
Well, *requesting* said resurrection ...

It looks like the input normalization is in `mar/dill/belt/hoon`: https://github.com/urbit/arvo/blob/master/mar/dill/belt.hoon#L39... I had tried changing that file, adding printfs and changing behavior, but none of those changes seemed to take. I've seen seen the same behavior with other files before (`app/talk/hoon`)... I shouldn't need to `|reset` or `|reload` specific vanes for any files in `app/`, `lib/`, `mar/`, etc., right?

joemfb:
It actually looks like some characters are rewritten in `term.c`: https://github.com/urbit/urbit/blob/master/vere/term.c#L818

joemfb:
meta-. and meta-backspace work now (in conjunction with urbit/urbit#758) ...

It looks like bash also implements meta t,u,l,c, and r (http://ss64.com/bash/syntax-keyboard.html). Interested in any or all of those?

cgyarvin:
(Claps.). Sure!

Any general thoughts on the application system as you've learned it?  Questions?

joemfb:
(Commenting here -- still gun-shy about replying to the email thread after #191. And I have a bad habit of editing my github comments after posting, which, I just realized doesn't play well with the email notifications...)

Still coming to grips with the overall application system; I'm really just nibbling around the edges with these PRs. And I have fractals of questions, as usual ;)

As for `:dojo`, it's pretty cool to hotpatch low-level components so easily. It's not yet easy for me to follow the control flow by reading, but I've just added printfs liberally. (It'd help if there was a way to output a stack trace without crashing... And an analogue for inspecting wires, when I get a better handle on moves)

For this PR: where should I put "utility" gates I've added? And is there some Hoonian alternative to regexps, or do I have to just assert the codepoint ranges as I've done?

cgyarvin:
Printfs are a great way to debug Hoon.  We love printfs (~&).

(Sorry about the **** stack traces (^C and stack overflows crash instead of
producing a good trace).  That should be a higher priority bug.  It's
worked for most of Urbit's history, we just let it break while we, you
know, made videos and stuff.  Don't miss the stacktrace printf rune ~|,
though.  Everyone has printfs, but not everyone has stacktrace printfs.)

There's nothing wrong with enumerating the codepoint ranges.  It's not like
those ranges are going to change.

If a core could be used by arbitrary userland code, but isn't likely to be
needed in the kernel, it goes in zuse.hoon.  Zuse has an unfortunate
resemblance to a garbage dump, which should be corrected at some point.

Turning on |verb will show you all vane-moves that are made.  It's a toggle
command.

We had a jetted regexp library, but it was kind of poor quality and so we
threw it out.  Would be nice to have another.

joemfb:
I had missed `~|`, that's perfect!

joemfb:
I've added meta-c,l,t, and u. I need to spend more time with the history implementation to figure out meta-r.

One problem: certain combinations are supposed to be cumulative in the kill-ring -- meta-backspace, meta-backspace, ctrl-y should yank both kills, while ctrl-k, meta-backspace should just yank the latter. I'm not yet sure, but I think the `kil` structure will need to be enhanced with some kind of flag to indicate the nature of the previous kill.

Also, this entire PR could probably use a Hoon style review; from faces to whitespace to favoring slag/scag over snag -- I can defend my choices, but that doesn't make them right ...

juped:
I would probably slag and scag too, because they just do what they should, unlike the *stupid* zero-based snag.

juped:
As a point of style, you may be leaning too hard on `=+` to use a top-down "imperative" style rather than an outside-in style, but you may also not be - sometimes you really do need to control the order of computation. Definitely `=|` a mold for things like your loop counters, though.

joemfb:
Thanks! Some of those `=+` twigs are just trying to control the right margin; there's a lot of intermediate gate calls on positional deltas, for instance

lib/drum.hoon
```diff
@@ -698,10 +698,115 @@
...
+            =+  ^=  prod
+            %-  ta-hom
+            :~  %mor
+              (ta-cut p.next (sub q.next p.next))
+              (ta-cat p.next (slag p.prev (scag q.prev buf.say.inp)))
+              (ta-cut p.prev (sub q.prev p.prev))
+              (ta-cat p.prev (slag p.next (scag q.next buf.say.inp)))
+            ==
+            prod(pos.inp q.next)
+            ::
```
joemfb: This is awkward, but lets me order the edits in a straightforward manner. The alternative includes a lot more offset calculation...

lib/drum.hoon
```diff
@@ -698,10 +698,115 @@
...
+            ?:  =(pos.inp (lent buf.say.inp))
+              ta-bel
+            =+  case=?:(?=($u key) cuss cass)
+            =+  sop=(add pos.inp (next-word (slag pos.inp buf.say.inp)))
+            =+  f=(jump-fwrd sop buf.say.inp)
+            %-  ta-hom
+            :~  %mor
+              (ta-cut sop (sub f pos.inp))
+              (ta-cat sop (tuba (trip (case (tufa (slag sop (scag f buf.say.inp)))))))
+            ==
```
joemfb: This seems absurd, but copying the stdlib gates `++cuss` and `++cass` so I can change the sample mold also seems absurd.

lib/drum.hoon
```diff
@@ -698,10 +698,115 @@
...
+      $bac  ?:  =(0 pos.inp)
+              ta-bel
+            =+  b=(jump-bwrd pos.inp buf.say.inp)
+            %-  ta-hom(kil `(slag b (scag pos.inp buf.say.inp)), ris ~)
+            (ta-cut b (sub pos.inp b))
+            ::
+      $b    ?:  =(0 pos.inp)
+              ta-bel
+            +>(pos.inp (jump-bwrd pos.inp buf.say.inp))
+            ::
```
joemfb: Should I call `++ta-hom` with a `$nop` edit, instead of directly change `+>`?

lib/drum.hoon
```diff
@@ -698,10 +698,115 @@
...
+            =+  sop=(add pos.inp (next-word (slag pos.inp buf.say.inp)))
+            =+  ^=  prod
+            %-  ta-hom
+            :~  %mor
+              [%del sop]
+              :+  %ins  sop
+              (turf (cuss [(tuft (snag sop buf.say.inp))]~))
+            ==
+            prod(pos.inp (jump-fwrd sop buf.say.inp))
+            ::
```
joemfb: This is also awkward, but pretty much required... It's an odd shortcut: a single char edit, then a positional jump.

juped:
Yeah. For right margin control, we have the `%` family to call gates without the attack of the parentheses. But clarity is king above all.

juped:
Also, I don't know what you mean by "copying gates to change the sample mold".

joemfb:
`++cuss` and `++cass` would work as-is for changing the case of a `tuba

joemfb:
oops...

but I have to convert to and from a `tape` to be able to call them, do the the type constraints of their "sample molds" (the mold that normalizes the sample?).

juped:
Using those gates is causing you more trouble than it solves, since they take tapes and produce cords for some weird reason. Helper functions are fine (and give you a place to swap in more comprehensive casing functionality, or replace with the name of an eventual zuse casemapping facility), perhaps defined in a `|^`?

joemfb:
I broke down and trimmed trailing whitespace in this fail... I can back it out of that's a bridge too far.

I added utility gates `++lcas` and `++ucas`; they're wet gates taking `(list @)` right now, but I can't fight the urge to do something like this:

    |*  a/?((list @) @)
    ^+  a
    =+  b=|=(a/@ ?.(&((gte a 'A') (lte a 'Z')) a (add 32 a)))
    ?:  ?=(@ a)
      (b a)
    (turn a b)

That's a `nest-fail`, but it seems like a reasonable piece of code ;) I can do subject surgery to get around it, but I don't know how to do that polymorphically with the list

juped:
It doesn't need to be general, and in fact shouldn't, you're just making a helper function. That's why I suggested a `|^` inside the u/l case. Drum isn't the stdlib.

juped:
Also, your sample code is a dangerous fork abuse. You need to tag your unions most of the time...

    > ? =+  a=*?(@ (list @))  ?>(?=(@ a) a)
      ?(@ $~)
    0

joemfb:
Yeah, i'm just complaining ... It seems like something that should be pretty straightforward, but maybe that's just my dynamic lang experience ...

I need to store two bits of state to support properly cumulative yanks: if the last belt was a yank, and, if so, in which direction. (From the middle of a line, ctrl-u,ctrl-k,ctrl-y should leave the line unchanged. That applies to any consecutive series of yanks, but any other input or navigation should break the chain).

That state could be in one face: `(unit ?($l $r)`, but it'd obviously be disproportionately disruptive to modify all the non-yanking belt codepaths to correctly reset that face. It could be handled in `++se-belt`, modifying the subject before applying the switch expression, but that also seems inelegant. Any thoughts on where that should go?

joemfb:
Too the best of my knowledge, this PR now implements all bash meta shortcuts except for `y`, which would require a real kill-ring.

The "cumulative yanks" issue I described above is a pre-existing problem, but acerbated by meta-d and meta-backspace.

juped:
>bash

You mean emacs keys :angry:

joemfb:
"emacs" keys ... meta-r, ctrl-u are bash only, meta-t, meta-. have different behavior

juped:
You're not allowed to add things that are bash-only, since Urbit is not Ubuntu. :) Fortunately, none of those things are bash-only... just normal shell emacs keys. (Except ^U which is in the terminal driver.)

joemfb:
I went ahead and added ctrl-w, since I had all the pieces ... I also made the utility gates more urbity

joemfb:
I implemented cumulative kills by storing the current and previous belts as a pair in `++target`.

joemfb:
Changing that data structure means a `prep-mismatch` on sync:

```
gall: %hood: prep mismatch
[%talk-peer-replaced 2 i=%~.sole t=/]
[%dojo-peer-replaced 2]
activated app base/dojo
activated app base/talk
```

juped:
Do you know how to write a `++prep`?

joemfb:
I do not, but had meant to ask ... That'd need to go in both `:dojo` and `:talk`, right?

juped:
Anything that uses the library - though you'd write the actual updater in the library itself, of course.

ohAitch:
Hm, the only file I see changed is drum, which means the relevant prep is
%hood, or rather a %hood hookup to in drum itself. I think examples of
those exist, but it definitely isn't a standard app one.

joemfb:
Added a kill-ring, 'cause I can't stop myself ...

`++prep` should be a gate that takes the old app-state and produces the new ... so, I should write a gate in drum that can handle that conversion, and then create a `++prep` arm in hood that gets `%drum` out of  `lac` in `hood-0` and calls into drum to upconvert it?

Will that be sufficient? I'm not really clear on the relationship between dojo and talk have with hood ...

juped:
Drum is the userspace interface to terminal stuff, I think. You need a gate that takes the old state type and produces a new one; see talk for a good example that spans multiple state versions.

Let me know when this is done done.

joemfb:
I'm trying to understand the state-versioning in `:talk`, but I'm stuck on the mold "transformations":

    ++  house-3                                         ::
      %+  cork  house-4  |=  house-4                    ::  modern house with
      +<(stories (~(run by stories) story-3))           ::  old stories
    ++  house-4                                         ::
      %+  cork  house-5  |=  house-5                    ::  modern house with
      +<(shells (~(run by shells) shell-4))             ::  no settings
    ++  house-5                                         ::
      $:  stories/(map knot story)                      ::  conversations
          general/(set bone)                            ::  meta-subscribe
          outbox/(pair @ud (map @ud thought))           ::  urbit outbox
          folks/(map ship human)                        ::  human identities
          shells/(map bone shell)                       ::  interaction state
          log/(map knot @ud)                            ::  logged to clay
          nik/(map (set partner) char)                  ::  bound station glyphs
          nak/(jug char (set partner))                  ::  station glyph lookup
      ==

`cork` is compose-forward, so `++house-3` and `++house-4` call the mold against their sample, then call the gate against the molded sample... That seems backwards, if the sample is an old version of the state, it first molds it against a newer version, than transforms it in some way? Won't the portions of the newer-version mold that *don't* match the sample just bunt?

The other thing that has me stuck is this pattern:

    ++  story-3  (cork story |=(story +<(|10 &11.+<)))

What do `|10` and `&11.+<` mean? (I know `+<` is `+6`, the gate sample)

ohAitch:
The old ones are improper molds, which only work as types. Really they
should probably be
    $_  %*  .  *new-mold
but in some cases you want the new bunt available.

For story-3, +<(|10 &11) is "story without the settings field", which is at
|11 in the new mold. Really the correct solution here is probably adding
new fields near the top, and having the step-down historical types be
consistenty |1.new |2.new etc. instead of new(end-of-new
old-last-field.new).

joemfb:
I can't find any docs for `|wing` or `&wing`; they appear to be yet another tree addressing syntax:

```
> &1:[1 2]
1
> |1:[1 2]
2
> &2:[[1 2] [3 4]]
3
> |2:[[1 2] [3 4]]
4
```

`&n` looks like "head of the cell at *n*" -- `|n` looks like "tail of the cell at *n*". Is that correct?

ohAitch:
Entirely. +>+>+>- and +>+>+>+ respectively, where the total length / number
of derefs is the n in |n or &n.

ohAitch:
I suppose -.+.+.+ and +.+.+.+ might be clearer here, for explaining
=(&4:"abcdef" 'd'), =(|4:"abcdef" "ef")

juped:
In other words, you're not right at all.

    > &0:[1 2 3 4 5 6 7 8 9]
    error
    > |0:[1 2 3 4 5 6 7 8 9]
    [1 2 3 4 5 6 7 8 9]
    > &1:[1 2 3 4 5 6 7 8 9]
    1
    > |1:[1 2 3 4 5 6 7 8 9]
    [2 3 4 5 6 7 8 9]
    > &2:[1 2 3 4 5 6 7 8 9]
    2
    > |2:[1 2 3 4 5 6 7 8 9]
    [3 4 5 6 7 8 9]

and so on

ohAitch:
Idk about not right /at all/: it's translating the tuple meaning of "at
index n" into an actual nock axis, so &2 becomes +<. The examples work
afaict.

juped:
As for the other thing, as the comment says, it's a modern house, but with the old stories, by molding the stories with an old story type. This is a weird intermediate state.

`+<` is axis 6, one of the children of axis 3.

joemfb:
Thanks for the `|n` / `&n` explanations!

I've committed a first attempt in 9713c19... It's kind of a mess, and I'm having trouble testing it. Should I mount a %sandbox desk, cp my changes there, then merge back into %home so that both files change in the same transaction?

I've copied the approach in `ohAitch/base:womb`, except if left full molds for the old state. I'm having trouble writing a gate that transforms some part of a mold inside of a collection (`++source` and `++target` are the molds i've updated).

I assume those changes in `ohAitch/base:womb` will eventually make it back into `urbit/arvo:master` -- how should overlapping state changes be handled? Should I rebase onto that branch?

Also, in `++drum-port` I have to call the molds specifically, I'm not able to cast them. I can't figure out why -- the structures seem to match ...

Sorry for all the questions and confusion :|. This state upgrade is conceptually simple, but I'm having a hell-of-a-time with it.

joemfb:
Sorry for the random pile of questions -- that's kind of a pain ... Here's some more ;)

I have it working now -- history is preserved for both :dojo and :talk. To test and refine it, I merged from a %sandbox desk so that I could apply all the changes in this PR in one transaction. I also found that it was easier to add new faces to the head of the mold, where possible.

I think my changes need to be refactored; I've included the full molds for all state versions, so it's very verbose.

And my question about `womb` still stand: if those changes are going to be merged in the future, should I adapt this branch to include the state-structure changes from that one?

joemfb:
I figured out `++target-0`; any tips for `++source-0` or `++drum-pith-0`?

ohAitch:
Were you looking at the drum-pith-0 on the womb branch? (That'll be a fun
merge)

joemfb:
Yeah, that's what my `womb` questions were in relation to ... Seems like it's inviting disaster to not serialize these changes. (My last question is just about refactoring the molds)

Y'all are already running `womb` on multiple piers, right? If so, it'd make sense for my state changes to layer on top of those. As much as I'm impatient to get this merged, it's much less important in the scheme of things ...

ohAitch:
@juped @cgyarvin, thoughts on merging womb? The relevant drum changes
should definitely make it into the mainline, though womb itself may make
sense as a separate repo. (At least the implementation, a stub would be
needed in hood similar to the current check in of %jael).

cgyarvin:
Super happy to merge all this stuff.

The great renaming breach must also include a general purge of state adapters...

juped:
Womb should have been a separate repo to begin with, though I understand why it ended up not being so. You should not combine separate work in the history; that's really not helpful to the history. Anton has some thoughts on the thorny issues that *do* arise when parallel new versions of a state arise (thorny only because you can't automerge ++prep; we are spoiled by textual three-way merge), but that doesn't quite apply yet. Those questions still need sensible answers.

Though it's off-topic for this thread, it might be necessary for the eventual integration: why is womb in hood? and why *wouldn't* womb be an eventual part of the base system, as there will be 2^16 ticketing authorities in the long term?

ohAitch:
Womb is in hood because this is the canonical workaround for gall app
autostart being broken⌥⌫ intentionally disabled; the main argument for
keeping it in the base system is in fact "we're planning to have all the
newfound star owners use it to distribute planets".

Parallel versions arose due to a feature womb /depends/ on, but which
probably should have already been merged separately(except that it's a
hack, but we don't really have a non-hack solution for |quit and |put).
What is needed here long-term is autogenerated upgrade code for "bag of
fixed-name elements" versioned nouns, perhaps using similar treap
collapsing as cores. Defaulted bunt/drop + also
autogenerated container-walking would probably handle 90% of state
upgrades, and merge cleanly as long as names remained unique.

cgyarvin:
I would put womb in base and make it its own app.

cgyarvin:
Why does womb need to be autostarted?

We could make state adapters a fair bit more convenient in a number of ways.  But we're trying to keep our features frozen...

ohAitch:
Long-term as in "when hoon is significantly unfrozen", yup. This hopefully
isn't happening at /least/ until we get back to 2013-2014 stability levels.

What was the case again for taking out auto-start, at least for local
pokes/peers? My basic reasoning here is "don't force 2^16 users to take an
extra step, unless there's a concrete scenario where the step requires user
authorization", and good defaults over interaction in general.

juped:
Cool. Womb for base 2016. This thread can return to topic now. @joemfb, just do your thing in isolation; if you had to know that parallel branches exist, that would be a major process failure. Let us know when your pull request is a pull request.

ohAitch:
+1

cgyarvin:
We took it out because we didn't understand it.  But please don't make design mistakes for a little bit of convenience.

joemfb:
One last question: should arvo be "safe" on a per-commit basis? That is should the state changes and adapters be included together in a single commit?

juped:
Yeah. Though people skip this all the time. But that's a bad idea.

joemfb:
Ok, that's done. I squashed the last 5 commits.

I've noticed an odd behavior when transitioning state from new back to old: `:talk` dumps it's backlog to the console. I'm not sure what's causing that -- I don't really understand :talk too well yet. I have confirmed that history is preserved when transitioning state from old to new.

Outside of that, I think this is done.

juped:
Talk dumps when it's unlinked and relinked, I hear. It probably shouldn't.

Thanks!

joemfb:
Let me know if you'd like this rebased or adjusted further

cgyarvin:
I continue to be slightly amazed that Joe's gone from zero to writing state
adapters in a month.

Must... resolve... docs PRs...

PR #194, opened 2016-07-05
talk@2ba76c1, tree@ca059b5

galenwp:


galenwp:
I branched from master, overwrote some js, and now I have a conflict?

juped:
It's fine, just tape a black piece of cardboard to your monitor in that spot.

juped:
(It's because this commit is based on something from June, and we've touched talk and tree in the meantime - but reasons to care about that are in very short supply.)

PR #195, opened 2016-07-07
Fix url for ;help in :talk

gkelly:


galenwp:
Ah, thanks.

galenwp:
(I tend to let @juped close these)

PR #196, opened 2016-07-07
Put robots.txt back in eyre :/

ohAitch:
Upon review, /robots.txt is actually a magic web filename, which shouldn't interact with the tree frontend etc.
This PR keep the config change from "DENY" to "ALLOW"


juped:
Don't we have, like, tree hiding magic? This seems dubious without some other way for a user to edit their robots.txt

ohAitch:
*looked around for tree hiding magic, couldn't find it*

PR #197, opened 2016-07-08
adds |syncs hood generator to print all active syncs

joemfb:


joemfb:
I first asked about this functionality in `urbit-meta` ~2 months ago ;)

In that conversation @cgyarvin said the sync state should have a `scry` namespace and a web interface  -- I don't know how to add either, so I just stuck a `poke` arm in `hood` and `kiln`. Can that be poked over the web?

ohAitch:
Messages like that used to go to your :talk porch, but "this is more than
64 characters" never really got a reasonable UI. (In general, responses
/directly to user actions/ can and should be multiple lines, which
activatable messages are a bad medium for)

For putting a scry namespace in a hood unapp, and a web interface on top of
that, I would look at the womb peek+peer arms/routing, and its frontend.
(For a scryful |syncs this would also work actually, see +womb-balances)

cgyarvin:
Super useful!  We really need a "control panel" UI to watch this...

joemfb:
@ohAitch, where can I find the server-side components of %womb?

philipcmonk:
I believe this is the most recent branch:  https://github.com/ohaitch/base/tree/womb

PR #199, opened 2016-07-12
disallow dup arms

juped:
Putting this in the issue tracker because I can't figure out why hoon.hoon nest-fails when it's integrated. (Ignore all the changes outside ++wisp, they're in error.)

### DIFF
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -778,24 +778,6 @@
   ^+  t.a
   [i.a $(a (skim t.a |=(c/_i.a !(b c i.a))))]
 ::
-++  spin
-  |*  {a/(list) b/_|=({* *} [** +<+]) c/*}
-  ::  ?<  ?=($-([_?<(?=($~ a) i.a) _c] [* _c]) b)
-  |-
-  ?~  a
-    ~
-  =+  v=(b i.a c)
-  [i=-.v t=$(a t.a, c +.v)]
-::
-++  spun
-  |*  {a/(list) b/_|=({* *} [** +<+])}
-  =|  c/_+<+.b
-  |-
-  ?~  a
-    ~
-  =+  v=(b i.a c)
-  [i=-.v t=$(a t.a, c +.v)]
-::
 ++  swag                                                ::  infix
   |*  {{a/@ b/@} c/(list)}
   (scag +<-> (slag +<-< c))
@@ -10079,7 +10061,16 @@
     ++  wisp                                            ::  core tail
       %-  ulva
       %+  cook
-        |=(a/(list {p/term q/foot}) (~(gas by *(map term foot)) a))
+        |=  a/(list {p/term q/foot})
+        =|  take/(set term)
+        =+  cur=a
+        |-
+          ?~  cur
+            (~(gas by *(map term foot) a))
+          ?:  (~(has in take) p.i.cur)
+            ~|  taken-name+p.i.cur
+            !!
+          $(take (~(put in take) p.i.cur), cur t.cur)
       (most muck boog)
     ::
     ++  toad                                            ::  untrap parser exp
--- a/arvo/dill.hoon
+++ b/arvo/dill.hoon
@@ -76,9 +76,6 @@
   $%  {$nice $~}                                        ::
       {$init p/ship}                                    ::
   ==                                                    ::
-++  sign-gall                                           ::  see %gall
-  $%  {$onto p/(unit tang)}                             ::
-  ==                                                    ::
 ++  sign-clay                                           ::
   $%  {$mere p/(each (set path) (pair term tang))}      ::
       {$note p/@tD q/tank}                              ::
--- a/arvo/zuse.hoon
+++ b/arvo/zuse.hoon
@@ -1661,13 +1661,6 @@
               heg/(map hand code)                       ::  proposed
               qim/(map hand code)                       ::  inbound
           ==                                            ::
-++  coal  *                                             ::  untyped vase
-++  code  @uvI                                          ::  symmetric key
-++  cone                                                ::  reconfiguration
-          $%  {$& p/twig}                               ::  transform
-              {$| p/(list @tas)}                        ::  alter
-          ==                                            ::
-++  chum  @uvI                                          ::  hashed passcode
 ++  claw                                                ::  startup chain
           $:  joy/(unit coal)                           ::  local context
               ran/(unit coal)                           ::  arguments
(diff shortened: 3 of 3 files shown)


ohAitch:
Yes please. Really the correct place for "noncolliding gas" is in ++by,
though ~| %taken-name would presumably still go in the wisp cook wrapper.

arvo/hoon.hoon
```diff
@@ -10079,7 +10061,16 @@
     ++  wisp                                            ::  core tail
       %-  ulva
       %+  cook
-        |=(a/(list {p/term q/foot}) (~(gas by *(map term foot)) a))
+        |=  a/(list {p/term q/foot})
+        =|  take/(set term)
+        =+  cur=a
+        |-
+          ?~  cur
+            (~(gas by *(map term foot) a))
```
joemfb: Shouldn't this be `(~(gas by *(map term foot)) a)`? Could that be the `nest-fail`?

chc4: Looks like that's correct! No clue why the nest-fail was in such a weird place due to it, though.

ohAitch: gas~by syntax 2017

joemfb: Or `(gas:by a b)` ;)

Is the `:door` (`|_`) an aesthetic thing (a la DRY), or is there some functional or performance advantage?

ohAitch: The main benefit of |_ is precisely DRY, which imho has
significant functional benefits when not overdone. I suppose there is some
minor performance improvement from keeping the core around for submethods,
but if that becomes relevant you probably want jets.

joemfb:
If #225 is approved, this PR could be reduced to something like

```
|=(a/(list {p/term q/foot})
~|  %duplicate-arm
(~(gaf by *(map term foot)) a))
```

juped:
I approve! Man, github makes a mess of this, though.

On August 1, 2016 12:42:07 PM PDT, Joe Bryan <notifications@github.com> wrote:
>If #225 is approved, this PR could be reduced to something like
>
>```
>|=(a/(list {p/term q/foot})
>~|  %duplicate-arm
>(~(gaf by *(map term foot)) a))
>```
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/urbit/arvo/pull/199#issuecomment-236685085

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

joemfb:
I implemented my suggestion here: https://github.com/joemfb/arvo/blob/disallow-dup-arms/arvo/hoon.hoon (not opening a PR yet, since it's based on #270).

It works, but it's a really crappy experience, mostly because of urbit/urbit#700. After printing a *really* long stacktrace, you get something like

```
[%duplicate-key 1.886.221.672]
%duplicate-arm
```

I tried `~_((sell vase) !!)` and other `++slog` variants, but I can't find a work-around. If none exists, probably best to skip `+-gaf:by` for now.

Also worth noting: cores defined in `:dojo` don't seem to pass through `++wisp`. This change seems to catch duplicate arms everywhere except in `:dojo` expressions.

chc4:
It would probably be worth it just to change the `~|` to also print out the arm name. It doesn't matter if `+-gaf` doesn't pretty-print if we do in another trace.

You might have to pill with the new compiler for :dojo? not sure about that

joemfb:
Yeah, it's not hard to add the appropriate trace in `++wisp`, it just makes `+-gaf:by` irrelevant -- you have to iterate in `++wisp` so that you know which arm was duplicated.

cgyarvin:
It is a crappy experience.  But...

(1), ++by doesn't have access to ++sell, because it is good to have
hoon.hoon not be a total shitshow of randomly ordered crap.

(2), one of the nice things about Hoon is that you can use anything as a
key.  But not anything can be printed in a family-friendly way.  Even if
++sell were available, deploying it there would make me slightly
uncomfortable.

I think the right way to put this in is just to not use ++gaf or ++gas in
++wisp.  It's retarded that we silently ignore duplicate arms, and there's
no shame in a little cut and paste.  Also, a version of ++put that returned
an indicator that the key was already in the map might be the best building
block here -- crashing isn't always what you want to do.

joemfb:
Thanks, that makes sense.

While you're here, can you point me towards `:dojo` expression evaluation enlightenment? I've been reading source, but it's still not clear to me how (or why) `:dojo` parses expressions without invoking `++wisp`.

cgyarvin:
I'm as puzzled as you are!  But I didn't write the tall-form parser in
:dojo.  Maybe Anton knows something about it...

joemfb:
I take it all back -- it does work! I had to sync `:dojo` first for some reason. Even after a `|reset` ...

cgyarvin:
Interesting.  The %kiln logic that makes sure a reset resets everything is
slightly flaky in my eyes...

joemfb:
I should point that "work" is relative: a core with duplicate arms crashes the `:dojo`. Is there a simple way to indicate parser failure inside of `++cook` gate?

*still needs to brush up on parser combinators...*

ohAitch:
Iirc this is waiting on gall getting a message to reload all apps upon a
zuse reset(possibly from ford, which could track the "kernel" dependency).
Ofc officially you're supposed to bump the kelvin of hoon, zuse, and
consequently all /?s…

ohAitch:
Crashes in dojo are not handled properly by the ++sole protocol. But yes,
the hoon parser should generally error instead of crashing, possibly with
a printf as happens when you type /====

cgyarvin:
Replace with ++sear :-)

joemfb:
Slightly better user experience implemented here: https://github.com/joemfb/arvo/commit/8407852a51ae67894e4b455755b5c6b34dda009d

```
:-  %say  |=  *
=<  abc
|%  ++  abc  42  ++  abc  43  --
```

```
~zod:dojo> +test
[%duplicate-arm %abc]
syntax error: 3 31
ford: fade /~zod/home/23/gen/test
ford: core /~zod/home/~2016.9.29..20.06.44..2f8d/gen/test
```

Note that the column offset is 31, right before `--`.

Typing the core is a little strange; there's inconsistency with regards to whether and when `%duplicate-arm` is printed.

In one case, the message is printed on every keystroke starting from column 27 (before `43`), but typing is not blocked until 31. In the other, the message is not printed, and typing is blocked from 27 on. I haven't figured out what distinguishes the two.

cgyarvin:
I refer you to Anton, master of the ++dojo multiline parser...

cgyarvin:
But that's still a considerable improvement!

joemfb:
After rebasing, that commit is now https://github.com/urbit/arvo/commit/f592af138c6fb4507f3452c6b8e518bfe41d947d. I'll open a PR after #270 is merged.

joemfb:
And that should've been linked on my own repo: https://github.com/joemfb/arvo/commit/f592af138c6fb4507f3452c6b8e518bfe41d947d. I have no idea why that other link works ...

ohAitch:
I don't think typing "|%  ++  a  ~   ++  a  ~  --" on a line counts as "multiline parsing", but yeah the inconsistency is weird.

juped:
I'm still not sure if this works yet.

juped:
Superseded by #293

PR #203, opened 2016-07-16
adds error message for file-not-found to |cp, |mv, and |rm

joemfb:
I was looking through old issues for low-hanging fruit, and came across #78 ... I checked `|cp` and `|mv`, and saw that they exhibited the same behavior, so I changed them too.

I pulled the `|cp` and `|mv` implementations into `kiln`, since it's simpler to print the error that way.

Also, the assertions in `|cp` and `|mv` should probably have an associated error message, but I'm not sure what it should say. Something like *"paths &lt;a&gt; and &lt;b&gt; must have the same mark/logo/extension/final-knot"*?

closes #78

PR #206, opened 2016-07-19
updates |cp and |mv to print errors without stacktraces

joemfb:
As a follow-up to #203, this prints the error messages *without* stacktraces.

Note: I'm passing a bunted `++toro` along with the error messages, so that I can use `++poke-kiln-info` in `hood`. That bunted `toro` is included in the move that's passed to `++emit`, so it ends up in `moz`, the `kiln` state. If that's a problem, I could add a `++poke-kiln-error` or somesuch gate that doesn't produce a `move`.

### DIFF
--- a/gen/hood/cp.hoon
+++ b/gen/hood/cp.hoon
@@ -6,10 +6,10 @@
 |=  {^ {input/path output/path $~} $~}
 :-  %kiln-info
 ?.  =(-:(flop input) -:(flop output))
-  ["Can't move to a different mark" *toro]
+  ["Can't move to a different mark" ~]
 =+  dir=.^(arch %cy input)
 ?~  fil.dir
   ~&  "No such file:"
-  [<input> *toro]
+  [<input> ~]
 :-  "copied"
-(foal output -:(flop input) [%atom %t ~] .^(* %cx input))    ::  XX type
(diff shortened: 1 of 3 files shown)


juped:
Since I don't actually know if that's a problem, I'll tag @cgyarvin since he could possibly answer and is reachable.

joemfb:
Thanks! The other option would be to modify `++poke-kiln-info` to take a `(unit toro)` ...

joemfb:
I went ahead and made that change in 07358c8, see 20666d1 for the previous version using a bogus bunted `++toro`

cgyarvin:
I don't know either.  We're in the territory of @pcmonk, I'm afraid.

ohAitch:
I would like to register a strong objection to not crashing on user error
because stack traces are ugly: the solution here is to fix/hide stack
traces, not return a ford/gall-level "success" on a failed operation. I
know end-to-end(delayed) acknowledgements have been broken for over half a
year, but the proper(old) response here is to abort the "Enter"
keystroke event that caused an incorrect |cp to be run, not a %talk syslog
message of "your incorrect input has been accepted and caused this
side-effect".

philipcmonk:
I approve of the `(unit toro)` version.

PR #213, opened 2016-07-25
:time app, measures time between poke and subsequent wakeup

dhaffey:
`:time expr` roughly measures the time it takes to compute `expr`. It's a very crude approximation because it includes the time it takes the kernel and gall to parse and route a value to the app and then wait for a wakeup, but it's easy enough to get a sense of that overhead (and its variation) by running `:time ~` a few times.

### DIFF
--- a/app/time.hoon
+++ b/app/time.hoon
@@ -1,0 +1,19 @@
+::
+::::  /hoon/time/app
+  ::
+/?    310
+|%
+++  card  {$wait wire @da}
+--
+|_  {bowl $~}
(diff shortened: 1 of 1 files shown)


chc4:
It took me a while to figure out why this actually works. I thought that :dojo would compute the expr first, before any of the %ford stuff happened, instead of computing the entire door+sample and *then* evaluating when pulling the arm. Fun trick using %wait like that.

juped:
I for one find this a useful base system service

PR #221, opened 2016-07-28
reject :talk ++command pokes from foreign ships

### LINKED ISSUE #208: Can impersonate other ships in talk.
%Talk can be made to send messages using %poke moves. Problem is, I can poke someone else's %talk to have them send the message for me. (At least, that's what I figure is happening here.)

Example:

```
[ost %poke /repeat/1/shipname/stationname [~zod %talk] (said our somestation %talk now eny [%leaf "i am fake ~zod"]~)]
```

Results in this line in the station the message was sent to:

```
          ~zod+ i am fake ~zod
```

In the example, `~zod` can be changed to any other ship name, and it will "impersonate" that instead.

joemfb:
closes #208 

Crashing rejects the poke, right? (As opposed to producing `[~ +>.$]`, which would accept and acknowledge?)

### DIFF
--- a/app/dojo.hoon
+++ b/app/dojo.hoon
@@ -1098,10 +1098,8 @@
   he-abet:(he-arm +<)
 ::
 ++  peer-sole
-  ~?  !=(src.hid our.hid)  [%dojo-peer-stranger ost.hid src.hid]
-  ?>  ?|  =(src.hid our.hid) 
-          &(=(%earl (clan src.hid)) =(our.hid (sein src.hid)))
-      ==
+  ~?  !=(our.hid src.hid)  [%dojo-peer-stranger ost.hid src.hid]
+  ?>  (team our.hid src.hid)
   =^  moz  .
     ?.  (~(has by hoc) ost.hid)  [~ .]
     ~&  [%dojo-peer-replaced ost.hid]
--- a/app/talk.hoon
+++ b/app/talk.hoon
@@ -2157,6 +2157,7 @@
   |=  cod/command
   ^+  [*(list move) +>]
   ::  ~&  [%talk-poke-command src.hid cod]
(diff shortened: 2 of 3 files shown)


juped:
Yeah, I think that's right. But @cgyarvin needs to provide us with a gall design sketch.

joemfb:
And I initially botched that example ... for the benefit of emailers:

```
?|  =(our.hid src.hid)
    &(?=($earl (clan src.hid)) =(our.hid (sein src.hid)))
==
```

chc4:
It probably should allow pokes from your moons. :dojo for example allows it, along with peering via sole. https://github.com/urbit/arvo/blob/master/app/dojo.hoon#L1102

There should probably be a helper function in zuse for this check, really, since it's what you usually want.

joemfb:
I added `++team` to `zuse` - seems like an appropriate name ...

To see if the crash "rejects" the poke, I added a `++coup` arm to a test app on a comet and poked `:talk` on a different comet. It was called, with a `++tang` showing the trace:

```
mack
/~tipsyt-middep-nommep-simdel--dapmyl-lapdev-bonnes-falfyr/base/0/app/talk:<[2.166 3].[2.170 24]>
/~tipsyt-middep-nommep-simdel--dapmyl-lapdev-bonnes-falfyr/base/0/app/talk:<[2.165 3].[2.170 24]>
```

ohAitch:
Naively, I would expect this to break [ra-transmit](https://github.com/urbit/arvo/blob/ab73f36/app/talk.hoon#L1568-L1571), i.e. all message posting. Does this work with like, a channel on a fakezod that a fakenec joins and writes to?

joemfb:
Sorry about that -- I opened the PR and then went mostly offline on vacation ... I saw it was merged, so I assumed it was fine

PR #222, opened 2016-07-29
adds ++key:by and ++val:by

joemfb:
before:

```
> =m (malt (limo ~[[1 2] [3 4] [5 6]]))
{[p=5 q=6] [p=1 q=2] [p=3 q=4]}
> (~(gas in *(set @)) (turn (~(tap by m)) |=({a/@ b/@} a)))
{5 1 3}
> (turn (~(tap by m)) |=({a/@ b/@} b))
~[4 2 6]
```

after:

```
> =m (malt (limo ~[[1 2] [3 4] [5 6]]))
{[p=5 q=6] [p=1 q=2] [p=3 q=4]}
> ~(key by m)
{5 1 3}
> ~(val by m)
~[4 2 6]
```

The names are overly-normal by urbit standards, but they're three letters, meaningful, and I couldn't think of urbity variants ... ;)

### DIFF
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -2823,6 +2823,17 @@
   +-  wyt                                               ::  depth of map
     |-  ^-  @
     ?~(a 0 +((add $(a l.a) $(a r.a))))
+  ::
+  +-  key                                               ::  set of keys
+    |-  ^-  (set _?>(?=(^ a) p.n.a))
+    ?~  a  ~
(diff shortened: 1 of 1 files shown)


philipcmonk:
While I definitely approve of this addition, I'd just like to point out that `|=({a/@ b/@} a)` and `|=({a/@ b/@} b)` are `++head` and `++tail`, respectively.

PR #224, opened 2016-07-29
Forum improvements

ohAitch:
deploy this before urbit/fora@160a2d0


juped:
I assume there's stuff I need to add to the headers of existing posts too?

ohAitch:
Those changes are mutually backwards compatible so I just dropped those in
with sed.

PR #225, opened 2016-07-31
adds +-gaf:by (non-colliding gas)

joemfb:
"gas, fail on dupe" -- as suggested in #199 ...

```
> =m (malt (limo ~[[1 2] [3 4] [5 6]]))
{[p=5 q=6] [p=1 q=2] [p=3 q=4]}
> (~(gas by m) (limo ~[[5 7]]))
{[p=5 q=7] [p=1 q=2] [p=3 q=4]}
> (~(gaf by m) (limo ~[[5 7]]))
ford: build failed ~[/g/~zod/use/dojo/~zod/inn/hand /g/~zod/use/hood/~zod/out/dojo/drum/phat/~zod/dojo /d //term/1]
```

### DIFF
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -2695,6 +2695,15 @@
       a
     $(b t.b, a (put p.i.b q.i.b))
   ::
+  +-  gaf                                               ::  concat, fail on dupe
+    ~/  %gaf
+    |=  b/(list _?>(?=(^ a) n.a))
+    |-  ^+  a
+    ?~  b
+      a
+    ?<  (has p.i.b)
+    $(b t.b, a (put p.i.b q.i.b))
+  ::
   +-  get                                               ::  grab value by key
     ~/  %get
     |=  b/*
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -2701,6 +2701,7 @@
     |-  ^+  a
     ?~  b
       a
+    ~|  duplicate-key+p.i.b
     ?<  (has p.i.b)
     $(b t.b, a (put p.i.b q.i.b))
   ::


joemfb:
Note: I used

```
|=  b/(list _?>(?=(^ a) n.a))
```

instead of 

```
|*  b/(list {p/* q/*})
=>  .(b `(list _?>(?=(^ a) n.a))`b)
```

which is what `+-gas:by` does. Those seem semantically equivalent ... Are they?

joemfb:
Most gates that can "fail" produce a unit (see `+-get:by`), while some have variants that crash (`+-got:by`). In many cases, the cause of the failure will be obvious, but, in this case, any one of the pairs could've caused the failure. It'd be useful to know which one...

In other langs, I'd use some type of runtime exception. 

Is there some kind of type pattern for producing error details? Something like

```
?|  {$| _?>(?=(^ a) p.n.a)}
    {$& (map _?>(?=(^ a) p.n.a) _?>(?=(^ a) q.n.a))}
==
```

Or should I just use `~|`?

```
~|  duplicate-key+p.i.b
?<  (has p.i.b)
```

ohAitch:
++each is the unit-like error format, but I'd just use ~| yeah. The
hoon parser should make a best effort to not to crash when syntax-checking,
but since this requires a value as well as a key I don't think it could be
triggered by partial input in dojo; something like the json reader should
definitely crash(duplicate keys are disallowed by spec).

cgyarvin:
@confirms ~|.

juped:
I also wrote an email saying to use ~|, but apparently didn’t press send.

joemfb:
Added `~|`:

```
> =m (malt (limo ~[[1 2] [3 4] [5 6]]))
{[p=5 q=6] [p=1 q=2] [p=3 q=4]}
> (~(gas by m) (limo ~[[5 7]]))
{[p=5 q=7] [p=1 q=2] [p=3 q=4]}
> (~(gaf by m) (limo ~[[5 7]]))
[%duplicate-key 5]
ford: build failed ~[/g/~zod/use/dojo/~zod/inn/hand /g/~zod/use/hood/~zod/out/dojo/drum/phat/~zod/dojo /d //term/1]
```

Would some formatting be preferable? Such as `~_(leaf+"duplicate key: {<p.i.b>}" ...)`, printing `duplicate key: 5`?

ohAitch:
I would go for raw ~|, the formatted version is harder to search for and
e.g. squashes p.i.b onto one line.

joemfb:
Right, I hadn't thought about complex keys ...

PR #226, opened 2016-07-31
fixes meta-. to capture to whitespace instead of by word

joemfb:
I used the wrong capture semantics for the `meta-.` keyboard shortcut ...


joemfb:
Also, contiguous invocations are supposed to move through history, but without *actually* moving through history. I.e., `meta-.`, `meta-.` should grab the last "argument" of the 2nd-to-last history entry, but a subsequent up-arrow should move from the *initial* history position. As I see it, that'll require an additional index into the history in the drum state, which will require another state transition ...

PR #227, opened 2016-08-01
urbit/tree@865ebc2 fix list component

ohAitch:


juped:
I think this build contains some of the reversions. Only some of the features are back; uptwolevel, reverse comments, and comments above comment box are not back

ohAitch:
It wasn't a revert, but rather a roll-back of the local master merge; ohAitch/fora 086f9e9 is in this PRs history, but may not be on urbit-org.

juped:
master was merged to urbit-org for deployment, and your this branch merged on top of that. the only difference between the result and your this branch is the timeout change from urbit-org.

ohAitch:
I think the problem is specifically with the "future of tree" post, which
got created with reverted code and consequently lacks the relevant headers.

juped:
Good catch!

ohAitch:
Superceded by #237

Issue #85, opened 2016-03-22
Change "backspace at start of line" to "backspace on empty line"


ohAitch:
As the trigger for "clear", which exits dialogs, multiline hoon, etc.

Fang-:
This change has been made in #235. This issue can be closed.

PR #235, opened 2016-08-08
Change "backspace at start of line" to "backspace on empty line" for clear

Fang-:
See #85. Not sure if this change is still desired, but it seemed like an easy fix, so here's a proposal.

Beeps if you backspace at the start of a line and it's not completely empty yet.

(Would it be appropriate to delete that comment (line 622) while I'm touching this part of the code?)


philipcmonk:
+1. This has driven me nuts for quite a while.

PR #240, opened 2016-08-10
remove broken sugar

philipcmonk:
#238 caused `(cury same %&)` to lose its meaning, so this desugars the instances of it in clay.hoon.

@juped, can we get this merged quickly?  Master is currently broken.


juped:
Will do, new kernel coming soon. How did it break?

philipcmonk:
As I understand it, new pills solided with master don't boot (crash in clay).

PR #243, opened 2016-08-12
use compiler cache in ++ap-slam check

ohAitch:


juped:
Unfortunately, this still introduces slowness - we're going to have to hold off on this stuff until next week

juped:
When you get a chance, can you confirm that this isn't slow so it can go back in?

ohAitch:
Will do, the slowness would be at least as noticeable as the event log one
was right?

juped:
It's supposed to be super slow, yeah

ohAitch:
Rebased on top of master and tweaked to not be performing an uncached `slot`.

PR #248, opened 2016-08-19
update :ask (|ask server) for the womb era

ohAitch:


galenwp:
Would love to deploy this ASAP.  This will get inviting back up.

ohAitch:
Specifically inviting of subs who ran |ask.

Notes:
- this should get on ~marzod so we can run  |start %ask; :ask|admins
~wisdyr-holpeg  on it
- and on ~zod so the hood/helm change ends up on new subs

Which probably just merits a full-network deploy.

PR #250, opened 2016-08-23
update fingerprint for new crypto

yebyen:
please accept my proof of identity as ~del

### DIFF
--- a/arvo/ames.hoon
+++ b/arvo/ames.hoon
@@ -179,7 +179,7 @@
         0w0                           ::   34, ~lut, Tlon Investor 9
         0w0                           ::   35, ~sep, Tlon Investor 9
         0w0                           ::   36, ~pes, Jennifer Kollmer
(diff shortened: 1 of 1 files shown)


yebyen:
I think this is not going to help me reconnect to the network unless you also breach... am I revealing too much if I say that I used the same passphrase to generate the oldkey as I have now used to generate the newkey?

ohAitch:
How not? If you were trying to connect using the new key without it being
in ames, I think all your packets would have gotten dropped, so it's as if
nobody even tried to talk to ~del. If this is pushed to everyone however,
your key will suddenly be valid. (Whether you recreate new-era ~del before
then or not)

yebyen:
Hm! I forgot that we are living in the future, where ~zod can tell every ship on the network to :reset

juped:
Apologies for the delay - some other people mentioned wanting to repope, but they never came forward and did it. Going in shortly!

yebyen:
Thanks a bunch!  I'll be around later today and I can try bringing ~del up when it's merged.

kingdonb:
Online!

PR #252, opened 2016-08-25
Absolute path for login CSS

ohAitch:


joemfb:
Should that be `/===/`, to support `|serve`'ing other desks?

galenwp:
Does `/===/` work through `%e`?

joemfb:
A quick test on a comet worked for me

juped:
I'm just merging this since it's a fix but double-secret-super-fixes are welcome

PR #256, opened 2016-08-27
serve css from the current desk

joemfb:
... instead of hardcoding home. see #252


galenwp:
Seems like a good idea to me.  @ohAitch, want to weigh in?

PR #257, opened 2016-08-31
Clay with comments!

philipcmonk:


joemfb:
What is this? This is not the urbit I signed up for!

looks awesome, btw

philipcmonk:
I'd encourage people to read this and ask questions.  I don't have time to be the sole owner of clay, so it'd be awesome if the community can take ownership of this!

juped:
This isn't in - you wouldn't expect the thing that just adds comments to cause any problems, but it contains a partial refactoring that does not compile

cgyarvin:
Not super worried about this -- I've pulled it into cc-release.

ohAitch:
Uh, does it compile in cc-release?

cgyarvin:
Strangely, it does!

PR #260, opened 2016-09-06
Use  navsort: bump  frontmatter field for posts

ohAitch:
And reorder the sidebar accordingly. urbit/tree@0565201

Technically I think this is only necessary on urbit-org? *continues to try to stay away from deploy*


ohAitch:
Old post retrofitting:

```
sed -i'' '/navuptwo/ s/^/navsort: bump\
  /' shipname/web/fora/posts/*.md
```

PR #262, opened 2016-09-08
updates ++cass and ++cuss to return a tape instead of a cord

joemfb:
While this isn't much of a win for the existing code, I think it's a big win for end-users. The "automatic `crip`" behavior is pretty counterintuitive, at least to me.

Also, `|=(a/tape (crip (cass a)))` could be `(corl crip cass)`, if not for the onerous "span check" in `++corl`.

### DIFF
--- a/arvo/eyre.hoon
+++ b/arvo/eyre.hoon
@@ -173,7 +173,7 @@
   |=  hed/(list {p/@t q/@t})  ^-  math
   %+  roll  hed
   |=  {a/{p/cord q/cord} b/math}
-  =.  p.a  (cass (trip p.a))
+  =.  p.a  (crip (cass (trip p.a)))
   (~(add ja b) p.a q.a)
 ::
 ++  fcgi                                                ::  credential caboose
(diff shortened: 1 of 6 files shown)


joemfb:
A couple more notes:

`(list ?(@c @tD))` is there for basically documentary purposes, since wet gates seem to ignore auras entirely. I want to support both `tape` and `(list @c)` -- is there a way to do that with dry gates?

These changes will allow for some cleanup in drum. I haven't included that here, since I'm prepping a drum PR, and they'd be mutually incompatible ...

ohAitch:
-1 on the type signature change, actually supporting (list @c) would
involve Unicode-awareness.

(also, wet gates have a high justification threshold, so using one for "two
specific cases" as opposed to "unboundedly nested structure" seems weird)

joemfb:
I guess I'm trying to cheat ... Since ASCII characters are the same in codepoints or "utf-8-flavored bytes", I want to be able to evaluate `(cuss (tuba "foo"))` instead of `(tuba (cuss (tufa (tuba "foo"))))` or `((list @c) (cuss (tape (tuba "foo"))))`. I'm not suggesting that a full Unicode case-conversion be provided ...

As for a dry gate with a fork, is there some way to recover and cast-to the original span? It seems like that span is thrown away, since the sample is cast.

joemfb:
That makes sense, thanks! I've dropped the fork

Issue #255, opened 2016-08-27
|unlink'ing a moon's :talk from its parent sinks the moon


joemfb:
I've done some more testing on moons impersonating their parents; some of the problems reported in #249 appear to have been exceptional, but others are trivially reproducible.

First, the good news: new moons automatically link to their parent's `:talk`. This works seamlessly across restarts on either end. A moon can simultaneously run it's own `:talk`.

The bad news: **you can't unlink a moon from its parent**.

After running `|unlink (sein our) %talk`, the moon dumps a stack trace, and then terminates. It restarts into a broken state, rejecting all input. I've reproduced this error on a handful of moons, across multiple planets. (The planets themselves appear to be unaffected).

I'll try to dig into this further, but it's a part of `drum` I don't yet understand well.

Unlinking:

```
[ ! /~your-moon/home/0/app/hood:<[106 3].[107 21]>
  ! /~your-moon/home/0/app/hood:<[107 3].[107 21]>
  ! /~your-moon/home/0/app/hood:<[107 9].[107 20]>
  ! /~your-moon/home/0/lib/drum:<[169 3].[172 20]>
  ! /~your-moon/home/0/lib/drum:<[169 16].[172 20]>
  ! /~your-moon/home/0/lib/drum:<[169 20].[169 27]>
  ! /~your-moon/home/0/lib/drum:<[489 3].[495 33]>
  ! /~your-moon/home/0/lib/drum:<[489 7].[489 18]>
  ! /~your-moon/home/0/lib/drum:<[489 11].[489 18]>
  ! /~your-moon/home/0/lib/drum:<[332 3].[335 22]>
  ! /~your-moon/home/0/lib/drum:<[333 3].[335 22]>
  ! /~your-moon/home/0/lib/drum:<[334 3].[335 22]>
  ! /~your-moon/home/0/lib/drum:<[335 3].[335 22]>
  ! /~your-moon/home/0/lib/drum:<[335 4].[335 22]>
  [%ap-lame %hood %diff-sole-effect-drum-phat]
]
[ ! /~your-moon/home/0/app/hood:<[106 3].[107 21]>
  ! /~your-moon/home/0/app/hood:<[107 3].[107 21]>
  ! /~your-moon/home/0/app/hood:<[107 9].[107 20]>
  ! /~your-moon/home/0/lib/drum:<[169 3].[172 20]>
  ! /~your-moon/home/0/lib/drum:<[169 16].[172 20]>
[...]

PR #263, opened 2016-09-09
refactors drum

joemfb:
advance apologies for the opening this massive PR without warning ...

I started refactoring as I looked into #255. I ended up blocked there, but some of my changes seemed generally useful. I kept them, and started on a long-overdue cleanup of `++ta-met` and related gates. I've tried to keep my commits separate, but I'm happy to cherry-pick across multiple branches if that'd be easier to review.

This PR covers several areas:

- restoring lost PRs

restores the fixes from #226 and #235 (although not their exact implementations), which were inadvertently removed in 786bce0

- dead code removal

everything related to `++master`, `++ukase`, and several `++se-*` arms were completely unused

- type safety

favoring `?=` over `.=`, avoiding `-.limb`, especially when pattern-matching a kelp, casting judiciously, etc.

- readability

a face for `++bowl`, bracketed moves, whitespace, etc.

- a long overdue refactoring of `++ta-met`

many cases were overly complex or contained redundancies

- comments

### DIFF
--- a/lib/drum.hoon
+++ b/lib/drum.hoon
@@ -516,31 +516,6 @@
     =+  pre=(dec pos.inp)
     (ta-hom %del pre)
   ::
-  ++  ta-kil                                          ::  build kil
-    |=  {a/?($l $r) b/(list @c)}
-    ^-  kill
-    =+  max=|=(a/(list (list @c)) (scag max.kil a))
-    ?.  ?&  ?=(^ p.blt)
-            ?|  ?=({$ctl p/?($k $u $w)} u.p.blt)
-                ?=({$met p/?($d $bac)} u.p.blt)
-        ==  ==
-      %=  kil
-        num  +(num.kil)
-        pos  +(num.kil)
-        old  (max [b old.kil])
-      ==
-    %=  kil
-      pos  num.kil
-      old  ?~  old.kil
-             [b]~
-           %-  max
-           :_  t.old.kil
-           ?-  a
-             $l  (welp b i.old.kil)
-             $r  (welp i.old.kil b)
-           ==
-    ==
-  ::
   ++  ta-ctl                                          ::  hear control
     |=  key/@ud
     ^+  +>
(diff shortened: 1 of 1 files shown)


joemfb:
There's some additional changes I'd like to make -- these would require state changes:

- stable drum rotation order

currently unpredictable, since `eel` is a set... A stable order in a list would be preferable, IMO

- meta-dot history rotation

but without actually adjusting `pos.hit`, so it'll require an additional history index

And, with regards to the talk-on-a-moon problems that got met started, the connection state-machine in `++se-abet` and related arms seems overly complicated as implemented. I looked into simplifying it, but quickly realized that I didn't understand the use-case behind `bin`, the map of `bone` to `source`. I wasn't able to find a situation in which `bin` had more than one entry. Is that a holdover from a previous drum architecture, future-proofing, or am I just stupid?

Another way to ask the question: `sys` stores the first `bone` from the bowl. Under what circumstances will the bowl have a bone != sys?

Finally, I have two comment-related questions:

- re `++peer`: dill peers with drum during initial startup. Are there other cases where this arm would be pulled?
- re `++quit-phat`: I've seen `%drum-quit` logs on various piers, but I haven't been able to trigger them myself. When is this arm pulled?

juped:
Thanks for catching those missing things. I didn't actually make those changes in the evil merge, but I bear all the responsibility for not reading them closely.

juped:
(This is completed work, right? Github culture sucks.)

joemfb:
Yeah, it's completed, just awkwardly presented since it's a big ask ...

If the dead-code removal is copacetic, that could all be one commit. Other commits should similarly be squashed, but are separate now in case they need to be backed out or cherry-picked. At there very least, some need better messages.

I'd also like to ask questions and discuss future changes... Would that be more manageable in a forum post?

juped:
Please feel free! Also feel free to just rewrite everything, if that's what you feel like doing.

cgyarvin:
Yay!  Nothing I love more than a really ginormous PR ambush.  It's my own style... forum post would be ideal!  Sorry to sound like a teenage girl.

PR #265, opened 2016-09-21
urbit/tree@f726ff4 add null check on shortShip

ohAitch:


galenwp:
Would like to get this out.

PR #272, opened 2016-09-29
removes +-gaf:by

joemfb:
The current `+-gaf:by` implementation prints `%duplicate-key` for every recursive invocation prior to detecting a duplicate.

There's still a bigger problem, which is that wet arms can't `++slog` in a type-aware manner (see urbit/urbit#700). Given that, and the inapplicability of `+-gaf:by` to #199, is it worth keeping around? Should I instead just remove the arm?

### DIFF
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -2731,16 +2731,6 @@
       a
     $(b t.b, a (put p.i.b q.i.b))
   ::
-  +-  gaf                                               ::  concat, fail on dupe
-    ~/  %gaf
-    |=  b/(list _?>(?=(^ a) n.a))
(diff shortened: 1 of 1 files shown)


cgyarvin:
Alas, I do think remove it.  We need to take a more sustained and serious look at what our containers should do...

joemfb:
removed

ohAitch:
I'd stick it in the JSON parser probably, I think that's where I first wanted `gaf` to exist

ohAitch:
For that matter, is there a compelling reason for `my` and friends not to use it? Seems like a sane defaults thing.

cgyarvin:
Can't think of one...

Issue #266, opened 2016-09-23
key:by breaks set invariant


ohAitch:
`++key` in `by` should use `put:in` instead of trying to be clever with preserving structure.

```
~pittyp-datfyn# ~(key by (my [/a 1] [/b 2] [/c 3] ~))  {[%c ~] [%a ~] [%b ~]}
~pittyp-datfyn# (sy /a /b /c ~)  {[%a ~] [%c ~] [%b ~]}
~pittyp-datfyn= one of these things is not like the other
~pittyp-datfyn# (~(put in ~(key by (my [/a 1] [/b 2] [/c 3] ~))) /a)  {[%a ~] [%c ~] [%a ~] [%b ~]}
~pittyp-datfyn# (~(has in ~(key by (my [/a 1] [/b 2] [/c 3] ~))) /a)  %.n
```

joemfb:
It's worth noting that `++apt` doesn't catch this. The right branch noun-order conditions* look suspicious to me:

```
?&((vor n.a n.r.a) (hor n.a n.r.a) $(a r.a))
```

cgyarvin:
I'm not sure what seems suspicious to you?  n must be above l and r
vertically, between n and r horizontally?

joemfb:
I didn't realize "h-order" meant "horizontal" -- I just mentioned the only part of `++apt` that looked (maybe) off ...

Am I correct in assuming that `++apt` should only produce `&` for valid sets? It works correctly here:

```
> (apt (sy 1 3 2 ~))
%.y
> (apt [1 [2 ~ ~] [3 ~ ~]])
%.n
```

But not here:

```
> (apt (sy /a /b /c ~))
%.y
> (apt ~(key by (my [/a 1] [/b 2] [/c 3] ~)))
%.y
```

cgyarvin:
Are you sure that isn't just a coincidence? :-)

joemfb:
I'm not sure what that means -- as distinct from a bug ;)

Is `++apt` just roughly documentary, or should it have the same perspective as `+-put:in`?

```
> ({n/* l/* r/*} (sy /a /b /c ~))
[n=[99 0] l=[[97 0] 0 0] r=[[98 0] 0 0]]
```

`(apt (sy /a /b /c ~))` reduces to the following comparisons:

```
(vor [99 0] [97 0])
(hor [97 0] [99 0])
(vor [99 0] [98 0])
(hor [99 0] [98 0])
```

```
[...]

PR #275, opened 2016-10-05
updates ++apt to transitively verify set constraints

joemfb:
I'm not sure that this is the most efficient approach. It maintains lists of ancestor nodes: `l` for ancestors whose left branch we descended, `r` for right. (there's a case to be made for reversing those names ...)

In any case, using the broken `+-key:by` implementation:

```
:+  (apt (sy /a /b /c ~))
    (apt ~(key by (my [/a 1] [/b 2] [/c 3] ~)))
    (apt (sy /a /b /c /d /e /f ~))
```

produces `[%.y %.n %.y]`

re #266

### DIFF
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -2388,10 +2388,13 @@
 ::
 ++  apt                                                 ::  set invariant
   |=  a/(tree)
-  ?~  a
-    &
-  ?&  ?~(l.a & ?&((vor n.a n.l.a) (hor n.l.a n.a) $(a l.a)))
-      ?~(r.a & ?&((vor n.a n.r.a) (hor n.a n.r.a) $(a r.a)))
+  =|  {l/(unit) r/(unit)}
+  |-  ^-  ?
+  ?~  a   &
+  ?&  ?~(l & (hor n.a u.l))
+      ?~(r & (hor u.r n.a))
+      ?~(l.a & ?&((vor n.a n.l.a) $(a l.a, l `n.a)))
+      ?~(r.a & ?&((vor n.a n.r.a) $(a r.a, r `n.a)))
   ==
 ::
 ++  in                                                  ::  set engine
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -2586,11 +2586,14 @@
 ::                section 2dB, maps                     ::
 ::
 ++  ept                                                 ::  map invariant
-  |=  a/(tree {p/* q/*})
-  ?~  a
-    &
-  ?&  ?~(l.a & ?&((vor p.n.a p.n.l.a) (gor p.n.l.a p.n.a) $(a l.a)))
-      ?~(r.a & ?&((vor p.n.a p.n.r.a) (gor p.n.a p.n.r.a) $(a l.a)))
+  |=  a/(tree (pair))
+  =|  {l/(unit) r/(unit)}
+  |-  ^-  ?
+  ?~  a   &
+  ?&  ?~(l & (gor p.n.a u.l))
+      ?~(r & (gor u.r p.n.a))
+      ?~(l.a & ?&((vor p.n.a p.n.l.a) $(a l.a, l `p.n.a)))
+      ?~(r.a & ?&((vor p.n.a p.n.r.a) $(a r.a, r `p.n.a)))
   ==
 ::
 ++  ja                                                  ::  jar engine


joemfb:
Any objections to replacing

```
?=($~ (skip l |=(b/* (hor n.a b))))
?=($~ (skip r |=(b/* (hor b n.a))))
```

with

```
?=($~ (skip l (cury hor n.a)))
?=($~ (skip r (curr hor n.a)))
```

?

joemfb:
I've updated this and removed the duplicated `++hor` comparisons. I suspect it can be made more efficient -- maybe making `l` and `r` a 2-deep window into "ancestors whose left/right branch we descended", instead of keeping the full list. I've my mind's right, that should transitively verify the order.

(it works for my broken example, but so does an even simpler approach with units instead of lists. I think the units work as a quirk of the broken set structure, but 2-deep lists seem to generalize correctly).

Thoughts? And style-tips re: to curry, or not?

/cc @juped @cgyarvin @max19

juped:
I like the use of `cury`/`curr`. Some might call me a demented Haskeller for it.

I don't have an algorithms textbook handy, but the approach I'd take is just carrying lower and upper bounds into each recursive invocation. Ignoring our noun orders for the moment and just considering a binary search tree of integers, if the root is 5, this imposes an "upper bound: 5" constraint on the entire left subtree. So push the upper and lower bound nouns on the subject as `(unit *)`, put the rest of it into a `|-`, and update the upper and lower bounds as needed in your `%=` recursions. Make sense? My explanation seems a bit jumbled to me...

juped:
Ah - and whatever you do, could you paste it over in the map invariant as well? :)

joemfb:
Your explanation simplifies the problem for me, thanks! (And I accept and deserve the "algorithms textbook" dig ... ;))

What you're describing is the simpler, unit-based implementation I mentioned. It appears to verify the same constraints as the old, broken implementation -- the difference is that it verifies them for every `n`, instead of on every branch.

I can replicate this approach for `++ept`, but, don't we want a broken example to test, too? I haven't been able to create one ...

cgyarvin:
I hate cury/curr, if only because they're misspelled.  Well, not only
because they're misspelled.  But I recognize that that's just a personal
quirk.

juped:
No dig there, just feeling the absence of a copy of Okasaki's book. His doctoral thesis just isn't up to the same standard...

cgyarvin:
I'd also go all the way and actually make the (set) and (map) molds call
++apt and ++ept.

joemfb:
Would that go in `++whip:al`? The `$herb` branch, perhaps?

juped:
Dependent molds are a wonderful frontier that should be explored. However, we can't break `hard` and `soft` without serious thought. They would be required to crash on failure.

It has nothing to do with code generation. `set` and `map` are just gates that produce gates, you can write them as manually as you'd like. The `$` rune family is a convenience, nothing more.

juped:
e.g.
```
++  set
  |*  a/mold
  |=  n/*
  =/  s
    ((hard (tree a)) n)
  ?>  (apt s)  s
```

I have merely written this off the top of my head, I have not attempted to check it for validity or workingness. But that's the idea.

cgyarvin:
What Ray says is quite true, of course, and makes me excited about writing
Urbyte 3 where I actually explain this.

I'd add that ++set should cast its product to the thing it does now.  That
way, when the compiler infers through it -- which is of course the main use
of a mold -- it doesn't have to think about ++apt.

ohAitch:
Seconded on "don't break soft"

What about

++  set
  |*  a/_*
  %+  cork  (tree a)
  |=  b/(tree a)  ^+  b
  ?.  (apt b)  ~
  b

cgyarvin:
Yeah, sure, don't break ++soft.

joemfb:
So I came up with this:

```
++  map  |*  {a/mold b/mold}                            ::  associative tree
         %+  corl  (tree (pair a b))                    ::
         |=  b/(tree (pair a b))  ^+  b                 ::
         ?:((ept b) b ~)                                ::
++  set  |*  a/mold                                     ::  set
         %+  corl  (tree a)                             ::
         |=  b/(tree a)  ^+  b                          ::
         ?:((apt b) b ~)                                ::
```

Worked fine in a generator, so I was feeling pretty good until a synced hoon.hoon.

**find.corl**

So, I did some tweaking, and came up with this:

```
++  map  |*  {a/mold b/mold}                            ::  associative tree
         |=  n/*
         %.  ((tree (pair a b)) n)                      ::
         |=  b/(tree (pair a b))  ^+  b                 ::
         ?:((ept b) b ~)                                ::
++  set  |*  a/mold                                     ::  set
         |=  n/*
         %.  ((tree a) n)                               ::
         |=  b/(tree a)  ^+  b                          ::
         ?:((apt b) b ~)                                ::
```

but, Alas! **find.ept**

Thoughts?

joemfb:
Should `++map`, `++set`, `++jar`, and `++jug` move to Section 2cI? Is there another option?

juped:
Molds belong in their own core prior to all others.

joemfb:
And yet, find.ept

juped:
This means that I think apt/ept can just be inlined. Or `|^`ed, more realistically.

joemfb:
Along with `++vor`, `++hor`, and `++gor`?

juped:
Good point! It's dangerous. Hmm.

cgyarvin:
Bear in mind that I've completely reorganized hoon.hoon on the relearning
branch.  This means I have to redo all patches manually.  Look at that
hoon.hoon if you're interested in how things should actually be ordered.

Yes, I'd just put the apt/ept code in ++set/map.  ++vor, ++hor and ++gor
are used elsewhere, sometimes...

ohAitch:
Random hack: rebind them in ++ zuse as (cork (^map a) ...), leave
validation out of hoon.hoon?

joemfb:
Right, the noun-orders are used heavily, so they'd have to become arms in the mold core, or (ick) have their implementations duplicated (along with parallel jet registration?).

joemfb:
In particular place in %zuse?

joemfb:
I just looked at relearning; since molds are no longer in a separate core there, do you just want to delay calling apt/ept from set/map?

Also, I've been opening lots of little kernel PRs -- is there something I could do to make the porting easier?

ohAitch:
Huh, I thought they had to be for bunting performance reasons?

cgyarvin:
(1) I honestly don't think apt/ept need to exist outside set/map.  When
else would we check the invariant?

(2) I appreciate the offer, though I'm super cautious about dumping
administrivia on contributors.  I mean, golly gee, you're supposed to be
here for fun.  I'm going to move relearning to a branch of urbit/arvo
proper, so you can certainly give us PRs for that...

belisarius222:
Related to https://github.com/urbit/docs/pull/281

belisarius222:
Nevermind; disregard previous comment.

ohAitch:
not *un*related, if we want to document the vertical orders as equivalent they should actually be changed to be equivalent :3

PR #277, opened 2016-10-11
updates ++snag to cast the product

joemfb:
as pointed out by @ohAitch on `urbit-meta` ...

### DIFF
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -798,7 +798,7 @@
 ++  snag                                                ::  index
   ~/  %snag
   |*  {a/@ b/(list)}
-  |-
+  |-  ^+  ?>(?=(^ b) i.b)
   ?~  b
     ~|('snag-fail' !!)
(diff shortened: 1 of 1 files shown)


juped:
Why the `|-`s anyway?

ohAitch:
I think otherwise the $(a t.a) tends to break?

joemfb:
I've dropped the `|-` -- I agree that it's unnecessary.

There's lots of other superfluous loops in other list processing gates ... Old code, or is there some reason to not recur into a wet arm?

cgyarvin:
I should have weighed in earlier.  In principle there's nothing wrong with wet recursion, but it's quite easy to change the core's type and cause weird shit to happen.  Sometimes you even want the loop to expand nontrivially.  But usually you don't, and drying it out shows that clearly...

joemfb:
Ok, thanks for the details! I've added it back.

cgyarvin:
Thanks!

PR #284, opened 2016-10-21
Fix presence, better this time

ohAitch:
Single-machine testing caught the loop previously experienced, and uncovers no issues with this version; multi-machine testing tbd. ~should~ be fine but I thought that last time P:

### DIFF
--- a/app/talk.hoon
+++ b/app/talk.hoon
@@ -138,6 +138,46 @@
         ==
       --
     --
+|%
+::  old protocol workaround door
+++  timed
+  |_  a/(map partner atlas) :: XX (map partner (pair @da atlas))
+  ++  strip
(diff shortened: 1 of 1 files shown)


juped:
What's "single-machine testing"? Fake federated stars?

ohAitch:
And planets, yeah.

cgyarvin:
Yes.

juped:
This is in

ohAitch:
I am somewhat alarmed that it will bring down the network again and we
won't know if that's actually its fault, but I suppose I really haven't
been making enough progress on the testing infrastructure to hold this off
forever

juped:
It never brought down the network to begin with, just spammed people.

PR #290, opened 2016-10-26
reapplies ++cass change too lib/elem-to-react (unbreak the web)

joemfb:
this change from #262 that was lost in the merge of #287


juped:
Thanks! This is both why VCSes should track renames, and why I should pay more attention when something is renamed.

ohAitch:
Really it's a type signature change, but a rename keeping the old code
would have solved the problem nicely, yes

ohAitch:
A VCS-mediated rename that is, presumably we'd still want to call the thing
cass

PR #291, opened 2016-10-26
fixes mar/sole/effect, purging styles on json:grow

joemfb:
this fixes the compile bug, and converts `++styx` to `++tape` for `$pro` effects

### DIFF
--- a/mar/sole/effect.hoon
+++ b/mar/sole/effect.hoon
@@ -31,6 +31,11 @@
   =+  rolt=|=(a/wall `tape`?~(a ~ ?~(t.a i.a :(weld i.a "\0a" $(a t.a)))))
   (rolt (turn (flop tan) |=(a/tank (rolt (wash 0^wid a)))))
 ::
+++  purge                                               ::  discard ++styx style
+  |=  a/styx  ^-  tape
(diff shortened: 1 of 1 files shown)


juped:
Thanks!

joemfb:
confirmed that this also fixes webdojo

ohAitch:
@cgyarvin any objections if I move `++purge` to zuse after various `cc-` merges are resolved? Seems like it'll be useful when incrementally porting talk and dojo to allow `styx` output.

cgyarvin:
I think all the style stuff including klr should go there.  Might lengthen some names too...

joemfb:
I'd like to see the `++klr` engine in drum moved as well (it intentionally depends only on the gate samples, not the subject or core context for this reason).

In the long run, I figure it'll all be folded into a `++tanq` engine ...

ohAitch:
The conversion from styx to the "flat" dill-blit ++stub, sure

ohAitch:
Oh, that is totally the namespace it should be under yeah. (I'm still a bit
weird about calling non-door cores "engines", the things already described
as such were like ++og and ++by)

PR #293, opened 2016-10-26
disallows duplicate arms in ++wisp

joemfb:
added this now that #270 has been merged -- see #199 for discussion.

closes #199

### DIFF
--- a/arvo/hoon.hoon
+++ b/arvo/hoon.hoon
@@ -10183,8 +10183,14 @@
     ::
     ++  wisp                                            ::  core tail
       %-  ulva
-      %+  cook
-        |=(a/(list {p/term q/foot}) (~(gas by *(map term foot)) a))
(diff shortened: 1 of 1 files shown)


juped:
But does it work? (Serious question, the status of this has been murky to me.)

joemfb:
Yes. As demonstrated in #199, it works in generators and `:dojo` expressions (although you have to manually for a "reload" of `:dojo` first). I just tested it in apps and vanes, it works there as well.

joemfb:
The UX could definitely be improved: from more sophisticated error logging (currently `[%duplicate-arm %foo]`) to earlier detection when parsing (currently errors at the arm *after* the duplicate arm, or the end of the core, whichever comes first)

juped:
Cool. Expect it Friday.

PR #301, opened 2016-11-08
Hack around a broken-mark subscription edge case

ohAitch:
Requires %zuse and %gall reload, which should succeed against master@219e587

@cgyarvin please review and acknowledge.


cgyarvin:
Reviewed.

ohAitch:
Aaaaa i left printfs in abort abort
(don't abort but like, mo-defer and mo-resume on every talk message should maybe get taken out)

juped:
Do you have a patch that removes them?

PR #303, opened 2016-11-09
add /===/mar and /===/arvo to %core req dependends

ohAitch:
This will cause many false positives (all apps reloaded on any mark change), but @cgyarvin says it's better than them drifting out of sync as they currently do.


juped:
Can this deploy safely?

ohAitch:
Should be, it might uncover bugs from apps being reloaded all the time, but no specific problems I would expect.

PR #305, opened 2016-11-15
urbit/tree@4d2e9c9

galenwp:
Might have mangled the history here.  Sorry.


juped:
Could you replace this with two new PRs updating tree, one based on `master` and one based on `maint-20160818`?

galenwp:
My default approach will be to just create PRs with fresh `.js`.  Sound good?

juped:
I think so? What does that mean, exactly?

galenwp:
Ah, don't worry about it.  Pretty sure that's right.

PR #313, opened 2016-12-08
%xmas compiles and integrates with %jael.

cgyarvin:


juped:
Why are there two of this?

cgyarvin:
Not sure how to delete the old PR.  I replaced it with one that doesn't
delete arvo.

juped:
You're assuming that Github PRs were designed by a sane person for sane people. They were in fact designed by the insane for the insane, and update themselves dynamically.

PR #318, opened 2016-12-09
Talk setting: notify on mention

Fang-:
Added the `%notify` flag. Having called `;set notify`, the terminal beeps whenever a message contains the first part (`~palfun`) of your ship name.  
Magically, it doesn't beep for your own messages.

Contains the telegram renderer core modification that's also included in #308.

Would love a code review on this (lines 440-449). I saw vaguely similar constructions in the rest of the code, but I'm still unsure if this is "good" Hoon or over-engineering. Let me know!

### DIFF
--- a/app/talk.hoon
+++ b/app/talk.hoon
@@ -436,8 +436,8 @@
     ::
     ++  sh-rend                                         ::  print on one line
       |=  gam/telegram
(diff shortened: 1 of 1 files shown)


juped:
What about it? The use of `%.`? I think it's alright.

PR #319, opened 2016-12-12
port |autoload forward for sys/

ohAitch:
Plus other misc cleanups, this is old zuseflat-pretty history


juped:
I'll allow it because it's ported from old history but pls no random cleanups mixed with other things in the future

ohAitch:
Sorry, I have a very selfish pain-avoidant reflex of whenever important
steps / invariants of my workflow are broken by upstream, I scramble to fix
them, which is necessarily* something I can only do as I hit the problems.
The proper way to deal with this is separate branches, but afaict git
doesn't have a usable patchset mode(work on the octopus merge of these
n heads, with one "foregrounded" as a default commit target), and if it did
this would probably still break "every commit is a working snapshot"
assurances.

* I suppose the actual answer to this would be automated tests and refusing
to make functionality changes to code that has regressions, but as an
automated testing ~fanboy~ I don't actually have the /skill/ to set this up
properly, just a heartfelt yearning for a less broken world.

PR #328, opened 2017-01-04
adds scry-desks feature to %clay (care %d)

joemfb:
This is just a first pass -- here's what it looks like:

```
> .^((set desk) %cd /===)
{%base %kids %home}
```

Some questions:

- Should there be an arm for `(set desk)` to make the `.^` simpler?
- Should it produce something other than a set?
- And what should the mark in the cage be?
- Is there any reason that `++take-foreign-update` can't just ignore `%d`?

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -44,12 +44,13 @@
 ::
 ::  Type of request.
 ::
-::  %u checks for existence, %v produces a ++dome of all desk data, %w with a
-::  time or label case gets the aeon at that case, %w with a number case is not
-::  recommended, %x gets file contents, %y gets a directory listing, and %z gets
-::  a recursive hash of the file contents and children.
+::  %d produces a set of desks, %u checks for existence, %v produces a ++dome of
+::  all desk data, %w with a time or label case gets the aeon at that case, %w
+::  with a number case is not recommended, %x gets file contents, %y gets a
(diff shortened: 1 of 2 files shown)


joemfb:
The change to `++care` messed up kiln:

```
%ap-slam-mismatch
{/ p/?({$~ u/{p/{p/?($w $y $u $x $d $z $v) q/?({$tas p/@tas} {$da p/@da} {$ud p/@ud}) r/@tas} q// r/{p/@tas q/{#t q/*}}}} $~)}
{way// rot/?({$~ u/{p/{p/?($w $y $u $x $z $v) q/?({$tas p/@tas} {$da p/@da} {$ud p/@ud}) r/@tas} q// r/{p/@tas q/{#t q/*}}}} $~)}
[! gall: %hood: call: %writ-kiln-autoload: type mismatch  [%ap-lame %hood %writ]]
```

I think kiln will need a state adapter to handle the `riff` in its `$warp` card, but there might be other molds too -- that mismatched span is non-obvious to me.

cgyarvin:
1. No.
2. No.
3. %noun is fine for now.
4. No. :-)

My quality concerns on code deemed legacy are relatively minor, as you see :-)

cgyarvin:
It's the development branch, there's no continuity....

joemfb:
Sounds good! I don't think I have anything else...

PR #329, opened 2017-01-07
updates path-with-placeholder parsing for more consistent namespace dereferences

joemfb:
It's common to write `:wish` / `.^` twigs with 3 sub-twigs: mold, vane/care, path. But they're actually just two twigs. The vane/care is the first segment of an Urbit-namespace path. This works fine for "fully-qualified" paths, but errors with `=` placeholders in the path, which is a little confusing. So I've updated `++posh:vast` to support both uses:

```
> .^(@ %cw /===)
1
```

```
> .^(@ /cw/===)
1
```

Previously, that second example would fail to parse, and log `%posh-fail`.

My solution is kind-of a hack: I'm just checking if the first path segment is a `%sand` atom with `%ta` or `%tas` as the aura, and a width of 2 bytes. I'm not sure if there's a better solution, but it seems worthwhile to remove the confusion.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -8440,10 +8440,19 @@
     |=  {pre/(unit tyke) pof/(unit {p/@ud q/tyke})}
     ^-  (unit (list twig))
     =-  ?^(- - ~&(%posh-fail -))
-    =/  wom/(list twig)  (turn wer |=(a/@ta [%sand %ta a]))
     %+  biff
-      ?~  pre  `u=wom
-      %+  bind  (poon wom u.pre)
+      =/  wom/(list twig)
+        (turn wer |=(a/@ta [%sand %ta a]))
+      ?~  pre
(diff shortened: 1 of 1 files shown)


ohAitch:
-1, overloading path interpolation feels too magic. .^(@ cw+/===) is the
canonical idiom I think?

cgyarvin:
I'm with Anton -- this one feels a little too magic.

(By "canonical idiom" Anton of course means "this thing I came up with that
everyone should use."  In this case I think he's right...)

ohAitch:
"I'm sure Curtis has seen it by now and he hasn't complained that I
remember", tbqh

joemfb:
Fair enough I was exploring the namespace and the dichotomy caught my I. It's good practice with the parser.

I have a bunch of tangential namespace/path questions -- I'll try to write a forum post soon.

PR #338, opened 2017-02-01
Implement doccords in hoon.

eglaysher:
This adds syntax to the language to embed documentation, and modifies
`?` in :dojo so that these 'doccords' can be read.


ohAitch:
… `=*` is incredibly confusing control-flow wise, making it impossible to read in error messages; please don't mandate its use for anything non-static.

cgyarvin:
Wut?

eglaysher:
- Moved ++dqut and ++wqut to ++vast. Experimented with moving ++qut and its helper gates, but it's used by talk.

- Replaced several =+ with =*.

- Do you have another alternative other than %dock? That name is already used as an arm name in hoon.hoon. (It's `(pair @p term)`.) Or if you're insistent on %dock, do you have an idea for a replacement name for the current ++dock? I suppose that I could use %help, but I don't like having twigs and spans have the same name. (I did so during an early draft of this patch and got really confused.)

- Commented dy-show-span using doccords. (I plan on documenting hoon.hoon with doccords, but in a future patch to solve the chicken/egg issue.) One thing I noticed is that I have to include empty space on otherwise blank lines for iny to work; I'm unsure about how to fix that.

cgyarvin:
One needs to be careful with that chicken-egg stuff!  Urbit works well
enough that sometimes you forget it's impossible.  But actually, it is.

What about %help?

juped:
Just pretend docs is short for docstring (which it is).

Two commits suffice to stage hoon.hoon docstrings, if you have some waiting in the wings.

cgyarvin:
Sorry, man, this particular pretend doesn't work for me!  Abbreviations
aren't generally part of our vocabulary.

This will be a great test of staging...

eglaysher:
Used %help. I'm not crazy about it, but I suppose there is precedent for using the same tag on twigs and spans.

cgyarvin:
I share your unease.  But we do have a static type system...

eglaysher:
Are there any further changes I need to make on this patch?

cgyarvin:
Not according to me!  Sorry about the choppy review process.  It's been a
strange week...

juped:
No. Closing because it's superseded by #347, though.

PR #339, opened 2017-02-01
Fixed urb.util.isURL to allow multiple successive hyphens in URLs.

Fang-:
Posting links to `comet--name.urbit.org` via webtalk causes it to post as a regular line, possibly breaking up. Turns out this is because of the `--`, which urb.util.isURL doesn't recognize as valid.

This change allows any number of successive hyphens. I couldn't find any evidence this isn't allowed in either subdomains or domain names (but maybe I'm looking in the wrong places?).


Fang-:
*Poke.* Any further comments, or can this be merged?

juped:
Nope, it's good.

PR #343, opened 2017-02-03
replaces $cork (|^) with an actually hygienic expansion

joemfb:
This works without any problems on a fake-zod -- is there any reason that it's not a desirable change?

closes urbit/urbit#616


cgyarvin:
"Hygienic" is a little strong here.  The existing |^ is hygienic as well.

I'm not a *huge* fan of |^ in general.  But it scares me slightly to change
semantics.  And if what you changed |^ to is all that |^ does, do we need
|^?  Discuss...

joemfb:
I don't think it's a strong claim -- the current implementation clobbers an extant `$` arm ... I guess "unhygienic" is imprecise, but it's certainly a foot-gun.

I'm not at all attached to the twig itself: `=>` or `=<` with `|%` seems plenty readable to me ...

joemfb:
It's a little easier to just change this, but I only see ~ 30 instances of `|^`, so it'd be easy enough to remove altogether.

cgyarvin:
"Hygienic" is accurate in the English sense of the word, but not quite in
the Lisp sense of the word.

It is a small foot-gun.  The thing is, I'm kind of not sure what to do
about it at the moment, and my general inclination is to keep the language
relatively frozen, especially when our automated testing is as weak as it
is.  I would like to get to the point where we can revise these things in a
more systematic way.

joemfb:
Fair enough, I'm just prowling for easy wins ...

ohAitch:
It's hygienic in spirit, `$` is a special arm name that a bunch of other `|` runes also implicitly create. That said, only 30 uses* sounds like an entirely legitimate reason to either remove it outright, or convert a bunch of existing `=< |%` to it instead(which wouldn't break anything, it's just slightly awkward when the $ content is itself a gate). I think last time this came up breaking user code was a concern, but now that we have a separate `master` with significant zuse restructuring the change would have little further effect.

*and little prototyping/learning value, which is why it's important to keep e.g. `!;` around despite little representation in checked-in code

cgyarvin:
Just noting that we still haven't decided what to do about this one.

joemfb:
A less disruptive option would be to use this instead of `++expr`:

```
|.  ;~  gunk
  loaf
  %+  sear
    |=  a/(map term foot)
    ?:((~(has by a) %$) ~ `a)
  wisp
==
```

cgyarvin:
I endorse this prospective patch. :-)  It doesn't change the long-term
debate, it just fixes the immediate glitch.

ohAitch:
This doesn't really feel like a syntax error, but I suppose either way it's
a corner case.

juped:
Closing as superseded by #364.

PR #347, opened 2017-02-08
Doccords on layer one

eglaysher:
This second request is more to gauge what people think about the style of the included documentation, before I go through the entire current docs, clean them up and apply them to `hoon.hoon`. The proposed style is a one line summary for the core overview, followed by any number of paragraphs of details.

In practice, this looks like this from the user's point of view:

```
> ? +2047
layer one: basic arithmetic, tree addressing and molds

Arms:
  ++ache : Either `a` or `b`, defaulting to `b`.
  ++add  : Unsigned addition.
  ++bloq : Blocksize
  ++cap  : Tree head.
  ++dec  : Unsigned decrement.
  ++div  : Unsigned division.
  ++dvr  : Unsigned division with remainder.
  ++each : Either `a` or `b`, defaulting to `a`.
  ++foo 
  ++gate : Generic mold matching any gate.
  ++gte  : Unsigned greater-than/equal.
  ++gth  : Unsigned greater-than.
  ++list : Null terminated list.
  ++lone : One-tuple mold generator.
  ++lte  : Unsigned less-than/equal.
  ++lth  : Unsigned less-than.
  ++mas  : Axis within head/tail?
  ++max  : Unsigned maximum.
  ++min  : Unsigned minimum.
  ++mod  : Unsigned modulus.
  ++mold : A gate which normalizes data.
  ++mul  : Unsigned multiply.
  ++pair : Two-tuple mold generator.
  ++peg  : Axis within axis.
  ++pole : Faceless null terminated list.
  ++qual : Four-tuple mold generator.
  ++quid
  ++quip
  ++sub  : Unsigned subtraction.
  ++trap : Producer generator.
  ++tree : Tree mold generator.
  ++trel : Three-tuple mold generator.
  ++unit : Maybe mold generator.

<33.pdv 2.wxc $149>
> ? tree
Tree mold generator.

A `++tree` can be empty, or contain a node of a type and
left/right sub `++tree` of the same type. Pretty-printed with `{}`.

<1.yat {a/<1|van {* <33.pdv 2.wxc $149>}> <33.pdv 2.wxc $149>}>
>
```

Suggestions?


galenwp:
This looks great to me. 

Will you be able to import the text from the docs in some reproducible way?  Such that, when we update the docs we can run something to update `hoon.hoon`?  

The `> ? tree` is sample output in `:dojo`, right?  That formatting seems to work just fine.

cgyarvin:
This *is* great.  The only thing I'd change, stylistically, just to match
our overall style feel, is universal lowercasing.  Unless anyone wants to
bikeshed? :-)

On Wed, Feb 8, 2017 at 11:20 PM, Galen Wolfe-Pauly <notifications@github.com

galenwp:
All lowercase would feel kind of forced here.  I'd prefer it to have normal capitalization.

cgyarvin:
It's not matching :talk that I mean, it's matching the reference comment
style in our latest/best vanes.  Uppercase in a Hoon file is generally kind
of jarring, since the code style doesn't use it.

The basic motivation for the style is that it's 2017 and people write
badly.  An all-lowercase style makes bad writing look better.

galenwp:
Sure, but I guess I'm curious about the answer to my prior question.  Is the text here the same text that appears in the docs?

eglaysher:
Updated everything to be lowercase. Fixed one spelling error and one style inconsistency.

This is mostly the text from the docs repo, minus some copy-editing, rewording, and documenting a few undocumented gates. This copy-editing was done by hand.

May I suggest that the hoon source be the canonical place for documentation? I encountered several gates which didn't have documentation in the docs repo. And why should they? That would require two patches to two different repositories. When the documentation is embedded in the code, either through docstrings or something like javadoc, there's a higher chance of actually having documentation written or updated; it's absence can be remarked on in code reviews.

(And the documentation here remains incomplete. I can't document `++quid` or `++quip` just from context, and I don't understand why the core has a `++foo` arm.)

ohAitch:
+1 for having fewer untracked dependencies between repos

galenwp:
I'd be completely fine having the docs generated from the source.  I agree that's the easier place to make edits.  The main thing I'd be concerned with is the tooling.  Like, can we run something that will update the docs?

ohAitch:
I mean, the pipe dream version of this is ford parses hoon.hoon into html

galenwp:
Yeah, I don't think we're quite there.

ohAitch:
So you could just have e.g. libraries/3c.md contain "<DocMD
file='sys/hoon.hoon' section='3c' />", plus some yaml if really necessary

galenwp:
Oh I wasn't really thinking about it that way, that's interesting.

ohAitch:
Hm actually if it's in the types you don't need to reparse it, you can just
pull the strings directly out programmatically; and if each core has a
++doc-name or sth, you can even use those to query from the markdown.

ohAitch:
Yeah like the actual structure of the docs contains information that's not
in hoon.hoon / zuse, and possibly shouldn't be; that usually wouldn't be
affected by functionality edits though, which is the difficult thing to
manually keep in sync.

ohAitch:
Like similarly, the text of the eyre spec definitely belongs in eyre, but
the "1. Models. 2. Requests" organization can be a dozen lines in the docs
repo specifying which doccords exactly are getting pulled in, and in what
order.

cgyarvin:
Elliot, any interest in weighing in on these "fat doccords" ideas?

Now that we have *some* doc in the source and it seems to work, is there a
logical way to extend or complete this idea?

eglaysher:
It depends on how fat we're talking about.

Right now, in addition to the text I extracted, each entry in the docs repository has an Accepts, Produces, Source and Examples section. The source could be produced from the actual hoon source, the accepts/produces could be generated if the `++span` information was expanded to record the string representation of the molds used. What's left is the Examples section, which could be rolled into the doccord. We could even do some magic to say something like quoted code lines ("> `(add 5 5)`") get parsed as a twig, with the example result autogenerated. (But that requires a (mini)Markdown parser in hoon.hoon.)

As for Anton's idea of putting the ordering information in a separate repository, I worry that that leads to the documentation not getting definitions for new gates.

(I tried experimenting with switching 1a, 1b, 1c into there own cores, but there was some sort of weird problem in vere and I gave up on that.)

ohAitch:
I disagree that you'd need a markdown parser in hoon.hoon, you can keep the
docs as plain markdown and convert them in userspace as required.

I'm generally in favor of having both the example code and results
available, if possibly not using the same doccord mechanism; it'd be almost
like unit tests :D

ohAitch:
(But in particular, if a documented example e.g. stops compiling, it would
be nice to know what the expected result was without digging around for a
working revision) (and conversely to document gotchas via type-failing
examples, in a manner that is clearly intentional)

chc4:
I added `<pre urb:exec>` tags a forever ago, which runs the code inside and says the result. This was definitely a mistake, but you can definitely only run the code in the examples when it's being rendered by the `:tree`.

eglaysher:
Anton, you have a point about expected results. Perhaps a simple <include "++add"> is best after all.

eglaysher:
Reminder to whoever merges this: you'll need https://github.com/urbit/arvo/pull/338 (and its jet patch in the urbit repository) merged first to parse the new doccord syntax used in this patch.

juped:
"Merging first and then fiddling later" is an extremely dangerous practice. Fortunately, it is not relevant here - there's nothing wrong with doccords as a complete feature, they are good to go.

Final result might end up rebased slightly depending on how the staging goes.

cgyarvin:
Yep, it's one of them fun time merges...

eglaysher:
(Closing old pull request now that the final one has things separated into three patches.)

PR #349, opened 2017-02-11
Remove <urb:pre> tags

chc4:
I added these a forever ago with plans on making the library docs self-explaining, but never did that for a number of reasons (speed, looks, %eyre going insane and breaking which I could never track down...)

I'm, uh, mildly concerned that this is still here. Because `slap`, although purely functional, doesn't time out Nock evaluations. And everything that is posted to `:tree`, including third-party content, is ran through this. You can connect the dots.

### DIFF
--- a/lib/elem-to-react-json.hoon
+++ b/lib/elem-to-react-json.hoon
@@ -31,26 +31,6 @@
   ?>  ?=({{$pre *} _;/(**) $~} src)
   ;codemirror(value "{v.i.a.g.i.c.src}");
 ::
-::  special handling for <pre urb:exec>foo</pre>
-++  urb-exec                                            ::  eval inline hoon
-  |=  src/manx
-  ?>  ?=({{$pre *} _;/(**) $~} src)      ::  verify it's only a text node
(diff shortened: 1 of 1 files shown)


ohAitch:
I think I've used this in a fora comment yeah :P

We could also be stricter about stripping onload etc components, right now
you could definitely XSS everyone into |hi-ing you their +code; or more
similarly, do some manner of billion laughs thing with the markdown.

In a mature urbit network, the answer to DoS is "don't make mechanisms
available to subs, if a planet does it delete the comment and ban the
planet"; for something like this, "recompile the markdown in the same event
so the write-request crashes and the comment is retroactively never written
down" would probably be sufficient. But while fora is public this seems
reasonable to remove yeah.

juped:
This should also be put on maint

ohAitch:
Are we worried that much about being spammed with uncomputable fora posts?
Any given one can just be deleted from clay.

chc4:
It would leave the hosting ship (and thus all of /fora) spinning until someone ssh's in and hits `^C`. That's pretty bad, I think.

PR #350, opened 2017-02-13
simple channel filtering

galenwp:
This makes it possible to filter talk on a specific channel.  It's really simple, but it does work.

### DIFF
--- a/web/talk/main.js
+++ b/web/talk/main.js
@@ -235,8 +235,12 @@ Persistence = _persistence({
         'cabal': 'cabal'
       });
     },
-    createStation: function(name) {
-      return Persistence.createStation(name);
+    createStation: function(station) {
+      Dispatcher.handleViewAction({
+        station: station,
+        type: "station-create"
+      });
+      return Persistence.createStation(station);
     },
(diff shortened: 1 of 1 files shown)


galenwp:
I did test this on a comet.  If anyone is impatient for the feature, feel free to throw the `web/talk/main.js` onto your ship.  Could be a good idea to `|merge %sandbox ~your-urbit %home` first and `|serve %sandbox`.

ohAitch:
Don't we already do something similar if you put a room in the query string?

galenwp:
Do you remember the parameter name?  This basically just makes the UI understandable.

joemfb:
I scanned this diff briefly -- it looks like it removes special-case handling of fora posts. Apologies if I'm must confused ... ;)

galenwp:
Good catch. Let me take another look.

galenwp:
You're right, and I don't want to remove that.  But I also can't find that code in the `/talk` repo.  @ohAitch ?

galenwp:
Got it.  I think this is ready now.

PR #352, opened 2017-02-14
remove tap arg

joemfb:
in order to preserve the crucial Strategic Parenthetical Reserves ...

for each `++in`, `++by`, and `++to`, i've replaced the `+-tap` implementation, indirecting through `+-rap` in case you want to keep the old name. The commits are granular, so it'll be easy to de/re-construct as needed ...

I've also done some dubious-quality work on the jets -- my latest version just dumps core. All this works without jet changes, though. I find that surprising ...

Finally, it's not based on the latest master, but a slightly older master: I couldn't get %clay to work on an up-to-date fake ~zod.

### DIFF
--- a/sys/vane/ames.hoon
+++ b/sys/vane/ames.hoon
@@ -1319,7 +1319,7 @@
           urb.ton
         %-  ~(gas by *(map ship sufi))
         %+  turn
-          (~(tap by urb.ton.fox) ~)
+          ~(tap by urb.ton.fox)
         |=  {p/ship q/sufi}  ^-  {p/ship q/sufi}
         :-  p
         %=    q
@@ -1422,7 +1422,7 @@
     ::
     ++  kick                                            ::    kick:am
       |=  hen/duct                                      ::  refresh net
-      =+  aks=(turn (~(tap by urb.ton.fox) ~) |=({p/ship q/sufi} p))
+      =+  aks=(turn ~(tap by urb.ton.fox) |=({p/ship q/sufi} p))
       |-  ^-  {p/(list boon) q/fort}
       ?~  aks  [~ fox]
       =^  buz  fox  zork:(kick:(um i.aks) hen)
--- a/sys/vane/ames.hoon
+++ b/sys/vane/ames.hoon
@@ -370,10 +370,10 @@
     ++  aver                                            ::  verify
       ?>  (lte cur.saw max.saw)
       ?>  !=(0 max.saw)
-      ?.  =(cur.saw (lent (~(tap to liv))))
-        ~&  [%aver-cur cur.saw (lent (~(tap to liv)))]
+      ?.  =(cur.saw (lent ~(tap to liv)))
+        ~&  [%aver-cur cur.saw (lent ~(tap to liv))]
         !!
-      ?>  =(rey.saw (lent (~(tap to lop))))
+      ?>  =(rey.saw (lent ~(tap to lop)))
       ?>  =+  |=  {a/coal b/coal}
               &((lth out.a out.b) (lth lod.a lod.b))
           |-  ?|  ?=($~ liv)
@@ -432,7 +432,7 @@
           [& $(liv l.liv)]
       ?~  ack  [~ ~ liv]
       =.  ded  ?:(top [n.liv ded] ded)
-      =.  ded  ?:(vig.clu.u.ack (~(tap to r.liv) ded) ded)
+      =?  ded  vig.clu.u.ack  (weld ~(tap to r.liv) ded)
       =.  lov  ?:(top [n.liv lov ~] lov)
       [ack ded lov]
     ::                                                  ::
@@ -512,7 +512,7 @@
         ::
         ::  everything in front of a dead packet is dead
         ::
-        $(liv l.liv, ded (~(tap to r.liv) [n.liv ded]))
+        $(liv l.liv, ded (welp ~(tap to r.liv) [n.liv ded]))
       =+  ryt=$(liv r.liv)
       [p.ryt [n.liv l.liv q.ryt]]
     ::                                                  ::
--- a/app/ask.hoon
+++ b/app/ask.hoon
@@ -44,7 +44,7 @@
 ::
 ++  adrs
   =-  (sort - lor)
-  %+  turn  (~(tap by adr))
+  %+  turn  ~(tap by adr)
   |=({a/email b/time c/invited} [tym=b ask=a inv=c])
 ::
 ++  new-adrs  (skim adrs |=({@ @ inv/invited} =(%new inv)))
--- a/app/dojo.hoon
+++ b/app/dojo.hoon
@@ -553,7 +553,7 @@
       ?+  a  a
         {?($cube $face) ^}  a(q $(a q.a))
         {$cell ^}  a(p $(a p.a), q $(a q.a))
-        {$fork *}  a(p (silt (turn (~(tap in p.a)) |=(b/span ^$(a b)))))
+        {$fork *}  a(p (silt (turn ~(tap in p.a) |=(b/span ^$(a b)))))
         {$core ^}  `wain`/core
         {$hold *}  a(p $(a p.a))
       ==
--- a/app/dojo.hoon
+++ b/app/dojo.hoon
@@ -644,7 +644,7 @@
       :+  (dy-silk-vase !>([now=now.hid eny=eny.hid bec=he-beak]))
         (dy-silk-sources p.cig)
       :+  %mute  (dy-silk-vase (fall (slew 27 gat) !>(~)))
-      %+  turn  (~(tap by q.cig))
+      %+  turn  ~(tap by q.cig)
       |=  {a/term b/(unit dojo-source)}
       ^-  (pair wing silk:ford)
       :-  [a ~]
--- a/app/gh.hoon
+++ b/app/gh.hoon
@@ -322,7 +322,7 @@
     [~ +>.$]
   ::  ~&  response=response
   :_  +>.$
-  %+  turn  (~(tap in listeners.u.hook-data))
+  %+  turn  ~(tap in listeners.u.hook-data)
   |=  ost/bone
   [ost %diff response]
 --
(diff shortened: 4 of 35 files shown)


joemfb:
Also, I'm completely lost by the jet-compatibility hack in `+-tap:to`.

It's completely necessary: removing it produces

```
fine: parent not found (7016793f)
improper core 2a9080b6
improper label: [%tap 28265 %two %one 'k149' 0]
```

But how? There are no hints ...

juped:
Don't feel basing things on the latest `master` is necessary - that's a decadence which is reluctantly _permitted_, not a good practice :)

I think I know the deal with the jets, will let you know after I look

joemfb:
Here's what I've done so far: https://github.com/urbit/urbit/compare/master...joemfb:detap

again, that branch core dumps on startup (after a ~20 second delay)

juped:
That probably means you got a refcount wrong somewhere.

joemfb:
I just tweaked the existing `in_tap` jet to be argument-less, a la `in_wyt`, but obviously messed something up.

I did try refcounting the null here (since it's decremented later) https://github.com/joemfb/urbit/blob/5201eff5e7e66f4b76a1c4caa22bb15f0545f835/jets/d/in_tap.c#L31, but that didn't seem to help ...

something like

```
u3_noun b = u3_nul;
u3k(b);
```

juped:
Direct atoms aren't really refcounted and u3k/u3z are no-ops on them (though you should still get the refcounting semantics right on them just because), so that can't be it.

joemfb:
The only other thing I'm not sure about is the `u3r_mean`, but `u3x_con_2` (`+14`) seems intuitively to be the right axis.

Also, 

```
> =<  +2  |_  a/@  ++  b  |.(a)  --
[[1 0 14] 0 1]
```

cgyarvin:
Is this still stalled on the broken jet?  It looks to me like you need u3k(b) in your base case -- the arguments to _tap_in are retained, not transferred.

joemfb:
Sorry about this, I'm sure I'm just missing something obvious ...

I'm not sure what mean by the base case, though: the recursion at line 23? (https://github.com/joemfb/urbit/blob/5201eff5e7e66f4b76a1c4caa22bb15f0545f835/jets/d/in_tap.c#L23)

I've tried with and without u3k(b) there... no difference. FWIW, there isn't one in the current version: https://github.com/urbit/urbit/blob/master/jets/d/in_tap.c#L24

Here's a backtrace of the crash: http://pastebin.com/vZ4NL2C3. Does anything jump out?

cgyarvin:
I skimmed this too casually.  I'll look again later...

cgyarvin:
No, it'd just look weird.  It's also kind of abusing the type system.

The problem is also that I have a vague desire to someday extend auras to
all spans, not just atoms -- to do things like differentiating between a
set and a map.  Real runtime bugs happen because we don't have this
feature.  Also, it's retarded that jet matching is this broken.  (We also
need to put hash recognition back in.)

joemfb:
This now works with the latest in urbit/urbit#820 (which has a link to a bootable pill).

Per my earlier question, is there any way I can jet these arms without what are effectively nested loops?

And I would assume `+-tap:to` should have no jet hint -- is that right? Any other changes I should make?

sys/vane/ames.hoon
```diff
@@ -432,7 +432,7 @@
           [& $(liv l.liv)]
       ?~  ack  [~ ~ liv]
       =.  ded  ?:(top [n.liv ded] ded)
-      =.  ded  ?:(vig.clu.u.ack (~(tap to r.liv) ded) ded)
+      =?  ded  vig.clu.u.ack  (weld ~(tap to r.liv) ded)
```
joemfb: This is the one call site (along with a similar expression in `++flay` that had to be actually refactored (where I couldn't just remove parenthesis). The same expressions are repeated in`xmas.hoon`. `+-tap` used to prepend to its argument, so this isn't exactly a complex change. I just wanted to call it out specifically. I haven't tested the expression in a generator, but I haven't tested network operations ...

cgyarvin:
Good question -- you shouldn't need the extra |. there.  Instead, decompose
the |- into |. and =<  $.  Put the jet hint between the |. and the =< $.

I'm not super worried about +-tap:to.

Perhaps I will never understand why you changed the semantics of ++tap,
then added ++rap to implement the old semantics.  On the other hand, this
does force you to simplify all the code that taps with ~ -- a small heroic
effort.  Also if you miss a spot in a merge, the compiler won't be shy
about letting you know.

joemfb:
I implemented the new semantics in rap, changed all the call sites to use that, then renamed everything back to tap. My goal was always to remove the extra parens, but I wasn't sure how much of this would be approved -- there was some argument in the original issue. Since this is all going in (and a new pill/ivory.c is required), there's no good reason for the indirection. This PR (and urbit/urbit#820, for that matter) could be squashed into a single commit. Would you prefer that?

As you suggest, I originally had just one decomposed loop:

```
=<  $
~%  %tap  +  ~
=|  b/(list)
|.  ^+  b
```

but that axis is wrong and the jet wasn't being called. Correcting the axis to `+>` 
(or just using `~/  %tap`) didn't call the jet either ... The duplication I have in this PR is the only combination I could find that properly registered and called into the jet. Any idea why that might be?

cgyarvin:
You're trying to jet something that isn't a core.  The `~%` and the line
below it need to be inverted.  Sorry, this must have missed my attention
earlier.

As for commits, what's ideal is a sequence of steps to follow when merging,
each of which boots from the pill produced by double-compiling the last.
In rare cases we may produce a pill which cannot compile itself, only the
next stage.  This is to be avoided!

ohAitch:
Is double-pilling really still necessary with the new boot sequence? Like,
if you edited hoon.hoon run +metal in test mode, but the second compile
will happen either way

cgyarvin:
I have to add a flag to make it not do the test compile.

Decrementing the Hoon version is also very helpful in these situations...

joemfb:
I've tried both of these, neither calls into the jet:

```
=|  b/(list _?>(?=(^ a) n.a))
=<  $
~%  %rap  +>  ~
|.  ^+  b
```

```
=<  $
=|  b/(list _?>(?=(^ a) n.a))
~%  %rap  +>  ~
|.  ^+  b
```

The hint axis is relative to the core, right? It's the axis of the containing core, from the perspective of the jetted arm?

ohAitch:
I think you might want ~% %rap + ~ since it doesn't have a sample

cgyarvin:
Anton, ~% annotates the product, not the subject.

More precisely, the axis (here ->) is the axis at which the next core up in
the stack is found, within the current core.  Arms are not at issue.  In
fact, `~/` should work here.

So I can't explain why this wouldn't work.  Of course that's normal when
debugging.  Perhaps a jet registration problem.

joemfb:
I think I've confirmed that this is a dispatch problem, not a registration problem. I'll see if I can debug the dispatch.

This is most I've ever done will pills and breaching, and I'm not sure what you mean by double-compiling. I did apply these commits incrementally while developing and testing the jets, but they can be applied with a single pill.

- new fake-zod from master
- `|mount %`, `|autoload`
- checkout this arvo branch
- copy in `sys/*`
- `.tap/pill +metal ~zod`
- checkout and make urbit/urbit#820
- new fake-zod from tap.pill

So I don't see the utility of separate commits, now that everything works. Any objections to squashing this down to one?

cgyarvin:
I have no objections to squashing this down to one.  Just bear in mind that
we're doing this on master because there's no way to

I remain curious as to what's going in with the jet dispatch...

joemfb:
Looks like that got truncated -- there's no way to what?

cgyarvin:
Doh!  No way to do that over the air on maintenance...

juped:
Please leave it unsquashed (or at least be judicious and just combine the separate in/by/to commits into one each)

joemfb:
@juped, I'm happy to structure this into any arbitrary set of commits that makes things easier for you -- just say the word. It seems like the indirection through `rap` doesn't serve any purpose, so these could be split by containing core (as you suggest), and then maybe again by kernel vs userspace?

joemfb:
I spent a bunch of time with `jets.c` and `runtime.md` this weekend, working through exactly how jet registration and dispatch work. I was making a bunch of assumptions which turned out not to be case.

The issue I was having was simply that my `rap` arm and jet where matching the already existing `tap` arm and jet. As we've already discussed `|=(^ .)` is exactly equivalent to `=|(^ |.(.))`, and jets are dispatched by the hash of the battery ...

Looking back over this discussion, there was a big clue in my original comment: "all of this works without jet changes ...". (I don't remember exactly what I was thinking; I guess I assumed that the arm was being naively evaluated as nock.) So this has been another episode in "joe does something stupid, kicks off a big discussion, and bangs his head against the wall until something falls out". Story of my career ...

I've updated urbit/urbit#820 so that it no longer touches `ivory.c`, and simply refactors the tap jet calling conventions. But it's completely independent of this PR -- this could even go out over the air, if need be.

juped:
I thought you were having these problems at the end, after existing tap was removed?

joemfb:
I've done a pretty terrible job communicating throughout this PR. 

I did my initial testing with both `+-tap` and `+-rap` present. There's an explanation for this, but, as I started to write it up, I realized just how bad it was. This is all somewhat embarrassing...

I was debugging via printfs in the jet, and getting utterly inundated with them, since everything calls tap. Of course, I could've just put the printf in `wdi_tap`, which is never directly called from C.

juped:
This seems like a rite of passage for a lot of people. Probably just need an extra paragraph in u3.md emphasizing how it works.

joemfb:
Well, I'm sorry for kicking off a wild goose chase. I would've preferred to pass thru the rite in private ... I'll try to digest what I've learned here and summarize it in a forum post.

BTW, the hint doesn't have to immediately precede the jetted core. The tail of the hint formula is evaluated before being passed to `u3j_mine`, so this works fine:

```
~/  %tap
=|  (list)
|.  ...
```

PR #355, opened 2017-02-25
Prevent unauthorized talk message sending.

Fang-:
Before going through with a `%publish` command, checks whether source of the poke is in our `++team`.  
(We don't crash/throw an error because we don't want to bother the user with the useless efforts of whoever is trying to be malicious.)

Solves #208. I believe this is the fix #221 was looking for.

Authorized messages to and from ships and their channels seem to still function fine with this change. I believe I'm catching the right case in the right part of the code here, but as always it'd be cool if someone could verify.

### DIFF
--- a/app/talk.hoon
+++ b/app/talk.hoon
@@ -1366,7 +1366,10 @@
       (ra-config p.cod u.q.cod)
     ::
         $review   (ra-think | her +.cod)
-        $publish  (ra-think & her +.cod)
+    ::
+        $publish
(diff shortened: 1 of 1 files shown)


galenwp:
The one place we exploit this behavior is on the web ship. We create session-specific comet names and spoof the author. The web ship acts as a proxy. 

More or less the same thing happens for fora posts. 

@ohAitch, what do you think? This makes patch makes sense. Maybe we just want to run custom talk code on the web ship.

PR #356, opened 2017-02-25
Doccords on layers 3 and 4.

eglaysher:
Finally, add doccords to layers 3 and 4. A lot of the incoming documentation still needs editing. I've tried to make grammatical corrections where I could, along with trying to get rid of anachronisms (odor -> aura), but there's simply so much editing that needs to be done here that I think that it's better to get it into the tree first.

layer 4's documentation was sparse and there's a lot left undocumented. layer 5 has almost nothing, to the point where I'm just going to ignore it.

### DIFF
--- a/app/dojo.hoon
+++ b/app/dojo.hoon
@@ -514,8 +514,8 @@
         %+  dy-print  cay
         =+  mar=|.(?:(=(%noun p.cay) ~ [%rose [~ "    " ~] >p.cay< ~]~))
         ?-  p.p.mad
-          $0  ~
-          $1  [[%rose [~ "  " ~] (skol p.q.cay) ~] (mar)]
+          $0  ?:(?=({$ex {$base $bean}} q.q.mad) (dy-show-overview cay) ~)
+          $1  (dy-show-span cay)
           $2  [[%rose [~ "  " ~] (dy-show-span-noun p.q.cay) ~] (mar)]
         ==
       ==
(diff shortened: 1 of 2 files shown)


cgyarvin:
One thing that would make the whole doccord PR as easy as possible to merge, I think... to merge a change like this effectively, we need to know the hashes of *two* key commits:

(1) an arvo that can be compiled by the master it branched off from, that implements doccords but *does not use* doccords;
(2) the actual final commit designed to be merged.

Obviously, (2) is a very poor substitute for (1).  It's probably not hard to figure this out from your commits, but I feel it should be noted as formally as possible.

cgyarvin:
I am still tempted, however, to fiddle with your doccord syntax.  It does not fit super-cleanly into the reference style I've devised, that you'll see in jael.hoon and zuse.hoon.  hoon.hoon has a slightly older style, but looking at the file one feels as if these forms are competing rather than collaborating.

One deficit I'm working with is that I don't have a vi mode that colorizes them.  (The vi mode is actually rather in need of attention.)  But as a language designer I like working with bad or no highlighting -- highlighting should supplement readability, not replace it.

One feature of a lot of Hoon code is the 19-letter arm summary after a :: at column 57.   This form of comment stands out from the code very nicely, and is easy to search for visually because it's on the right margin.  The new standard has one 19-letter summary, and right above it a :: line with just the arm name preceded by ++, no space.  This again makes for easy searching both visually and with the editor.  It's a shame to waste the right margin and I feel this standard looks good.

The downside is that your arm has to explain itself in 19 letters.  This is much less onerous if there is also a formal model for a longer explanation, as of course there is in doccords.  But there are also a ton of these 19-letter explanations and it seems lame not to formalize them.  I also cannot find a way to make doccords + this existing format look good together.

I'm also not sure I like the blank line.  Previously, good Hoon had no blank lines.  It is possible to change -- but when I read Hoon, I like to follow the column of runes down the page, and hitting a line where there is no column feels weird.

(I would post this to the fora and probably will eventually, but I would prefer to keep bikeshedding to finite levels!)

eglaysher:
I've gone and reordered the sequence of patches into three:

- A patch which *just* implements the new hoon syntax but which does not use it so that it can be compiled by the arvo kernel at 4288aab089c46cc.
  
- A patch which *just* implements all the new features in dojo, which requires a kernel with the previous patch.

- A patch which *just* adds all the documentation to hoon.hoon. (This includes further lower-casing; I went through with a regex and think I got the last of the doccords which have upper case characters in them.)

Note in case you didn't see it: you'll need to find out what's causing issue #357 before you'll be able to merge anything to the master branch, as really trivial code isn't compiling anymore. A previous version of this patch

eglaysher:
As for the rest...I don't really know how you make the two formats play nicely together.

Even before working on doccords, I didn't rely on the right hand side to identify where the breaks between arms were. In Emacs, arm names are fontified with a face not used in any other context. The arm name itself becomes the visual delimiter. A separate face for doccords enhances this effect (though admittedly a trickster could embed a doccord in the middle of an arm instead of at the top). Immediately after I'd commented the relatively tiny layer one as a proof of concept, I stopped work on everything else to write urbit/hoon-mode.el#7.

Regarding the blank lines, I sympathize on this one. I just don't see a way around that though if you want doccords to be a markdown subset. I suspect that quality top-level core doccords will be multiple paragraphs long.

ohAitch:
The obvious javadoc answer is requiring some kind of prefix on every line,
but that really isn't practical without editor plugins that are a lot
heavier than anything I'd want to require/encourage, similarly to the "does
this look decent with 0 highlighting" principle.

cgyarvin:
For a language *user*, a great editor mode is a godsend.  In fact I should
seriously consider switching to one of those godawful vi-in-emacs layers
just to try your awesome mode.

For a language *designer*, I'm not so sure this is true.  I subscribe
pretty strongly to the belief that a language's base usability should be
defined in plain old text.  If this comes out right, highlighting, fonts,
etc, can only improve it.

Making writing harder (without special editor assistance), as in "I have to
reformat all my block comments by hand" is not that tough a tradeoff for
me.  I admittedly should write more block comments in my Hoon.  But I write
plenty these days, and it really doesn't bother me to have to align them by
hand.  I mean sure, I'd rather I had a tool.

So sticking a rune, or pseudo-rune, on every line of a block comment
doesn't bother me terribly, not as much as blank lines.  It also eliminates
weird problems with the ''' syntax, which was never exactly my favorite
irregular form in the first place.

I think realistically the best way to address this is to merge the patch
right now, since it's a good patch and it works.  I've now written enough
code in my reference convention (which, I agree with Ray, is a little too
ornate) for me to have at least a sense of where to go next.  The result
may involve some modifications to docstring syntax and/or semantics, and
should produce an actual canonical style guide document for writing
good-quality Hoon.  That alone should be very helpful for the contribution
process.

Elliot, any more opinions on this?  It's already clear that we'll find
doccords pretty essential to the next round of documentation...

eglaysher:
As for some sort of prefix syntax, are you envisioning something like this?

```
++  blah                                            -:  performs blahness on args
  -:
  -:  here starts a full paragraph about the blah operator, and why
  -:  even though it looks like a simple add, it actually has a long
  -:  and illustrious intellectual tradition on the continent.
  |=  {a/@ b/@}
  (add a b)
```

`-:` is highly likely to be the wrong rune, but if conceptually that's what you're thinking of, I don't have any problem with it. Either way, I agree that merging as is, and making further modifications is probably the easiest way to go.

I will dissent on the Emacs mode being awesome; it is only now approaching MVP. It really only is fontifying the buffer (and even that is incomplete; i.e. molds in `?+`). Indentation, in-file navigation, outlining...the thing has a long way to go before I'd feel comfortable, say, submitting it to MELPA.

cgyarvin:
Something like this -- but the interesting problem is applying it to the
syntax I usually use in the models section.

Also, note that there are actually three spans you can decorate in a normal
gate like this.  Ideally they should be decorated separately -- the core
itself, the sample, and the product.  Much better than one big
indiscriminate block comment.

eglaysher:
Are you thinking about something like this?

```
++  data-type
  :>  a short description of when this is used.
  $%  {$thing @ud}                            :<  copying a constant
      {$joke tape}                            :< another comment
  ==
++  join-tapes
  :>  intersparses `a` between each `b`.
  |=  $:  a/tape                              :<  separator to add between items
          b/(list tape)                       :<  tapes to be joined
      ==
  ^-  tape                                    :<  the output
  [...]
```
(Name the twigs `$glee` and `$sad`. The only issue here is that these really aren't cell twigs.)

ohAitch:
Ooh, I definitely like the usage of ><

cgyarvin:
It occurs to me that while Elliot's design is almost perfect, there's an interesting question as to whether :< and :> are mold-specific.  

For instance, the `:<` on `{$thing @ud}` above is clearly a description of the product of the mold function `{$thing @ud}`.  But on a normal gate, the header describes the gate, not its product.  Same on a `|%`.  

So there wants to be another degree of freedom in the syntax, but this is clearly too much.  The alternative would to break some rules and do a little smart post-parsing.  We do try not to do this in Hoon, but it's certainly not an ironclad rule (as the recent discussion of preludes shows).

In general, the more I think about it, the more I'm convinced that few if any comments in a Hoon file should not end up in the twig.  For instance, arguably inline block comments should have the semantics of ~|, so that you see them in any stacktrace that contains them.  Extra points if these live block comments can interpolate Hoon via the {} or `` syntax.

Ray's view, and I think he is right, is that my reference formats (jael/zuse/parv) are a little too ornate.  But I think we can construct files with the same basic look, in which all or almost all comments are formal content that go into the twig.

I also think a "parse from rune + 2 spaces, to end of line" syntax should replace ''' for blockquotes.  There is nothing in principle wrong with ''', but a programming language should think about text in only one way.

ohAitch:
Relatedly, we should probably have official doc for "semi-tall form" that
ends on EOL, I feel like there are a couple existing places where it shows
up.

cgyarvin:
Can you be more specific?

ohAitch:
I guess it's just comments and a couple html templating ones

Like, iirc doc sez "all hoon is tall or wide form; there are some irregular
non-runic forms but they are also wide", but there's also the
midway-between-tall-and-wide form which is "the entire rest of the line is
treated as a string"

cgyarvin:
Can you be more specific? :-)

juped:
Superseded by #389

Issue #175, opened 2016-06-21
clay has no default behavior for marks without a `++grad` arm


joemfb:
I'm not exactly sure about the nature of this bug (or which aspects of it should be highlighted), and, therefore, the title of this ticket ...

I ran into issues after adding an HTML file to `/=home=/web`. Specifically, my syncs from `~doznec` failed, as did my attempts to update the HTML file through the filesystem mount.

Here some of the errors I saw: https://gist.github.com/joemfb/0b6474348c88539064806a3693a642a4

I solved the issue by adding a `++grad` arm to `mar/html.hoon`:

```
++  grad  %mime
```

It seems like `%clay` should have this or similar behavior as the default. The discussion in :talk suggested that there may be a `%ford` bug (or a compilation bug presenting in %ford). I don't have any additional details about that.

joemfb:
I just reproduced this on a comet. I copied an HTML file to `web`, then save a change to that file:

```
+ /~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/home/539/web/trivial/html
[%bad-softing 'crud']
[%bad-softing 'warn']
crude: all delivery failed!
  /~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/home/~2016.6.20..18.48.22..0691/arvo/hoon:<[10.975 15].[11.006 57]>
  /~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/home/~2016.6.20..18.48.22..0691/arvo/hoon:<[10.976 15].[11.006 57]>
  /~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/home/~2016.6.20..18.48.22..0691/arvo/hoon:<[10.977 15].[11.006 57]>
  /~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/home/~2016.6.20..18.48.22..0691/arvo/hoon:<[10.979 15].[11.006 57]>
  /~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/home/~2016.6.20..18.48.22..0691/arvo/hoon:<[10.981 15].[11.006 57]>
  /~wolsyp-filtuc-havrut-londys--pinbel-ballut-fodmyl-litmer/home/~2016.6.20..18.48.22..0691/arvo/hoon:<[10.981 29].[10.981 44]>
[...]

PR #358, opened 2017-03-02
Add `++  grad  %mime` to html.hoon mark.

keatondunsford:
Fixes issue where ford doesn’t know which of two diffed html files to
build.


joemfb:
related to #175 (commenting for posterity)

keatondunsford:
@joemfb Oh damn, ha, this is exactly what was going on. Yeah, this came up when I was trying to serve some raw .html for an Urbit single-page-app tutorial I'm working on. A total fix of that across ford / the marks would be awesome. Maybe that's something I can do.

PR #359, opened 2017-03-02
Full ++rh core, other fixes

### LINKED ISSUE #289: Floating point rounding glitch/jet mismatch
```
> (sub:rs .1 (bit:rs [%f & -26 3]))
.9.9999994e-1
> (bit:rs (sub:pa:ma:rs [%f & --0 1] [%f & -26 3]))
.1
```

In `++sub:fl`, there's a shortcut that checks whether the subtrahend is less than 1/2 the ULP (unit in last place) of the minuend, and if that's the case, doesn't bother actually subtracting but rounds the minuend in a particular way to get the right result. This keeps, for example, `(sub:fl [%f & --0 1] [%f & -10.000 1])` from performing a 10000-bit shift when the answer is already clear.

But even if a subtraction passes the shortcut conditions, the difference can sometimes be one degree more precise than the minuend. In this case, "round difference to nearest ties to even" cannot simply be replaced with "round minuend to nearest ties toward zero". Although addition has a similar shortcut, this problem doesn't happen, because the result can only get _less_ precise.

In the example code, minuend's ULP is `2^-23`, and subtrahend is `0.75 * 2^-24`; the ULP of the correct, jetted result is `2^-24`.

The fix is very simple: in `++sub:fl`, replace `=+ ^= x %+ sum:si e.b (sun:si mb)` with `=+ ^= x %+ sum:si e.b (sun:si +(mb))`, making the shortcut require the subtrahend to be less than **1/4** the minuend's ULP instead; this way, no round-to-nearest subtractions will result in the difference being more precise than the minuend. Pull request to cc-release branch is in the works.

max19:
Includes half-precision support, removes some unnecessary code

fixes #289

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -2019,9 +2019,9 @@
           $sm  [%f & zer]
           $ce  [%f & spd]
           $lg  [%f & spd]
-          $ne  ?:  s  [%f & ?:((^lte b (bex (dec q))) zer spd)]
+          $ne  ?:  s  [%f & ?:((lte b (bex (dec q))) zer spd)]
                [%f & ?:((^lth b (bex (dec q))) zer spd)]
-          $nt  ?:  s  [%f & ?:((^lte b (bex (dec q))) zer spd)]
+          $nt  ?:  s  [%f & ?:((lte b (bex (dec q))) zer spd)]
                [%f & ?:((^lth b (bex (dec q))) zer spd)]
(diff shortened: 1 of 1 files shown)


cgyarvin:
This is super duper awesome.

Someday it will be time to make Urbit use hard floating point -- very, very
carefully.  Until then you'll have to run your climate simulations, or
whatever, on Unix...

PR #360, opened 2017-03-02
adds sticky prompt scrolling

### LINKED ISSUE #18: Sole scroller always right-aligning is confusing
Currently, if you type more than _termwidth_ characters, the left side of your input scrolls off the side of your prompt; and if you go back any, anything to the right of the cursor scrolls back off the right side. Would it make more sense for the current visible slice to be "sticky", with only motion to the left or right edges moving it? I know I find it confusing when I see a typo, scroll back to it, and can no longer see the five words I've entered since.

joemfb:
In my perpetual quest for bite-sized pieces of work, I stumbled on #18. Seemed easy enough, so I dove in, with the additional goal of adding truncation markers where appropriate. That led me down a rabbit-hole resulting in https://github.com/joemfb/arvo/tree/prompt-ellipses, which correctly adds ellipses, but doesn't truncate. I had to basically bruteforce the solution, since `term.c` does some weird things with the cursor based on the buffer length (and those seem to interact poorly with the constantly-refreshing approach ...). My conflicted inclination to rewrite term.c intensifies ...

So I switched back to the original task. 

To have a stable, sticky input window, you have to know how long the prompt is. Once you're there, it really makes sense to make the prompt itself sticky - the alternative is kinda strange. A sticky prompt could be a problem if the prompt is *really* long, so this includes the old-style scrolling for prompts >= 35% of the console width.

Then you have to reset the offset for any buffer discontinuities (like navigating through history), and when rotating the drum. That means your offset can change when you rotate the drum all the way around - not a big problem, but it can be irritating. The solution would be to store per-target offset: does that seem worth it?

I wrote this against maintenance for ease of development, but I'm happy to port it to master if you'd prefer to have it there (some of the `=.` twigs are structured to take advantage of `=?` when the time comes).

closes #18

### DIFF
--- a/lib/drum.hoon
+++ b/lib/drum.hoon
@@ -412,7 +412,11 @@
   ^+  +>
   =/  pol  (lent-char:klr pom)
   =/  end  (sub edg pol)
-  =.  off  ?:((lte p.lin end) 0 (sub p.lin end))
+  =.  off  ?:  (gth p.lin (add end off))
+             (sub p.lin end)
+           ?:  (lth p.lin off)
+             (min p.lin (dec off))
+           off
   %+  se-show
     (sub (add pol p.lin) off)
   (welp pom [*stye (swag [off end] q.lin)]~)
--- a/lib/drum.hoon
+++ b/lib/drum.hoon
@@ -647,9 +647,7 @@
   ::
   ++  ta-off                                          ::  reset buffer offset
     |=  ted/sole-edit
-    =.  off  ?:  |-  ?.  ?=({$mor *} ted)
-                   ?=({$set *} ted)
-                 |-  ?~(p.ted | |(^$(ted i.p.ted) $(p.ted t.p.ted)))
+    =.  off  ?:  (any:edit ted |=(a/sole-edit ?=($set -.a)))
                0
              off
     +>
@@ -884,6 +882,11 @@
         (cut pos num)
         (cat pos txt)
     ==
+  ++  any                                             ::  matches?
+    |=  {a/sole-edit b/$-(sole-edit ?)}
+    ^-  ?
+    ?.  ?=($mor -.a)  (b a)
+    (lien p.a |=(c/sole-edit ^$(a c)))
   --
 ++  offset                                            ::  calculate offsets
   |%
--- a/lib/drum.hoon
+++ b/lib/drum.hoon
@@ -408,10 +408,14 @@
   +>(mir lin)
 ::
 ++  se-just                                           ::  adjusted buffer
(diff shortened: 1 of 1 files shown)


lib/drum.hoon
```diff
@@ -631,22 +645,31 @@
       ==
     ==
   ::
+  ++  ta-off                                          ::  reset buffer offset
+    |=  ted/sole-edit
+    =.  off  ?:  |-  ?.  ?=({$mor *} ted)
+                   ?=({$set *} ted)
+                 |-  ?~(p.ted | |(^$(ted i.p.ted) $(p.ted t.p.ted)))
```
joemfb: I just have to say, I'm inordinately proud of this formatting

cgyarvin:
Woah.  It's a little... radical... for me... but the set of Hoon joseki is
definitely not bounded or canonical.

cgyarvin:
This is neat - Anton is currently in the mode of trying to connect the new
console to ++drum.  He should look at this PR as well.  Yes, it would be
more interesting to have on master.

I have a feeling that before we rewrite term.c or anything, we should do a
better job of redesigning it...

joemfb:
The real question is, will a `$mor` sole-effect ever contain a `$set`. Practically speaking, I would expect the answer to be no - it'd be a pretty stupid sole-effect. So the condition could just be `?=({$set *} ted)`. But it is technically possible ...

joemfb:
should I do the forward port in a separate pr, or just rebase this onto master?

juped:
Another PR - ideally they'd be the same but we are not there yet.

cgyarvin:
I don't think so, but... it's certainly time to revisit this whole pathway.

lib/drum.hoon
juped: Not vertical enough!

joemfb: but the symmetry!

cgyarvin:
Right, but we strive against the mentality that the appearance of
readability is more important than the reality.  This looks pretty, but
it's relatively difficult for my eye to parse it into a twig...

lib/drum.hoon
ohAitch: How about
```
++  any-in
  |=  {a/sole-edit b/$-(sole-edit ?)}  ^-  ?
  ?.  ?=($mor -.a)  (b a)
  (lien p.a |=(c/sole-edit ^$(a c)))
...
=.  off  ?:  (any-in ted |=(a/sole-edit ?=($set -.a)))
           0
         off
```

joemfb: Thanks! That's pretty clean.

PR #364, opened 2017-03-04
prevents unsafe $cork (|^) expansion in the parser

joemfb:
This supersedes #343 (at least, temporarily). Should that be closed, and instead tracked in an issue?


joemfb:
strangely enough,  automatic `|reset` worked when testing this on a master fake-zod

chc4:
Nitpick: maybe use a different arm name than `++expx`? That's a pretty high-value title, and sound more general-use than it is.

joemfb:
I agree, but lack creativity ... `++rexp`?

PR #365, opened 2017-03-07
adds optional %tstr aliases prelude to %brcb

### LINKED ISSUE #361: Add an `aka` for the identity of a core
You've surely seen those arcane typecasts of `^-  {(list move) _+>.$}` in state machines and app arms. It is, frankly, scary and hard to explain since it's basically the equivalent of dereferencing a pointer literal in C, and pretty much the only spot where Nock shines through. It's basically the only reason for the explanation of lark syntax.

`=*` allows you to define a variable as an alias instead of by-value copy. They also update through context updates (`(|=(@ =*(b +< =+(c=1 b))) 4)` compared to `(|=(@ =+(c=1 +<)) 4)`), unlike the current `+>.$` invocation which you have to potentially manually update if you add a pin.

Is there a reason that `|_` at least doesn't add a `=*  this  +>.$` to its expansion? I don't actually know how %core is expanded, so it might require something like adding the bull to the top of each arm in the core the door expands into, since the sub-twig of a core is actually a `(map term twig)` instead.

joemfb:
see #361 for discussion

I had to comment out the `1-d` assertion in `gen/metal` to make a pill from this. I've uploaded it here: https://www.dropbox.com/s/2nz4w5cxdwjqqmo/door-aliases.pill?dl=0

Here's my test case:

```
|=  arg/@
=<  abet:(~(flub . 10) arg)
|/  a/@
+*  this  .
    samp  +<
    cont  +>
++  abet  [~ a]
++  flub
  |=  b/@
  ?>  =(a samp)
  ?>  =(+>+>.$ cont)
  this(a (add a b))
--
```

How should I format the `$door` page to preserve the margin?

Also, there are duplicate `:door` keyword declarations, along with both the `|_` and `|/` runes. Should the keywords be deduplicated? And can `|/` be removed?

closes #361

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -5514,7 +5514,7 @@
     {$bcts p/toga q/root}                               ::  $= name
     {$bcsm p/twig}                                      ::  $; assembly
   ::                                            ::::::  cores
-    {$brcb p/chap q/root r/(map @ tomb)}                ::  |_
+    {$brcb p/chap q/root r/(list (pair term twig)) s/(map @ tomb)} ::  |_
     {$brcl p/chap q/twig r/twig}                        ::  |:
     {$brcn p/chap q/(map @ tomb)}                       ::  |%
     {$brdt p/chap q/twig}                               ::  |.
@@ -6535,7 +6535,14 @@
         {$halo *}  ~(clam al boil)
         {$bcsm *}  p.gen
     ::
-        {$brcb *}  [%tsls [%bunt q.gen] [%brcn p.gen r.gen]]
+        {$brcb *}  :+  %tsls  [%bunt q.gen]
+                   :+  %brcn  p.gen
+                   %-  ~(run by s.gen)
+                   |=  a/foot  ^+  a
+                   =-  ?:(?=({$ash *} a) [-.a -] [-.a -])
+                   |-  ^-  twig
+                   ?~  r.gen  p.a
+                   [%aka p.i.r.gen q.i.r.gen $(r.gen t.r.gen)]
         {$brcl *}  [%tsls [%ktsg q.gen] [%brdt p.gen r.gen]]
         {$brdt *}  :+  %brcn  p.gen
                    =-  [[0 [~ ~] -] ~ ~]
@@ -6552,8 +6559,8 @@
                    :+  %brcn  p.gen
                    =-  [[0 [~ ~] -] ~ ~]
                    (~(put by *(map term (pair what foot))) %$ ~ [%elm r.gen])
-        {$brts *}  :^  %brcb  p.gen  q.gen 
-                   =-  [[0 [~ ~] -] ~ ~]
+        {$brts *}  :^  %brcb  p.gen  q.gen
+                   =-  [~ [[0 [~ ~] -] ~ ~]]
                    (~(put by *(map term (pair what foot))) %$ ~ [%ash r.gen])
         {$brwt *}  [%ktwt %brdt p.gen q.gen]
     ::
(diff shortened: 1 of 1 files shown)


chc4:
The duplicate `:door` is due to `|/` being the rune to create a wet door, which is no longer possible (killed in 151k). It should be perfectly fine to coopt the rune.

I love you for this.

joemfb:
*appreciates your platonic expression of affection*

So wet/dry used to be for the whole core, not per arm?

Note: this works with all three forms: `|_`, `|/`, and `:door`.

Apropos of nothing, you can do some interesting things with aliases:

```
> =<  a  |_  ^  +*  this  ?>(=(0 +<-) .)  ++  a  +<:this  --
[0 0]
> =<  ~(a . [1 2])  |_  ^  +*  this  ?>(=(0 +<-) .)  ++  a  +<:this  --
(crash)
```

As long as they're "read only":

```
> |_  ^  +*  this  ?>(=(0 +<-) .)  ++  a  +<:this  --
<1.syg {{* *} {our/@p now/@da eny/@uvJ} $~ our/@p <22.pss 9.vbi 8.slw 118.dat 209.nlg 55.wea 117.wvl 33.xnr 2.wxc $149>}>
> |_  ^  +*  this  ?>(=(0 +<-) .)  ++  a  +<:this(+< *^)  --
(crash)
```

cgyarvin:
Wait, writing through aliases isn't working?  `this` is not a perfect
substitute for `+>`?

joemfb:
Writing through *an alias wrapped in an assertion* doesn't work. But a simple alias works just fine -- see the generator above.

I was just messing around to see how far I could push aliases ...

cgyarvin:
*takes a closer look* oh yeah, that ain't gonna work.

joemfb:
This works now. I'm still unable to `|reset`, but I can boot a fake-zod from a brass pill (after applying #433).

belisarius222:
@joemfb This PR no longer seems relevant for hoon 143. If you agree, we should probably close this.

joemfb:
Well, it worked on %143 -- the current conflict is just `twig` -> `hoon`.

I think it'd be a useful feature, particularly in nested state machine cores. I've needed to update subtrees of parent samples without dedicated faces, and had to use lark or local `=*`. Do you think having all of `++`, `+-`, +=`, and `+*` is one too many?

ohAitch:
I think the motivating example in the PR is unconvincing, since all three of those can be ++ without problem; however if the intention is to use the syntax purely for non-`this` writes, I can see it being useful. I'd still have one alias per `+*`, for consistency with the other arm syntaxes.

ohAitch:
Ideally though the "faceless subtrees of parent samples" would just get optional names in situ, through some kind of moldable syntax sugar for `=>(^=(face foo) =,(, .))`, instead of them being declared inside the core itself.

cgyarvin:
It's stretching the metaphor of a core a little but I think it's a good
idea.  Any particular reason to restrict it to |_?

joemfb:
@ohAitch, [new ames](https://github.com/urbit/arvo/blob/dc25979f5945d6028ed2324f9e525744a92d56a7/sys/vane/xmas.hoon#L141) uses exactly the pattern I have in mind (also [here](https://github.com/urbit/arvo/blob/dc25979f5945d6028ed2324f9e525744a92d56a7/sys/vane/xmas.hoon#L578) and [here](https://github.com/urbit/arvo/blob/dc25979f5945d6028ed2324f9e525744a92d56a7/sys/vane/xmas.hoon#L998)).

Situational names are definitely an improvement over raw lark, and these core aliases would probably overkill in some cases. `=,(, .)` is a very cool idea, although maybe too powerful -- I could see it being a little hard to trace back and find the source of limbs.

@cgyarvin, I just figured there wasn't much of a need in `|%`, or the one-armed variants. Rank paternalism on my part. ([This](https://github.com/urbit/arvo/blob/dc25979f5945d6028ed2324f9e525744a92d56a7/sys/vane/xmas.hoon#L11) in new ames suggests a use case ...)

Mostly, I just want to support and encourage 1) the regular use of names 2) in consistent patterns. I don't know if this is optimal syntax, but I think it's an improvement.

cgyarvin:
I also have to wonder about:

?:(?=({$ash *} q.a) [-.q.a -] [-.q.a -])

joemfb:
see #366

PR #369, opened 2017-03-17
Fixed ++oust to not require faces for [a b].

Fang-:
This functioned fine when it was originally implemented but, you know, wet gates.


cgyarvin:
You've got a lot of whitespace changes in this PR -- are they germane?

Fang-:
Oops, those slipped under my radar. Not intentional/relevant, reversed.

juped:
Don't reverse it in a separate commit, this rather misses the point

Fang-:
You're absolutely right. Properly cleaned up now!

PR #374, opened 2017-03-21
Generate a coherent pill that includes userspace.

cgyarvin:


ohAitch:
pls no pr "not quite working yet", if the commit message is wrong the commit can be amended

cgyarvin:
Huh, I thought I rebased that message out...

juped:
Isn't this a step backwards? Why is it desirable?

With the boot sequence upgrades, we're very close to the desirable goal of taking vanes out of the opaque binary blob and leaving it compiler-only.

cgyarvin:
It's not a binary blob, it's a list of events containing source.  Only the compiler is delivered compiled, and even that we recompile.

ohAitch:
It's jam; shipping everyone the full arvo OS and not just the kernel module
is desirable, at least for star ames load reasons, but it's true that for
development/reproducibility we also want the binary-hoon-only "slim" pill
which pulls all sources from unix

juped:
OK, but it's a binary blob.

Shipping vanes as source inside the blob was a great intermediary step towards cutting down the blob.

It has nothing to do with development or star load or whatever.

ohAitch:
For new planets being started in the near future, it is better if
arvo@latest is initially downloaded over https and not ames, because https
is mature and can ship the source tree without tying up the serving process
for multiple seconds

That said I'd also be more comfortable with e.g.
bootstrap.urbit.org/latest.tar.gz, which itself contained a pill
with boot/compiler nock, and the initial clay contents as unix-legible
text.

juped:
I am not talking about that; rather, I am asking Curtis a question about the bootstrap blob

ohAitch:
The bootstrap blob having all of userspace in it is meant to replace -A(/at
least in common usage), by putting all the dependencies needed to boot
it(ignoring vere) in the same place; and help with clay initial-merge
traffic

juped:
Getting this thread back on topic: Curtis, isn't this a step backwards? Why
is thus desirable? We are close to being able to take the vanes out of the
binary blob, reducing it only to the compiled compiler - a desirable thing,
as far as I can see. If self-hosting compilers *could* be shipped as
source, that would be even better, but they can't.

Perhaps I am misunderstanding the nature and this is a work in progress,
with source to be replaced with placeholders much as identity is?

cgyarvin:
No, the optimal thing is to go in the other direction -- I'd prefer to ship
a checkpoint file with the boot events baked in.  So there would be zero
startup time.

Why is the thing you see desirable?  Who desires it, and why?

I can tell you why the coherent pill desirable: installing the whole system
in one sequence means (a) one less way to screw up dependencies, and (b)
lets you get a running urbit without either loading from a github repo or
urbit network sync.

Sync is a service, not an assumption.  Your urbit will work just fine
without it.  And we can sync any urbit ever created - the OS package that
contained the pill will never become stale.

(a) is right now making it much, much easier (not that it's easy) to modify
++span for doccords 2.0.

juped:
Imagine for a moment the magic world where self-hosting compilers don't need to be bootstrapped, and we can just distribute the whole system as source.

Clearly, in this world, we'd distribute the whole system as source, and no one would ever have thought of distributing it as a blob. Sure, you would have thought of distributing a post-boot checkpoint file, but that's totally irrelevant and can be dealt with elsewhere.

In the real world, Adam ate the apple and his children have to bootstrap with a precompiled blob. For unfortunate historical reasons, this has non-compiler stuff in it, because booting used to be largely a sham.

Now we're at the point where booting isn't a sham - the historical reasons are gone, and we can finally stop distributing anything but 149.blob. I am therefore perplexed by your desire to not only leave the non-compiler stuff in the blob, but put *all* non-compiler stuff into it.

Blobs are not good! Maybe if you're shipping a comprehensive blob editor and blob-based version control system so that we can stop using text files on Unix as source, and just edit the blob via blob-editor magic, it's a different story. I don't think this has been produced or proposed.

cgyarvin:
The boot sequence is the boot sequence.  The question is not whether it's
source or binary.  The question is whether it's created locally or sent
over the network.  Anything sent over the network is not part of the boot
sequence, by definition.

In the new arvo boot sequence, we install userspace before we start the
network.  Sync is handled in userspace.  We can't sync in the sync code.
This design is final.

It would be a fairly small effort to build the boot sequence as source in
your unix distribution, rather than building it as a binary and shipping it
as a blob.  Many restaurants have a neat trick where they come to your
table and prepare food, which is served flaming hot.  My view is that bits
is bits.  So I am not sure it matters that we ship it as a blob.  If it
does matter, this can be corrected fairly easily.

Does it matter? You didn't answer my question: who desires this and why?
Please attribute this desirability to some user or users, and and
motivation.

ohAitch:
Ignoring the checkpoint thing because we're not doing the checkpoint
thing(that's what an old pill was for the most part); jammed source !=
blob. The hoon to unpack the whole thing as e.g. shar would be <100 lines,
which I'll probably end up writing the fifth time a newpill fails to boot.

Curtis, if I used dark magic to make the jam byte-align so it's viewable in
vim, would this offend your sensibilities overmuch? Assume that I could fit
this in an opaque one-liner which could be commented out if desired, and
that anyone trying to ~edit~ the thing assumes responsibility for
overwriting bytes only, retaining the /length/ of any atom source.

ohAitch:
I desire strongly that I can start a fakezod with a historic, non-forwards-compatible zuse version, without having to hunt down / create a pill. This is a question of development, however, not end-user distribution.

juped:
"The network" is entirely irrelevant - just assume it doesn't exist. The question is *why a blob*? "Because you didn't give me a poll of users" is not an answer, but I have posted a poll to urbit-meta. It is strange to want to *add* the extra wrinkle of a blob where one doesn't absolutely have to - as the advocate of a simpler, more obvious system, I disclaim the burden of proof. I am a user, and I want to control the urbit code I run, and not have to constantly `+metal` to do so.

Can you, for example, elaborate on your remark about "making it much, much easier (not that it's easy) to modify ++span"? I don't understand how a blob helps here, but I can see the hindrance caused to nearly everything by the requirement of a new, extra step (generate blob).

cgyarvin:
At this point I have no idea what you're talking about.  I'm convinced the
solution to all communications problems is to convince people that they
can't send pointers over the network.

There is always a boot sequence.  It always contains exactly three early
events.  These events are Nock formulas, not text files.  All the other
events in the boot sequence deliver source.

It is much easier to modify ++span when I'm sure that app/ and sys/ came
from the same commit.

juped:
>These events are Nock formulas, not text files. All the other events in the boot sequence deliver source.

Source *is* text files. There's actually no reason I don't wire up my text editor to run uuencode/uudecode on write/read - because it's the kind of thing you need a good reason to *do*, not to *refrain from*. Why, then, wouldn't the text files be left as text files?

>It is much easier to modify ++span when I'm sure that app/ and sys/ came from the same commit.
Under what circumstances are you unsure of this? Note also that it's sometimes useful to deliberately mismatch them. I do this a lot. But I deliberately chose to, every time.

cgyarvin:
That's true, but other parts of the boot sequence need to be generated by
Urbit.  So it's easiest just to generate the whole thing at once and treat
it as a single noun.

Again, if I have a thing I get from two places, I have a process I can fuck
up.  I rather wouldn't have that process.

I don't hugely mind pulling source files from a resource directory in the
urbit install, and assembling the boot sequence thus.  They can't come from
the urbit network.  And I don't see a huge advantage to delivering the same
bits this way.

juped:
Again, **"the network" is _entirely irrelevant_ - just assume it doesn't exist**. The question is whether to minimize or maximize the portion of the system that's in the opaque blob (which you can't disassemble without running it first). It's an unfortunate reality that self-hosting compilers require a bit of blobtrust, but even paranoid-fringe crypto nuts like me are mostly resigned to this reality. The additional effort you're asking of every developer (meaning, at this stage in the lifecycle of urbit, everyone) is painful as well.

Maybe you can't pass pointers in a discussion, but you can't expect an identical calling environment on the other side either. I don't have boot sequence nerdview - I suspect you are the only person who will ever spend time inside the boot sequence for any reason other than personal curiosity. The method of system distribution should, therefore, not be determined by the view from deep inside that "well, sane boot sequence noun, what's the problem?". From outside, you're requiring that anyone who wants to boot some source must first manually jam it inside a blob file.

keatondunsford:
I agree with Curtis: the metric to prioritize absolutely is speed. At least for the default boot sequence; if technical users just want the compiler and want the vanes and userspace from source, let's add an additional boot flag or something similar to what we do with fake ships. `- A ./arvo` still pulls `latest.pill` and boots the same, if I remember correctly, requiring a `|reboot` of the sourced `%home`.

Why continue to make onboarding tough if a faster boot sequence could make it better? A ship should boot as fast as possible by default; a default pill should be no different from an [AMI](https://en.wikipedia.org/wiki/Amazon_Machine_Image) or any other snapshot of an OS like all the VPS hosting services provide. If Tlon wants to be a hosting company as well down the line, this seems like the right decision. 

[This stuff does in fact matter](https://glinden.blogspot.com/2006/11/marissa-mayer-at-web-20.html).

juped:
What he does here doesn't make anything faster - the prebuilt checkpoint stuff talked about above would, but that's different.

keatondunsford:
Huh okay. Well let's at least do that. 🚀

ohAitch:
That's what the maintenance urbit.pill is, minus userspace

cgyarvin:
"Anyone who wants to boot some source" is not a workflow I recognize.  If
you want to run some source, you run it in a live, previously booted
urbit.  If you want to edit the boot sequence, you edit the boot sequence.
Yes, very few people will do this.

Whenever I make a boot sequence (ie, a pill), I make it from within urbit.
  The files I use are source files within urbit.  I don't make a boot
sequence from unix.  No one does.

I understand that it's better to distribute source files.  It's trivial to
unpack the pill and repack it in front of the user, Benihana style.  This
is not a huge concern for me.  I understand why you find it relevant,
however.  In my mind, "blobtrust" is atomic.  We need it already, so it's
no big deal to rely on it again.

I think we understand each other and won't learn anything more on this
chain.

juped:
I disclaim any understanding of why you would ever want to do this.

ohAitch:
I disagree that you understand ray

Or at least, my take of his point is that 149.pill can be a /lot/ more
frozen than "any changes in e.g. talk.hoon", and therefore we shouldn't be
using an opaque urbit-specific archive format to bundle the former with the
latter.

> It's trivial to unpack the pill
If you have a running urbit; this goes back to Trusting Trust, which is the
exact thing that providing everything as source was supposed to make more
difficult.

juped:
I endorse that interpretation of my words (also a lot more frozen than "any changes in eyre.hoon", the real pain point)

cgyarvin:
That's why we have over-the-air synchronization.

The coherent build reduces the number of channels by which code is
delivered.  Either it's in the pill or it's over the air.  There is no
third channel.  No third channel is needed.  The first channel -- the pill
-- produces a working urbit.

Possibly one somewhat out of date.  If so, the first thing it does is
upgrade itself.

We have far too much important shit undone to be purity-signaling about the
inspectability of the boot sequence.

ohAitch:
> purity-signaling

I repeat, if purity is not a concern, how miffed would you be if I used black magic to byte-align the source code in the pill, so that it can be viewable in at least a hex editor?

cgyarvin:
Far be it from me to impose on anyone's copious spare time...

cgyarvin:
Sorry, I shouldn't be this snarky.

The boot sequence for a newly created urbit needs to be presented, as a
list of events, by the urbit interpreter (vere).  This list of events can
be generated as a tool within arvo.  From the perspective of arvo, the
simplest thing is to generate the complete sequence and save it as one
jamfile (pill).

This sequence right now contains all userspace files.  In future it will
also contain an initial certificate set (replacing the current built-in
galaxy table, but not limited to galaxies).

The simplest way to implement vere is just to deliver the whole boot
sequence in the package as a 2MB blob.  But the simplest way isn't
necessarily the right way.  Rather, the vere build and distribution process
should do whatever it feels is prudent with the boot sequence that +metal
(ie new +solid) generates.

ohAitch:
And then committed the presumably ugly code, and shipped the resulting
latest.pill's to everyone, I mean. Putting that down as "probably fine"

@juped, would being able to binary-diff the blobs and see the source code
changes assuage any of your concerns? For bonus points, it might be
possible to ship the blob + source as a linker script, bypassing any calls
to libnoun ++jam should you wish to.

juped:
No, it would not. Rather, the vere build and distribution process should do whatever it feels is prudent with the boot sequence. What I think is prudent is for it to load the source file events from source files. I did a quick log check yesterday, and 41% of arvo commits touch the compiler or vanes, while only 17% touch the compiler (and this will go down even further as the language matures and cools down). 41% agony to 17% agony is an improvement, 41% agony to 100% agony is agonizing. We can, and should, get rid of this painful binary-artifact-management headache altogether very easily.

juped:
Superseded by #389

PR #377, opened 2017-03-22
Set key for ~fed.

cgyarvin:


galenwp:
In general, we should probably sign these commits — but I can certify that this one is real.

PR #373, opened 2017-03-20
Enable js urb.bind to a foreign {ship}

ohAitch:
by implementing the "subscription with a mark translation on all results" %peel code path in gall

@cgyarvin requesting sanity-check

PR #378, opened 2017-03-23
Enable js urb.bind to a foreign {ship}

ohAitch:
by implementing the "subscription with a mark translation on all results" %peel code path in gall

(dupe of #373)

### DIFF
--- a/arvo/eyre.hoon
+++ b/arvo/eyre.hoon
@@ -776,7 +776,8 @@
         ::  ~?  ?=($lens r.q.tee)  hen=hen^hcuf=-.cuf
         (get-ack:(ire-ix p.tee) q.tee ?~(p.cuf ~ `[-.cuf u.p.cuf]))
       ::
-          $doff  !!
+      ::  REVIEW handle in gall?
+          $doff  (execute tee (norm-beak -.top) [%vale p.cuf q.cuf])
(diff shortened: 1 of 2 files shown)


ohAitch:
Does the main `%l` part have problems long-term, which I should open an
issue for?

cgyarvin:
%|?  What mean?

juped:
`%l` (Anton tried to monospace it, but web 3.0 tech failed him), not `%|`

ohAitch:
Right, pointers
The majority of this patch is a [new type of rook](https://github.com/urbit/arvo/pull/378/files#diff-3d799378415e1e20fa632da2978b72b3R16), i.e. gall "forward ames message", which handles "translate me later" %peel subscriptions.

ohAitch:
No actually my mistake there was trying to put md formatting in an email reply instead of using the actual Web 3.0 tech :P

ohAitch:
Anyone on a righteous Web 0.5 mail client shouldn't see any non-monospaced text anyway…

ohAitch:
(Also in answer to my own question, it might be causing the defer-mack collisions, I was right that the %doff handling should be in gall somehow, giving it a chance to unregister the crash handler duct)

PR #380, opened 2017-03-24
Enable js urb.bind to a foreign {ship}: v2

ohAitch:
Supercedes #378

In addition to adding remote %peel, this reverts %gall handling of foreign diff results to an earlier version, which incorrectly translates the data using the desk of the local app with the same name. This is caused by an arvo limitation, which necessitates ames negative ackgnolwedgments be given explicitly instead of being generated from a crash.


galenwp:
I'd like to approve this, but I have no idea how.

PR #381, opened 2017-03-25
%~ no longer requires closing ==

hoclun-rigsep:


joemfb:
There are some conflicts with other open PRs: #364 adds `++expx` and #365 adds `++expy`. (#352 changes adjacent lines, but that is a simple resolution...)

It's merely a question of naming the twig parser idioms, but we are running out of letters ...

juped:
Per earlier discussions, `%-` both is n-ary now, and has to be n-ary because of electroplating...

I still think fixing `%-` and reviving `%:` is the right answer.

joemfb:
If I'm understanding this correctly, the twig and expansion is still n-ary, this just changes the parser to be unary.

juped:
That's something you can't do. Then you can never invoke the arm of

```
|_  {a/(container) b/(container)}
+-  operation  [...]
--
```

which is terrible. It's `%-` which is broken by being confusing-inconsistent-ary, not `%~` which is broken by not being confusing-inconsistent-ary.

joemfb:
You're saying this would be break on the wet arm?

```
%~  operation  foo
:-  a=*  b=*
```

joemfb:
linking the discussion on urbit/docs#153

juped:
There's two separate issues

- it's the wrong design, `%-` is a broken abomination
- the wrong design also prevents one from using wet arms, because of electroplating
- no, you can't require people to put their own faces on :)

juped:
(The first is a matter of taste, but the second is a reality in the current implementation)

belisarius222:
@cgyarvin I think you should weigh in on this one.

ohAitch:
And associated "make %- arity two in type and not just practice, bring back
%: for irregular (a b c d e) and tall form %-  a :* b c d e ==  usage"
discussion

My guess is both are righteous because repainting exists now?

ohAitch:
Could this be split into a series of staged commits, each of which is compilable by the previous hoon.hoon and itself? Something like: introduce `%#` rune with desired semantics, change all tall `%~`s to `%#` with corresponding lack of ==, change `%~` semantics to match `%#`, change all `%#`s back to `%~` and remove the `%#` code.

cgyarvin:
That’s normally how I change h.h.

Fang-:
There's instances of `%~` in `hoon.hoon` and `/app/talk.hoon` that aren't addressed in this PR.

Additionally, this needs to be updates to fit into the most recent version of `hoon.hoon`, and as described above be split into multiple PRs that get pushed live sequentially.

Finally, it seems like `%-` and maybe `%:` need some more discussion here. Unclear to me if there should be any changes/additions there.

ohAitch:
%- and %: should be discussed/implemented, but probably not here; afaict the work on this PR is in fact the conflict resolution and phased ++hoon update.

hoclun-rigsep:
I've been out of the loop but preparing to get back into the loop. 100 years ago when I got this to work I did compile hoon with the parser changes and then compile again with the corrected (`==` deleted) use of the rune, and I can obviously split into two commits or two PRs (but why?) if desired. At the time that was comprehensive; I could search for the instances Fang- mentioned getting in since then.

ohAitch:
having it in three commits would be fine, we can merge/deploy them in
sequence and let GH magic figure out the PR is merged when the last one is

joemfb:
@ohAitch, what is in scope for this commit sequence? Anything beyond updating the %cnsg parser to be unary?

keatondunsford:
Superseded by #698, so closing this. Thanks, everyone!

PR #384, opened 2017-04-07
fora fix

galenwp:
See https://github.com/urbit/tree/pull/35


galenwp:
This isn't important, it's going on urbit-org first.

PR #385, opened 2017-04-10
Fix ++spin/++spun

chc4:
Currently, you can't get the changed state out of these, which was a major oversight. This changes ++spin so that it returns a `{(list _-:(b)) _c}` (that is, a list of returned values and the new state) and ++spun so that it just results a `(list _-:(b))`, for when you only need a stateful accumulator.

(I don't know if there is procedure for changing stdlib functions, but these are relatively unused...)

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -471,23 +471,21 @@
   ^+  t.a
   [i.a $(a (skim t.a |=(c/_i.a !(b c i.a))))]
 ::
-++  spin
+++  spin                                                ::  stateful turn
   |*  {a/(list) b/_|=({* *} [** +<+]) c/*}
-  ::  ?<  ?=($-([_?<(?=($~ a) i.a) _c] [* _c]) b)
+  ^-  {(list _-:(b)) _c}
+  =/  acc/(list _-:(b))  ~
   |-
   ?~  a
-    ~
-  =+  v=(b i.a c)
-  [i=-.v t=$(a t.a, c +.v)]
+    [acc c]
+  =^  res/_-:(b)  c  (b i.a c)
+  $(acc (welp acc ~[res]), a t.a, c c)
 ::
-++  spun
+++  spun                                                ::  internal spin
   |*  {a/(list) b/_|=({* *} [** +<+])}
   =|  c/_+<+.b
-  |-
-  ?~  a
-    ~
-  =+  v=(b i.a c)
-  [i=-.v t=$(a t.a, c +.v)]
+  =+  [p q]=(spin a b c)
+  p
 ::
 ++  swag                                                ::  slice
   |*  {{a/@ b/@} c/(list)}


juped:
It is that way on purpose, making this more of a new function. I don't think the value needs to be produced with the sort of things this was meant for, but maybe I'm wrong.

ohAitch:
They look like they're supposed to be a special case of roll for mutating
elements, yeah.

(Also I'm curious what the platonically correct form of line 476 is; the
commented out type assert is a bit obscure, especially since in this case
the assert seems to be going the wrong way, but I'm pretty sure I get the
author's documentational intent. Ideally we could just have "$- except with
a readable sample", and then all then list, parser, etc higher-order
gate samples could be a lot clearer?)
(in particular it would be super clarifying to be able to phrase wet gates
as "use the wet sample to compute a monomorphized dry inner gate, and
invoke that", beyond just using |- because there's no good way to phrase
"cast to gate bivariant on only its +>")

chc4:
Spin/spun were designed as a "stateful turn" - you give it a piece of data that it forwards through the turn. Reel/roll, but instead of returning the accumulator it returns the elements. One of the usecases I had for this when I first wrote it was to forward an ++og instance, to generate a list of random numbers. Hand-rolling code to repeatably call a state machine is pretty annoying otherwise. Currently, you can do this, but then you can't continue to use the rng afterwards since it is still in the state before you called spin/spun. I consider that it's only a problem with ++spin, since that is the one that you provide an explicit instance for.

I have used these functions a few times myself (and am the original author, but I don't think that should actually matter), but I think it makes more sense for spin to also return the bookkeeping data for continuation of data after the function call.

The type hint is bad, yeah. It actually crashes urbit if you try to uncomment it, iirc, but I'm not sure how to do it otherwise unless you use currying so you can use the type of `a` and `c` - if you want to write something like `|*  {a/* b/(list _a)} ...` for example, then you're kinda screwed. Gates just make it worse, since you can't cast them correctly afterwards (could you use `?^`, actually?)

cgyarvin:
"Crashes urbit"?  You mean, the compile fails?

Yes, I think the right way to do this kind of wet gate involves more
currying.  Ultimately I think we'd like a list core that works more like
++in and ++by.  It is hard to look at Hoon today and realize that the
reason ++turn and ++by are so different is that the former is considerably
older than the latter.  Prehistory shouldn't matter this much.

I think my only objection to this patch is that it should come with a lot
more documentation.  It's in an old, poorly documented part of hoon.hoon,
so clear, well-documented code would stick out like a sore thumb.  All the
better... it's the rest of the hand that's sore.

ohAitch:
The code is also older because lists are considerably more frequent than
maps, and it's obvious which of `(roll (turn "abc" dec) add)` and `((~(rep
ly (~(run ly "abc") dec)) add)` is easier to write, read, and follow.

cgyarvin:
All this is true, Anton.  I'm still not satisfied with it though.

ohAitch:
I think the syntax I proposed last time this came up was something like `(run~ly "abc" dec)`, though having `~foo` appear outside of a coin literal seems iffy

chc4:
No, I'm it actually crashes vere with the type hint. I don't remember the actual error code, however. The same thing happened a week ago when I tried to fish a gate type, so I assume it's still a problem.

It's a shame cores are a bit unwieldy, or else you could do something like `abet:(roll add):(turn dec):(toly "abc")` - which would probably have terrible performance and you can't store the raw ++ly in app state or a map without the kernel going out of date.

By add docs, do you mean adding a section to the docs repo for this? None of the stdlib uses the vane-style headers, so it would look extremely odd...

cgyarvin:
Can you put together a reproduction case?  I could but it sounds like it'd
be easier for you.  I don't like to let crashes fall through the cracks.

Yes, I mean documenting it inline using the latest fashions.  I don't care
if it looks extremely odd.

cgyarvin:
This needs to be looked at again with the new repainting model.  Postponing it once more, sadly.

Fang-:
Do we still want this in? If yes, this needs to be looked at again as mentioned above.

keatondunsford:
Status?

keatondunsford:
Oops, nevermind, didn't see #649 above.

PR #396, opened 2017-05-26
tree@da7808a actually disable post box upon submission

ohAitch:
urbit/tree@da7808a actually disable post box upon submission


galenwp:
@ohAitch did you deploy this?

PR #404, opened 2017-06-18
adds warnings and a confirmation prompt to %/gen/moon

joemfb:
I've seen #327 affect many new urbit users -- this seems like the simplest temporary solution.


Fang-:
Looks great, thanks!

PR #405, opened 2017-06-19
Honor previously-heard %received notifications

ohAitch:


ohAitch:
signed off by @cgyarvin

PR #406, opened 2017-06-20
adds error message to +ticket

joemfb:
This is another frequent question on `urbit-meta` -- the current error message ("blocking not care: ~.a") is both wrong and unhelpful.


juped:
Request analogous patch for 0.5 if you have time

PR #408, opened 2017-06-28
eagerly evaluate "eyre: execute" error info

ohAitch:
to avoid issues from lazy traps in defunct requests

(cherry picked from commit 46320bf9d9579508fa0492b88ab055979929e049)

### DIFF
--- a/arvo/eyre.hoon
+++ b/arvo/eyre.hoon
@@ -960,7 +960,7 @@
     |=  {tea/whir bek/beak sil/silk}
     %+  pass-note  tea
     :^  %f  %exec  our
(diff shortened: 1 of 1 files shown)


juped:
Review requested, and do we want this on 0.5, Anton?

ohAitch:
Mostly we want it on urbit-org, where it's already deployed; also wouldn't mind review, the obvious thing is to just change the `%dude` silk to be `tank` instead of `(trap tank)`, but surely there's some reason for the latter.

belisarius222:
This already got merged into master https://github.com/urbit/arvo/commit/c8d98a12

PR #410, opened 2017-07-01
insert list in list function

joshuareagan:
> (ljam "something" 22 "Something is between ** the stars.")
"Something is between \*something\* the stars."

Seems like a handy function, would be useful particularly for tape manipulation.


galenwp:
I think you want to PR this against maintenance.  It also kind of opens up the whole issue of actual containers in Hoon.

joshuareagan:
Sounds good!  I'd probably better hold off on making any pull requests, knowing as little as I do.

galenwp:
@cgyarvin @eglaysher, you guys want to weigh in here?  I think this is just touching the tip of an interesting issue.

chc4:
[unsolicited opinion warning]

Anecdotally, I've used the body of this function many times and wished that it existed. The name seems kinda bad, though. `lace` maybe?

re: Containers, my belief is for a list core to provide an stdlib function that turns a list to a ++lo or whatever, and then having functions in it that map lo -> lo and one that returns the list. Something like `abet:(turn |=(a/@ (scow %p a))):(turn |=(a/@ (mul a 2)):(loan ~[0 1 2])`, which is a lot easier to read than needing tons of `~(turn lo (~(turn lo a) ...) ...)` like maps do. The other option is Anton's `(turn~lo (turn~lo a ...) ...)` syntax (for normal function call ordering). Just sticking all the current list functions in a namespace doesn't do anything except make everyone type `:lo` a lot, for no benefit.

joshuareagan:
`~sorreg-namtyv` kindly asks that we move the discussion to [fora](https://urbit.org/fora/posts/~2017.7.1..18.04.52..f098~/).

PR #412, opened 2017-07-07
escape </script XSS

ohAitch:


cgyarvin:
Fix already applied on a different branch.

PR #413, opened 2017-07-08
don't link moons to parent's talk on start up

ohAitch:


galenwp:
I'd like to test this one. Would that just involve ticketing all the way down to a moon on a fake network?

cgyarvin:
Invalidated by later changes.

PR #414, opened 2017-07-14
lower @f priority so '.nan' parses properly

ohAitch:


belisarius222:
This had already been fixed in the master branch.

cgyarvin:
It got merged in the master branch, but by hand rather than with the merge button.

PR #415, opened 2017-07-19
some changes to get :twit working

joemfb:
I finally got around to trying out `:twit`. The `++peek` interface was out-of-date*, as were the `|as` and `|feed` generators.

\* The `care` is now passed as part of the path, and the last path segment is ignored (presumed to be a mark?).

I had another problem trying to post: the negotiated access token was read-only and couldn't be upgraded. I created a new read-write access token, but can figure out how to get %eyre to stop using the old one. Is there currently a way to manually configure (or force a request for) a new token?

I also deleted the github driver config.

### DIFF
--- a/app/twit.hoon
+++ b/app/twit.hoon
@@ -29,7 +29,7 @@
   $%  {$quit $~}                                         ::  terminate
       {$diff gilt}                                      ::  send data
   ==
-++  gilt  
+++  gilt
   $%  {$twit-feed p/(list stat)}                        ::  posts in feed
       {$twit-post p/stat}                               ::  tweet accepted
       {$ares term (list tank)}                          ::  error
@@ -97,7 +97,7 @@
   |=  {pax/path mof/(list move)}  ^+  done
   =^  tym  ran  (dely pax)
   :_  +>.$
-  ?~  tym  
+  ?~  tym
     :: ~&  no-wait/ran
     mof
   :: ~&  will-wait/u.tym
@@ -124,7 +124,7 @@
       pus=(~(gas ju *(jug path bone)) (turn ~(tap by sup) .))
   ?~  (~(get ju pus) pax)
     ~
-  ~&  peer-again+[pax ran]  
+  ~&  peer-again+[pax ran]
   (pear | `~. pax) ::(user-from-path pax))
 ::
 ++  sigh-recoverable-error                              ::  Rate-limit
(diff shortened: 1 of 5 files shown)


joemfb:
`~poldec-tonteg` pointed me to `++discard-state` (I should check the docs next time ...)

joemfb:
New problem: any idea why a 200 response is being cast to %recoverable-error? And did there used to be an `%ares` mark?

```
> :twit|feed %joemfb
[%peek /home/joemfb]
>> [%camping %g %x [p=~master-morzod q=%twit r=[%da p=~2017.7.20..02.41.00..8038]] s=/~/joemfb/home]
[%peek /home/joemfb]
[%add-auth-header "https://api.twitter.com/1.1/statuses/home_timeline.json"]
[%sigh-tang /scry/home/joemfb]
%need
ford: casting %httr to %twit-feed
ford: cast %twit-feed
/~master-morzod/home/0/mar/recoverable-error:<[14 18].[19 7]>
/~master-morzod/home/0/mar/recoverable-error:<[15 5].[19 7]>
/~master-morzod/home/0/mar/recoverable-error:<[16 5].[19 7]>
/~master-morzod/home/0/mar/recoverable-error:<[16 14].[16 40]>
[%non-recoverable 200]
/~master-morzod/home/0/mar/recoverable-error:<[16 37].[16 39]>
ford: casting %httr to %recoverable-error
ford: cast %recoverable-error
ford: out of options
eyre: execute [%si ~]
fame: no /~master-morzod/home/25/mar/ares
ford: casting %ares to %~
ford: cast %~
ford-scry-made-fail
```

/cc @ohAitch

joemfb:
Those mark errors were caused be a `%need` crash in the `lib/twitter` parser. It wasn't accounting for newlines.

app/twit.hoon
```diff
@@ -195,11 +195,11 @@
...
 ++  compat  |=({usr/(unit iden) req/(unit iden)} ?~(req & =(usr req)))
 ::
-::  .^(twit-feed %gx /=twit=/~/home/urbit_test)
-::  .^(twit-stat %gx /=twit=/~./post/0vv0old.0post.hash0.0000)
-++  peek
-  |=  {ren/care pax/path}  ^-  (unit (unit gilt))
-  ?>  ?=($x ren)  ::  others unsupported
+::   /+  twitter
+::  .^((list post:twitter) %gx /=twit=/home/urbit_test/twit-feed)
+::  .^(post:twitter %gx /=twit=/post/0vv0old.0post.hash0.0000/twit-feed)
```
joemfb: I can't the `$post` portion of the namespace to work, and I haven't figured out what's going on yet. But it'll need a different mark too

cgyarvin:
It looks like this got resolved by the sec/fixes #443 checkin to master.  Please LMK if that's not the case.

joemfb:
There's some extra bugfixes in here that aren't captured in that other PR. But there's also merge conflicts. Should I resolve them here, or in a new PR?

belisarius222:
Note that this is a PR against maint, so we'll want to change that to be merged into master.

joemfb:
This is now rebased and pointed to master. I still need to test it, should have time this evening.

joemfb:
As @belisarius222 noted in 0b253ec, there's been a regression in the `lib/oauth1` (at least for twitter). Requesting an access token results in a 401/Error 32.

I've re-tested on maint, and everything still works there. The requests from master *look* the same as the requests from maint (parameters, encoding, order, etc.), and diff doesn't turn up anything obvious, just refactoring.

The next step will be to actually verify the values in the oauth header. Maybe something changed in the hmac implementation?

joemfb:
False alarm, it was the content-type header. Looks like it was originally a find/replace error, but my correction was not correct. The driver is working now.

joemfb:
As is the app.

ohAitch:
I was wondering what just yanking that out would break!

joemfb:
Removing the content-type header entirely results in the same 401/Error 32.

ohAitch:
Yup, and just that happened in urbit/urbit#890

joemfb:
Similar, but that's the Accept header. Removing content-type but adding accept in `lib/oauth1` also gives the same 401.

belisarius222:
@joemfb so this still works, right? Just checking. If it does, I'll merge it.

joemfb:
It did as of 12/05 -- I haven't looked since. I can't think of any subsequent changes that would impact here.

PR #417, opened 2017-08-08
correct an oversight in ++nest:ut (mismatch with jet)

frodwith:


belisarius222:
@cgyarvin this is kosher, right? If it is, let's get it merged.

cgyarvin:
Yeah, I thought that was merged already.

PR #426, opened 2017-08-28
tree@f7e376e

galenwp:
Fix sorting, fora locking, and twitter handle.  Not critical.  Already deployed to urbit.org, but others may want these small fixes too.


cgyarvin:
Also being merged into master.

PR #428, opened 2017-09-15
Fix ford ship and case parsing

belisarius222:
`/+` and `/-` are supposed to be able to parse ship and case, but the parsing was broken. Docs for this behavior are part of [this docs PR](https://github.com/urbit/docs/pull/192), which should be merged after this PR is merged.


cgyarvin:
Also manually merged into master.

PR #429, opened 2017-09-23
count and print core arms

joemfb:
gotta get my stats back up


joemfb:
this has been resolved independently

PR #431, opened 2017-09-28
Trim any trailing spaces from lines in dojo

### LINKED ISSUE #430: dojo fails on trailing spaces
If you manage to get a trailing space typed into your dojo (which you can do by inserting some characters in the middle of a line then deleting the characters after a space), when you hit enter, dojo will not process the command, and instead triggers the terminal bell. This took a minute to figure out, and I don't see why dojo should fail based on this, especially since hoon in general allows trailing whitespace.

This might be an easy fix suitable for a beginner.

ericfode:
Trim any trailing spaces on the line. fixes #430

### DIFF
--- a/app/dojo.hoon
+++ b/app/dojo.hoon
@@ -789,7 +789,9 @@
   ++  he-duke                                           ::  ++he-dope variant
     |=  txt/tape
     ^-  (each (unit (each dojo-command tape)) @ud)
-    =+  foy=(he-dope txt)
+    =+  trim-space=;~(pfix spac:poja (star next))        
+    =+  trimed-text=(flop (scan (flop txt) trim-space))
+    =+  foy=(he-dope trimed-text)
     ?-  -.foy
       $|  [%| q.p.foy]
       $&  [%& p.foy]
--- a/app/dojo.hoon
+++ b/app/dojo.hoon
@@ -113,7 +113,7 @@
       |=  {gol/goal mod/dojo-model}  ^-  dojo-command
       [[%poke gol] [0 [%ge mod(q.p [q.gol q.p.mod])]]]
     ::
-    ++  dp-command-line  ;~(sfix dp-command (just '\0a'))
+    ++  dp-command-line  ;~(sfix dp-command (star ace) (just '\0a'))
     ++  dp-variable                                     ::  %verb or %brev
       |*  {sym/rule src/rule}
       %+  cook  
@@ -789,9 +789,7 @@
   ++  he-duke                                           ::  ++he-dope variant
     |=  txt/tape
     ^-  (each (unit (each dojo-command tape)) @ud)
-    =+  trim-space=;~(pfix spac:poja (star next))        
-    =+  trimed-text=(flop (scan (flop txt) trim-space))
-    =+  foy=(he-dope trimed-text)
+    =+  foy=(he-dope txt)
     ?-  -.foy
       $|  [%| q.p.foy]
       $&  [%& p.foy]


belisarius222:
Thanks for fixing this!

I think we should use `ace` from hoon's standard library instead of `spac:poja`. `spac:poja` matches tabs and CR/NL characters in addition to `' '`, but I think we don't actually want to trim off anything other than trailing `' '` characters. The dojo won't accept tabs as far as I can tell, so we shouldn't need to match those. As for newlines, the only time the dojo will accept a newline is if it's in the middle of a tall-form hoon expression, in which case I don't think it matters whether the newline is trimmed or not.

Maybe I'm misunderstanding something here, though?

Also, a minor note: `trimed` should be spelled `trimmed`.

joemfb:
+1 to @belisarius222's comments.

I'm a little worried about performance though; `:dojo` is already a little challenged in that area, and this would reverse the buffer twice on every keystroke.

Here's an alternate approach (untested):

`++sfix` is a parser combinator that discards the result of the second rule (and therefore, subsequent rules if present):

```
> (scan "b " ;~(sfix alf ace))
'b'
> (scan "b a" ;~(sfix alf ace alf))
'b'
```

It's already being used in `++dp-command-line` to discard trailing newline characters. So I think it would work to add `(star ace)` there:

```
> (scan "b  a" ;~(sfix alf (star ace) (just 'a')))
'b'
```

belisarius222:
@joemfb I had the same thought about performance, and `++sfix` instead of `++pfix`. It seemed a little tricky to get `sfix` to run on the whole string, though ... we really do need to check the last characters of the string. I've been trying to think of a parser combinator that wouldn't just stop at the first space, but haven't thought of one yet.

I tried your snippet with an internal space and sure enough it errored out:
```
> (scan "b  a  " ;~(sfix alf (star ace)))
{1 4}
'syntax-error'
```
If we're concerned about the performance here, I have a feeling a call to `++roll` might be the way to go instead of a parser combinator.

I'm not sure flipping and reversing the tape is actually expensive, though, especially since no matter how you do this, you'd have to iterate through the tape to get to its tail anyway (although it's true you wouldn't have to iterate all the way backward or build a new list from scratch, I think), and `++flop` is jetted. I would say our debugging of dojo performance should be primarily empiricist rather than rationalist, i.e. guided by profiling as opposed to reasoning, unless you're supremely confident that `++flop` would cause a performance bottleneck here.

ohAitch:
The dojo performance problems are undoubtedly from it trying to parse
things every keystroke without debouncing in the first place, contrast how
fish shell pasting in a block will flash black before getting syntax
coloured.

joemfb:
I have no objections to empirical performance analysis! It should absolutely be done, and I should have some time to run this later today.

But I think the it's clear that double `++flop`s are unnecessary. And it looks to me like `++sfix` runs to the end:

```
> (scan "c  " ;~(sfix alf (star ace) (just 'a')))
{1 4}
'syntax-error'
> (scan "c  a" ;~(sfix alf (star ace) (just 'a')))
'c'
```

belisarius222:
~pittyp and I ran some tests in the dojo, and `++flop` takes about 3 microseconds for a fairly long (77-character) dojo expression, averaged over 1 million iterations.

~pittyp also pointed out to me that it might work to modify `++dp-command-line` like so:
`++  dp-command-line  ;~(sfix dp-command (star ace) (just '\0a'))`

If that works, I agree it would be cleaner. @ericfode do you think you could try it out that way and see if it works right?

ericfode:
Yes! In progress.

ericfode:
Yep looks like it works to me! :+1:

PR #433, opened 2017-10-04
removes unnecessary parentheses from ++tap:by in +glass and +brass

joemfb:


Fang-:
Much cleaner, thanks!

PR #445, opened 2017-10-24
Fix small issue in ++op:dejs-soft:format when calling ++my:nl

### LINKED ISSUE #444: [remaint] Find error in ++my:nl when calling ++op:dejs-soft:format
On the latest master commit (dbc802a9) with the latest pill (4018c1b5), triggered by [this call](https://github.com/Fang-/arvo/blob/31cd0593aa4150795d67c8688de3a3c0992c1ef0/lib/talk-json.hoon?utf8=%E2%9C%93#L379) to `++op` (`:dejs-soft:format`), I get a stack trace and the following lines:

```
%redo-match
####
-redo: dext: ref.{a/nlr({p/* q/*}) <1.dth 1.rlo 1.hru $143>}
####
-redo: dext: ref.{a/nlr({p/* q/*}) <1.dth 1.rlo 1.hru $143>}
####
-redo: dext: ref.a/nlr({p/* q/*})
####
-redo: dext: ref.a/nlr({p/* q/*})
####
-redo: dext: ref.nlr({p/* q/*})
####
-redo: dext: ref.nlr({p/* q/*})
####
-redo: dext: ref.nlr({p/* q/*})
####
-redo: dext: ref.nlr({p/* q/*})
-find.p.i.+2.a
find-fork-d
```

It seems [this call](https://github.com/urbit/arvo/blob/master/sys/zuse.hoon?utf8=%E2%9C%93#L3338) to `++my:nl` messes up.  
A temporary fix for this is to add faces (`p` and `q`) to the items in list `a` there (ie `a/(list _[p=(wonk *fel) q=(need *wit)])`), but seeing as how other existing calls to `++op` work fine, this shouldn't be necessary.  
An alternative fix is to make `++my:nl` use lark instead of `p` and `q`, but it's working with a `(pair)` so it really shouldn't need to?

To reproduce, get [Fang-/arvo/remaint-merge-talk](https://github.com/Fang-/arvo/tree/remaint-merge-talk), and then run in dojo:  

```
&talk-prize [%friend ~]
````

Fang-:
Fixes #444

### DIFF
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -3334,7 +3334,7 @@
     ++  op                                              ::  parse keys of map
       |*  {fel/rule wit/fist}
       %+  cu  
-        |=  a/(list _[(wonk *fel) (need *wit)])
+        |=  a/(list (pair _(wonk *fel) _(need *wit)))
         (my:nl a)
       %-  ci  :_  (om wit)
       |=  a/(map cord _(need *wit))


Fang-:
Seems like this got reversed in [7df058d](https://github.com/urbit/arvo/commit/7df058dd6fc1878d435823fa20487aeec992780f#diff-8f1f6f2de23bb6685b787c7186a0221eR3337) "Reset %zuse pre 143.". Don't think that was intentional?

cgyarvin:
Definitely not, sorry!

Issue #442, opened 2017-10-23
%eyre via lib/oauth2 not sending headers in post requests?


mattlevan:
# Problem
The code (likely some part of %eyre, I suppose) which [`++post-quay` of arvo/lib/oauth2](https://github.com/urbit/arvo/blob/a68497631aa728659113a12e4f33cec873088fe3/lib/oauth2.hoon#L13) relies on to send POST requests is not attaching headers correctly. 

# Background
As part of my arvo docs work, I have been going through the [security driver tutorial here](https://github.com/urbit/docs/blob/master/docs/arvo/security-driver.md), which guides the reader through configuring OAuth2 for use with the [GitHub API v3](https://developer.github.com/v3/).   

# Issue Reproduction
Run the instructions from the [OAuth2 section](https://github.com/mattlevan/docs/blob/arvo-finish/docs/arvo/security-driver.md) of my `arvo-finish` branch of the docs. The only things I've changed are the URLs to enter into GitHub when you make an OAuth app and the `access_token` API URL (from https://github.com/login/oauth/api/access_token to the correct https://github.com/login/oauth/access_token). 

## Network Trace
The reader will notice things running smoothly: `|init-oauth2` performs as expected, etc. Upon reaching the part in which the reader is instructed to run `+https://api.github.com/user` in the `dojo`, he will receive an appropriately formatted GitHub authorization URL (https://github.com/login/oauth/authorize?state&client_id=<client id>&redirect_uri=<redirect uri>&scope=client%20admin). The reader's network activity log from the developer console in his browser will look similar to this:

![oauth2-github](https://user-images.githubusercontent.com/4582375/31902112-8fa5e1ce-b7d8-11e7-8a80-deed35b1427e.png)

## %bad-json
He will also be met with a response from his urbit, still in the browser, which indicates "bad json":
```
eyre: execute [%se p=%receive-auth-response q=[~. <|com github|>]]
%need
  /~hopfen-binsec-rosnub-dalhes--marbet-sabwyd-ticdeb-darrul/home/0/lib/oauth2
  <[36 3].[36 40]>
~
  /~hopfen-binsec-rosnub-dalhes--marbet-sabwyd-ticdeb-darrul/home/0/lib/oauth2
  <[35 3].[36 40]>
[ %bad-json
  [ ~
    [ p=78
        q
      'access_token=<ommitted>&scope=&token_type=bearer'
    ]
  ]
]
[...]

PR #446, opened 2017-10-25
Added 'Accept: application/json' hiss header to lib/oauth2

mattlevan:
## Summary
This adds an `Accept: application/json` header to hisses sent from lib/oauth2 and is a solution to urbit/arvo#442. 

## Rationale
The OAuth2 library, as it is currently implemented, requires a JSON response from a service to parse the access token correctly. For this reason, I have added the `Accept: application/json` header to the `++post-quay` arm, as this arm is already used for adding the `Content-Type` header. Without this `Accept` header, some OAuth2 APIs respond with non-JSON formats by default, which the current lib/oauth2 is not prepared to handle, therefore resulting in a `%bad-json` error (see urbit/arvo#442 for more details). 

## Note
1. I've combined *running* and *3-fixed* [forms](http://urbit.org/docs/hoon/syntax/). Is this considered undesirable, in which case I should stick with either/or?
2. HTTP header field names are case insensitive; you'll see that the actual header is `accept` with a lowercase 'a'.

### DIFF
--- a/lib/oauth2.hoon
+++ b/lib/oauth2.hoon
@@ -19,7 +19,14 @@
   |=  {a/purl b/quay}  ^-  hiss
   =.  b  (quay:hep-to-cab b)
   =-  [a %post - ?~(b ~ (some (as-octt +:(tail:earn b))))]
-  (my content-type+['application/x-www-form-urlencoded']~ ~)
+  %^  my
+    :+  %accept
+      'application/json'
+    ~
+    :+  %content-type
+      'application/x-www-form-urlencoded'
+    ~
+    ~
 ::
 ++  mean-wall  !.
   |=  {a/term b/tape}  ^+  !!


cgyarvin:
This isn't how I'd format it but it's perfectly readable, which I think is
the gold standard for right now.

mattlevan:
A man of mystery @cgyarvin! How'd you do it? All running?

cgyarvin:
%^    a
      b
   c
d

ohAitch:
The slight pattern break used elsewhere in the system is
```
%-  my  :~
  :-  %accept
  ~['application/json']
::
  :-  %content-type
  ~['application/x-www-form-urlencoded']
==
```
as non `:~` tall forms for literal `(list)` arguments generally feel weird. I'd describe the motivation as approximately "adding/removing elements only touches those elements' lines".

PR #448, opened 2017-11-04
Added lore and nule

mattlevan:
Stumbled upon these missing entries while porting @asssaf's wiki.hoon to master.


keatondunsford:
Just stumbled upon this porting the Mail example app to Hoon 143. Also added `++role`. Thanks @mattlevan!

PR #451, opened 2017-11-09
Fix small issue in ++op:dejs-soft:format when calling ++my:nl

Fang-:
Re-does PR #445. It got overwritten in 7df058d.

### DIFF
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -3334,7 +3334,7 @@
     ++  op                                              ::  parse keys of map
       |*  {fel/rule wit/fist}
       %+  cu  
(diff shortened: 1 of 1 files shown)


ohAitch:
Shouldn't `++my` just be rewritten to be `^-  (list [p=_-.i q=_+.i])` as it is on maint?

ohAitch:
(or if you want to be a troll, `=>  .(a ^+((turn a |*((pair) +<)) a)`)

belisarius222:
Seems to me we should approve this as-is, and then if we want to change the behavior of `++my:nl` later, we can still do that.

PR #454, opened 2017-11-14
pull up to date unmark into hoon.hoon, mar/umd

ohAitch:


mattlevan:
This is great. As far as I can tell, Udon and Sail rendering is spot on (I've discovered only one quirk which may be intentional: consecutive Udon headers require a blank line separating them) but so-called `%frontmatter` doesn't work yet. We should be able to use the syntax below to achieve the same things as YAML headers did with Markdown files rendered with Tree:

```
;~  title+'Great Title'
    author+'Matt'
    date+~2017.11.19
    anchor+none
    next+true
    sort+1
==
```

belisarius222:
You want a `:~` instead of  `;~` there.

mattlevan:
Indeed thank you Ted.

joemfb:
That's one to note specifically for the eventual high-level compiler error messages. It's a very easy mistake to make, and not exactly easy to catch ...

ohAitch:
Also you want :- :~, see static.umd for an example that hopefully renders
at http://localhost:8080/static

mattlevan:
Awesome, that'll do it @ohAitch. Thanks. For reference, the following frontmatter header is working in one of my documents:

```
:-  :~  title+'Introduction'
        author+'Matthew Levan ~mastyr-bottec matt.s.levan@gmail.com'
        date+~2017.11.17
        anchor+%none
        sort+1
        next+&
    ==
```

Only hiccup I can see at the moment is that the next button rendered by Ford for this page is printing "Next: ~~~4e.ock" instead of "Next: Nock". Using a cord instead of a tape leaves "Next: undefined".

ohAitch:
A tape instead of a cord, you mean. Yeah, it's being escaped to urbit url-safe literal form, I should add an exception for text types.

mattlevan:
@ohAitch Commit [04579db](https://github.com/urbit/arvo/pull/454/commits/04579db47382e715d749428128aa067f0b24c750) takes this format:
```
:-  :~  title+'Test Title'
    ==

;>

# Header 1

;=
  ;list;
==
```

and fails to render anything except the list. This is the same format which renders as before (everything good except titles render to URL safe strings).

joemfb:
Try `(end 3 1 p.a)`

ohAitch:
Yes right

keatondunsford:
Should we make this `/mar/udon`?

PR #464, opened 2017-12-01
enable Travis CI, base test "does dojo run"

ohAitch:
The vere download chain is somewhat fragile, but this seems reasonable to merge ASAP, since any  breakage is going to be confined to false-positive red PR checkmarks


ohAitch:
Updated description to be less "WIP", requesting review and/or merge

.travis/get-or-build-pill.sh
```diff
@@ -0,0 +1,79 @@
...
+fi
+
+mkdir prev
+{
+  echo Pilling: trying pinned fakezod
+  wget -i pin-parent-pill-pier.url -O - | tar xvz -C prev/ &&
+  echo Downloaded prev/zod
+} || {
+  echo Pilling: Parent-pill pier not available, trying preceding pill commit
+  HASH2=$(git -C .. log -2 $HASH --format=%H -- sys/ | tail -1)
```
belisarius222: This is the hash of the previous commit. We should probably name this variable accordingly, like PREV_HASH.

.travis/get-or-build-pill.sh
```diff
@@ -0,0 +1,79 @@
...
+  mv zod prev/zod &&
+  export PIER_FRESH="y"
+} || {
+  echo Pilling: Out of ideas
+  exit 1
+}
+
+lsc <<done
+do
+  require! <[ stream-snitch once recursive-copy wait-on ]>
```
belisarius222: This is a long enough code block that it should be in its own livescript file.

.travis/get-or-build-pill.sh
```diff
@@ -0,0 +1,79 @@
...
+    console.log "\n\n---\nnode: detected error\n---\n\n"
+    set-timeout (-> process.exit 1), 1000
+  
+  <- on-next /dojo> /
+  {PILL_NAME} = process.env
+  do-pill = ->
+    urbit.write "|label %home %#PILL_NAME\r"
+    urbit.write ".urbit/pill +solid /==/#PILL_NAME/sys, =dub &\r"
+    <- wait-on resources: <[ prev/zod/.urb/put/urbit.pill ]>
+    urbit.write "\04"
```
belisarius222: Ctrl-D to kill the process. Comment would be nice here.

.travis/get-or-build-pill.sh
```diff
@@ -0,0 +1,79 @@
...
+  require! <[ stream-snitch once recursive-copy wait-on ]>
+  pty = require \pty.js
+  
+  urbit = pty.spawn './urbit' <[-FI zod prev/zod]>
+             .on \data -> process.stdout.write it
+
+  on-next = (re,cb)->
+    urbit.pipe (new stream-snitch re).on \match once cb
+
+  on-next /\n(\/~|ford: )/ ->
```
belisarius222: needs a comment indicating it's for handling errors.

.travis/test.ls
```diff
@@ -0,0 +1,30 @@
...
+urbit.on \exit (code)->
+  console.log "\nnode: urbit exited with code #code\n"
+  process.exit code
+
+fin = no
+urbit.pipe (new stream-snitch /dojo> /g).on \match ->
+    return if fin
+    fin := yes
+    console.log "\n\n---\nnode: got dojo!\n---\n\n"
+    set-timeout (-> process.exit 0), 1000 # should probably test further
```
belisarius222: There should be a comment explaining why we do this in a setTimeout instead of just immediately (it's to give Urbit time to finish printing whatever it's printing).

belisarius222: Also, this is a time-based heuristic approach. Maybe we could make this more rigorous.

.travis/test.ls
```diff
@@ -0,0 +1,30 @@
...
+    
+urbit.pipe (new stream-snitch /\n(\/~|ford: )/g).on \match ->
+    return if fin
+    fin := yes
+    console.log "\n\n---\nnode: detected error\n---\n\n"
+    set-timeout (-> process.exit 1), 1000
+    
+set-timeout ...
```
belisarius222: This should be removed to let travis handle the timeout. If we do leave it in here, we should have it exit.

gen/solid.hoon
```diff
@@ -11,26 +11,44 @@
   !:
 :-  %say
 |=  $:  {now/@da eny/@uvJ bec/beak}
-        {$~ $~}
+        {arg/$@($~ {top/path $~}) dub/_|}
```
belisarius222: This could use a doccord explaining that `top` is the top-level arvo directory, defaulting to `/=home=/sys`. `dub` is a flag that will cause `+solid` twice to catch bugs in the hoon compiler.

gen/solid.hoon
```diff
@@ -11,26 +11,44 @@
...
+      ~&  %solid-double-parsed
+      =+  one=(~(mint ut %noun) %noun gen)
+      ~&  %solid-double-compiled
+      =+  two=(~(mint ut p.one) %noun (rain arp rax))
+      ~&  %solid-arvo
+      [7 q.one q.two]
+      '''
+    =+  all=.*(0 q.one)
+    .*  all
+    :+  7  =<(+ .*(all [9 2 0+2 1+[p.one tri] 0+7]))
```
belisarius222: Desperately needs some comments.

gen/solid.hoon
```diff
@@ -11,26 +11,44 @@
...
-        {$~ $~}
+        {arg/$@($~ {top/path $~}) dub/_|}
     ==
 :-  %noun
-=+  top=`path`/(scot %p p.bec)/[q.bec]/(scot %da now)/sys
-=+  pax=`path`(weld top `path`[%hoon ~])
-=+  arp=`path`(weld top `path`[%ovra ~])
+?~  arg  $(arg ~[top=`path`/(scot %p p.bec)/[q.bec]/(scot %da now)/sys])
+=+  pax=`path`(weld top.arg `path`[%hoon ~])
+=+  arp=`path`(weld top.arg `path`[%ovra ~])
```
belisarius222: This is a little weird. Wouldn't we want this to be `%arvo`, not `%ovra`? @cgyarvin

gen/solid.hoon
```diff
@@ -11,26 +11,44 @@
...
 ~&  %solid-start
 =+  txt=.^(@t %cx (weld pax `path`[%hoon ~]))
 =+  rax=.^(@t %cx (weld arp `path`[%hoon ~]))
 =+  ^=  ken
-    =-  ?:(?=($& -.res) p.res ((slog p.res) ~))
+    =-  ?:(?=($& -.res) p.res (mean (flop p.res)))
```
belisarius222: crash here with the error from inside the virtualization context.

gen/solid.hoon
```diff
@@ -11,26 +11,44 @@
...
     ~&  %solid-parsed
     =+  one=(~(mint ut %noun) %noun gen)
     ~&  %solid-compiled
-    =+  two=(~(mint ut p.one) %noun (rain arp rax))
-    ~&  %solid-arvo
-    [7 q.one q.two]
+    ?.  dub
+      =+  two=(~(mint ut p.one) %noun (rain arp rax))
+      ~&  %solid-arvo
+      [7 q.one q.two]
```
belisarius222: I think variable names indicating that `one` is really `hoon` and `two` is really `arvo` would be helpful.

gen/solid.hoon
```diff
@@ -11,26 +11,44 @@
...
+      |=  [pax=path txt=@t arp=path rax=@t]
+      =+  gen=(rain pax txt)
+      ~&  %solid-double-parsed
+      =+  one=(~(mint ut %noun) %noun gen)
+      ~&  %solid-double-compiled
+      =+  two=(~(mint ut p.one) %noun (rain arp rax))
+      ~&  %solid-arvo
+      [7 q.one q.two]
+      '''
+    =+  all=.*(0 q.one)
```
belisarius222: `all` is the hoon compiler gate.

gen/solid.hoon
belisarius222: Calls the hoon compiler gate, replacing the sample with the type of hoon.hoon (`p.one`) and the cord containing the code to run the second `|solid`. This produces a pair of type and nock, so we grab just the nock using `+`.

gen/solid.hoon
```diff
@@ -11,26 +11,44 @@
...
+    .*  all
+    :+  7  =<(+ .*(all [9 2 0+2 1+[p.one tri] 0+7]))
+    [9 2 0+2 1+[pax txt arp rax] 0+7]
```
belisarius222: The second `|solid` is also a gate, so call it by replacing its sample with the actual arguments.

belisarius222:
The `gen/solid.hoon` code could use some more comments, but let's fix that post-breach. It's worth it to get this merged now.

PR #473, opened 2017-12-09
[CI] use urbit_amd64.deb from urbit/urbit release

ohAitch:


.travis/pin-urbit-release.url
```diff
@@ -1 +0,0 @@
-https://github.com/urbit/urbit/releases/download/redefault-0.5.1
```
belisarius222: I liked having a pinned release URL in a single file. Why move it to inline?

ohAitch:
Fair, I think I was aiming for something like "make it more of a hassle to
push an arvo that doesn't work on the deployed network" which is never a
good motivation

(Or like, makes sense on master, but when we inevitably get a next-breach
branch it should keep the old convention)

belisarius222:
Ping. Can you please reinstate the `pin-urbit-release.url` file? Then I'll merge this.

ohAitch:
Right, thanks

belisarius222:
@ohAitch any idea why the build failed? It looks like it couldn't apt install the .deb file for some reason.

ohAitch:
yup, I mixed up wget -o and -O, seems to be going fine now

PR #475, opened 2017-12-09
Fix erroneous presence handling

Fang-:
Currently, hall doesn't check to make sure that diffs it gets via a %present command, or a %status or %config rumor have any effect. This can lead to diffs being spread around to subscribers when nothing has actually changed.

This implements checks for catching those scenarios, and ignores the diff when they occur.

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -1562,6 +1562,8 @@
           $status
         %_  +>
             locals
+          ?:  ?=($remove -.dif.det)
+            (~(del by locals) who.det)
           %+  ~(put by locals)  who.det
           %+  change-status
             (fall (~(get by locals) who.det) *status)
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -797,26 +797,51 @@
       ::    ==
       ^+  +>
       ?-  -.rum
-        $new      ?:  =(src so-cir)
-                    (so-config-full ~ cof.rum)
-                  $(rum [%config src %full cof.rum])
         $bear     (so-bear bur.rum)
         $peer     (so-delta-our rum)
         $gram     (so-open src nev.rum)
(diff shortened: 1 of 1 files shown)


app/hall.hoon
```diff
@@ -797,26 +797,51 @@
...
+        ::  we only subscribe to remotes' configs.
+        ?.  =(src cir.rum)
+          ~!  %unexpected-remote-config-from-remote
+          !!
+        =/  old/(unit config)
+          ?:  =(cir.rum so-cir)  `shape
+          (~(get by mirrors) cir.rum)
+        ::  ignore if it won't result in change.
+        ?.  ?|  &(?=($remove -.dif.rum) ?=(^ old))
+                ?=($~ old)
```
belisarius222: If `old` is `~`, wouldn't we want to take the change, not ignore it?

belisarius222: Oh, I see. If the request was to remove something, but there was nothing there to begin with, then we should ignore the request.

app/hall.hoon
```diff
@@ -916,6 +915,21 @@
     ::>    arms that make miscellaneous changes to this story.
     ::+|
     ::
+    ++  so-present                                      ::<  accept status diff
+      |=  {who/ship dif/diff-status}
+      ^+  +>
+      ::  only have presence if you have write permission.
+      ?.  |((so-admire who) ?=($remove -.dif))  +>
```
belisarius222: If `who` sends a `$remove` command without write permissions, then shouldn't we ignore it or send back a permissions error? If I'm reading this code correctly, i we get a `$remove` from someone who does not have write permissions, we might still perform the removal.

Fang-: If someone has been banned, they still need to have their presence removed. Hence why the permissions check doesn't apply there.

PR #481, opened 2017-12-11
Rebuild msg number references, fix msg updating, provide debug tools

### LINKED ISSUE #477: Leaving a circle and rejoining it from a different star %ap-lames
Because of the federation problems, I had joined ~marzod/urbit-meta instead of just /urbit-meta. When I tried to switch, I got the following error:

```
--------------| ;leave ~marzod/urbit-meta
--------------| ;join /urbit-meta
--------------| :: onn /urbit-meta
--------------| bound '\'' {[hos=~binzod nom=%urbit-meta]}
[ ! /~ponnys-podfer/home/0/app/talk:<[2.187 3].[2.188 27]>
  ! /~ponnys-podfer/home/0/app/talk:<[2.188 3].[2.188 27]>
  ! /~ponnys-podfer/home/0/app/talk:<[2.188 11].[2.188 27]>
  ! /~ponnys-podfer/home/0/app/talk:<[312 5].[325 7]>
  ! /~ponnys-podfer/home/0/app/talk:<[313 5].[325 7]>
  ! /~ponnys-podfer/home/0/app/talk:<[324 7].[324 33]>
  ! /~ponnys-podfer/home/0/app/talk:<[331 5].[388 7]>
  ! /~ponnys-podfer/home/0/app/talk:<[332 5].[388 7]>
  ! /~ponnys-podfer/home/0/app/talk:<[336 7].[336 29]>
  ! /~ponnys-podfer/home/0/app/talk:<[427 5].[431 26]>
  ! /~ponnys-podfer/home/0/app/talk:<[428 5].[431 26]>
  ! /~ponnys-podfer/home/0/app/talk:<[429 5].[431 26]>
  ! /~ponnys-podfer/home/0/app/talk:<[431 5].[431 26]>
  ! /~ponnys-podfer/home/0/app/talk:<[449 5].[454 28]>
  ! /~ponnys-podfer/home/0/app/talk:<[449 9].[449 29]>
  ! /~ponnys-podfer/home/0/app/talk:<[449 13].[449 29]>
  [%ap-lame %talk %diff-hall-rumor]
]
[! gall: %hood: pour: no unto: /drum/home/talk  [%ap-lame %hood %unto]]
```

Over and over again until it started working.

Fang-:
In my haste to get #471 out the door, I completely overlooked the fact that other parts of state are set using `count`. This means that, for however long halls were running with incorrect `count`s, they accumulated incorrect references to messages. This (0e04242) performs a slightly more involved version of what #471 did, rebuilding `known` and `sourced` as well.  
(Note that `sourced` sadly can't be recreated with 100% certainty, but the approximation we do instead should be accurate for all but some rare edge cases. This is not a big deal, because nothing currently depends on it.)

Additionally, logic for updating existing messages was replacing the wrong message due to an off-by-one error. This includes a fix for that.

I've also included a `++poke-noun` arm in both hall and talk. Doing `:hall 'debug'` prints some sanity-check information which will help me identify if an issue relates to message numbering. `:hall 'rebuild'` will perform the same state rebuilding I added to `++prep`, so that users can easily trigger it as a potential band-aid solution in case of trouble.  
Hopefully this won't be necessary anymore, but I'd like to have this available for now just in case.

Should resolve #477 for people who are or have experienced it.

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -1636,8 +1636,8 @@
       ::  changed message
       %_  +>.$
         grams    %+  welp
-                 (scag (dec (max u.old 1)) grams)
-                 [gam (slag u.old grams)]
+                 (scag u.old grams)
+                 [gam (slag +(u.old) grams)]
       ==
     ::
     ++  sa-change-remote                                ::<  apply remote's delta
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -2554,4 +2554,43 @@
       our.bol
       (foal:space:userlib paf [%hall-telegrams !>(-)])
   ==
+::
+::TODO  for debug purposes. remove eventually.
+++  poke-noun
+  |=  a/@t
+  ^-  (quip move _+>)
+  ?:  =(a 'debug')
+    =-  ~&(- [~ +>.$])
+    %-  ~(urn by stories)
+    |=  {n/naem s/story}
+    =+  %-  ~(rep by known.s)
+      |=  {{u/serial a/@ud} k/@ud m/@ud}
+      :-  ?:((gth a k) a k)
+      ?:  =(u uid:(snag a grams.s))  m
+      ~?  (lth m 3)
+        :*  [%fake a u]
+            [%prev uid:(snag (dec a) grams.s)]
+            [%real uid:(snag a grams.s)]
+            [%next uid:(snag +(a) grams.s)]
+        ==
+      +(m)
+    :^  count=count.s
+        lent=(lent grams.s)
+      known=k
+    mismatch=m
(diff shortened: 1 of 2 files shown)


Fang-:
This doesn't change the fact that due to incorrect message numbers being sent out to subscribers, `sequence` (which is used to remember the last-heard message number, so we don't have to request everything if we re-subscribe after unsubscribing) may contain incorrect values.  
In cases where the reported message numbers were higher than their actual, correct ones, this may cause a `;leave` followed by a `;join` to not result in new messages right away. Workaround would be to do `;join ~some/circle/now` instead, making the subscription look at message dates rather than numbers.

PR #482, opened 2017-12-11
Enabled doccords in hall.

eglaysher:
This goes through the new hall files and makes the doccords syntax compile.

@Fang-

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -12,88 +12,91 @@
 ::      presence when we sub to someplace, and remove that when we unsub, but
 ::      don't make the target source do that.
 ::
-/-    hall                                              ::<  structures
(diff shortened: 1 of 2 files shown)


Fang-:
This is great, thanks so much for going over all my temporary-style comments!

I'm kind of sad to see the column-57 comments go, but it also reads more like an actual document this way.

ohAitch:
Imo they're incredibly helpful as context cues for where all the ++'s are,
but I've noticed that in general my spacing intuitions don't fit super well
with ++deco styling

eglaysher:
I don't particularly mind the column-57 comments on the gates. However, this doesn't compile:

```
++  gate         :<  short comment
  :>  long comment
  42
```

ohAitch:
Wh?

*looks at the parsers*

Issue #479, opened 2017-12-10
Status & config notifications getting rendered twice.


Fang-:
As evidenced by urbit-meta, most notably by presence notifications, federated circles seem to be sending changes twice. Curiously, subscribers are also processing (or at least relaying to clients) both these changes, despite the fix made in #475.

galenwp:
This seems relatively high priority.

Fang-:
Have had reports of people who *aren't* seeing double presence notifications, so my guess is there's something funky with sourcing going on. Definitely investigating further, but this luckily isn't "it's bouncing messages into infinity" levels of bad.

Hope to have this figured out sometime tonight.

Fang-:
I just received double config notifications when I invited someone to a village on my ship. I suspect the culprit is #469's change to talk's prep having resulted in double subscriptions for people where the hall-talk subscription was never dropped in the first place. Will have to research more to confirm, but that's fairly benign and might be an easy fix.

PR #488, opened 2017-12-11
Remove old hall susbcriptions in talk's ++prep

Fang-:
This really should've been included in #487. It's not as necessary here, but still good to get it back to baseline.


Fang-:
Included Yet Another Change to `++prep`, to resolve #479 for ships that are experiencing it. These are the ships that were working fine when #469 went live, which resulted in them having duplicate connections between their hall and talk.

Though it really doesn't matter in this case, we'll want to do another `++prep` cleanup after this goes live.

PR #489, opened 2017-12-11
Don't erase types in (lake %noun), which should always suceed

ohAitch:


belisarius222:
This could use a comment. Maybe something like:
```
Don't verify the product type when using the %noun mark, since that would cast to *,
which would prevent the product from being used as a gate or other typed structure.
```

belisarius222:
@ohAitch Travis build failed

ohAitch:
I noticed, with the comment it gets a SIGTERM while trying to pill (??)

belisarius222:
The joys of CI ... my first guess would be a timeout got triggered.

ohAitch:
After less than a minute?

belisarius222:
Oh ... probably not

ohAitch:
Restarted it and it built fine, idk what exactly was going on there

PR #492, opened 2017-12-12
Customizable filter for incoming hall messages

Fang-:
Creating `/app/hall/filter.hoon` allows you to modify messages before storing & resending them.

Intended to be used by stream for audience fixing. If we don't want to manually define the web ship's name in there, we may want to modify this to accept additional arguments. Maybe just the entire `bowl`. If that's desired, let me know and I'll put that in before getting this merged.

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -11,6 +11,11 @@
 /-    hall                                              ::  structures
 /+    hall, hall-legacy                                 ::  libraries
 /=    seed  /~  !>(.)
+/=    filter-gram
+      /^  $-(telegram:hall telegram:hall)
+      /|  /:  /%/filter  /!noun/
+          /~  |=(t/telegram:hall t)
+      ==
 ::
 ::::
   ::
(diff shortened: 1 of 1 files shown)


ohAitch:
Oh yeah, passing in the bowl or at least our seems reasonable, I think the
general principle is "we can copy all the files to a new ship and it'll
work"

(Like if the fora app starting and %home being federated were
straightforward to do/request in clay, that would also be helpful to check
in)

Fang-:
Starting fora is just a `%drum-start %base %fora` poke to drum I'm assuming, and creating & configuring %home is like two pokes to hall. Not sure where the right place to do those is though. Ideally we want to tack something on (as with the filter) rather than modifying existing files.

(You'd also still have to do `;source %urbit-meta ~web-ship/home` on ~marzod...)

ohAitch:
Yeah, I think ideally you have a like |tlon-init-urbit-org and a
corresponding |tlon-accept-new-stream-ship ~foo on marzod; I'm probably
overthinking things tho

(Other config: |serve /===/web/home, which could just be hardcoded
differently in eyre I guess)

Fang-:
Having at least the `|tlon-init-urbit-org` would be a good idea imo. But I think that would be implemented as yet another thing in /app/hood, because you can't have a generator send an arbitrary number of pokes to an arbitrary set of applications. That'd bring this back to another source modification for the web ship... Am I missing something?

ohAitch:
stick it in hood-helm or something, yeah. I'd just push that to master, the
urbit.org setup is not exactly confidential, and the tlon- prefix should
mark it clearly as an internal tool

PR #494, opened 2017-12-12
Add generators for easier webship ops

Fang-:
As discussed in #492.

On the web ship, run `|tlon-init-web ~marzod`. This will start the fora app, create a stream channel, and starts pulling messages from ~marzod's urbit-meta. (It loads a day's worth of messages, so that stream is immediately populated. Loading in the entire backlog is probably not necessary.)

On the linked urbit-meta's ship (here, ~marzod), run `|tlon-add-stream ~web-ship`. This adds the stream channel from the web ship as a source to urbit-meta, so stream posts made from that points forward get added to urbit-meta properly.

To ensure stream audience gets converted to urbit-meta, make sure urbit/arvo-ops-branches#1 is included on the web ship.


ohAitch:
Note to change [urbit/urbit.org](//github.com/urbit/urbit.org) main/stream.md to say station="stream" once it is no longer disabled.

PR #510, opened 2017-12-18
Enables more doccords in hall and talk.

eglaysher:
This isn't perfect[1], but it mostly enables doccords in hall and talk code. I haven't touched `%/lib/hall-json.hoon`, as the comments there are really minimal.

[1]: note the two arms in `%/app/talk.hoon` which aren't gates where I added todos about how adding doccords was leading to nest-fails.


Fang-:
Good stuff again, thanks so much for updating these!

PR #534, opened 2018-01-05
New key for ~tyr

mrdomino:
Also put on an n in compliance with current Steven Dee naming
conventions (and various government IDs.)


mrdomino:
@galenwp @cgyarvin?

PR #535, opened 2018-01-05
Split Tlon ops generators

Fang-:
@keatondunsford 

On the fora host, run `|start %fora`.  
On an urbit-meta host, run `|tlon-add-fora ~fora-host`.

(For completeness, this is preceded by the following steps, which have already been executed for the current setup using older version generators:  
On the stream host, run `|tlon-init-stream ~urbit-meta-host`.  
On an urbit-meta host, run `|tlon-add-stream ~stream-host`.)


keatondunsford:
@Fang- Amazing. Thanks!

Issue #501, opened 2017-12-14
%ames reload on ~zod didn't sync to network


belisarius222:
Trying a |reset now to force the issue, but this feels broken. Not sure what went wrong here.

ohAitch:
Talk seems to be crashing in response to an update hall is sending in
response to the merge(?), thereby preventing the merge from occurring.
@fang- help

belisarius222:
Also, it should be noted that performing a |reset on ~zod will not actually cause anything to sync to other ships. Only Clay changes will cause that to happen, and |reset doesn't modify Clay.

Fang-:
If talk's just printing a stacktrace pointing to [line 461](https://github.com/urbit/arvo/blob/master/app/talk.hoon?utf8=%E2%9C%93#L461), then that is a known issue. Shouldn't actually break anything though.

Didn't see this stack trace in ~zod's scrollback, and its talk seems to be working fine (it can send messages to its own inbox and have those show up), so I'm not sure what's causing the sync to not go through. Will investigate further.

Fang-:
It seems ~marzod and ~samzod got the update. (I can type `;sources %something` into their talks.)  
`|merge`ing their %kids desk into an old comet applies the update just fine. I'm assuming that their planets got the update as well.

~wanzod and ~binzod have *not* gotten the update, and are both showing the mentioned stacktrace along with the following:

```
  [%swim-take-vane %g %unto ~]
  [%gall-take /~wanzod/use/talk/~wanzod/out/hall/server/inbox]
[%fail-ack 0]
crude: all delivery failed!
  [%swim-take-vane %g %unto ~]
  [%gall-take /~wanzod/use/talk/~wanzod/out/hall/server/inbox]
crude: punted
```

I'll get to work on finding a reproduction case and fixing that talk error.

Fang-:
Breaking syncs in four easy steps:

1. Boot up a fakezod and accompanying star. Latest arvo is fine.
2. On the star, send a message. Default audience (inbox) is fine.
3. On the star, run `:talk 'screw'` in dojo. This should print `'screwing things up...'`
4. On the fakezod, make the following change:

```
  [~ ..prep(+<+ u.old)]
```

in ++prep, on or around line 123, needs to become:

```
  :_  ..prep(+<+ u.old)
  :~  [0 %pull /server/client server ~]
      [0 %pull /server/inbox server ~]
      peer-client
      peer-inbox
  ==
```

[...]

PR #538, opened 2018-01-08
Prevent gall from breaking things over bone 0

Fang-:
Maybe not ready for merging. Want to get a PR up for further discussion.

Apps that send `%peer` moves with bone 0 can get in trouble for that. When gall tries to kill the subscription (because some error occurred because of it) it tries to get at bone 0, but by its own rules, this isn't allowed.  
Considering bone 0 means "no cause", that's fair. Everything an app does has a cause. At the very least, that is "app was booted".

This change rejects moves that use bone 0, giving the same error message that is used for unknown bones. (And it technically is, because 0 only ever has empty ducts associated with it for some reason.)

An alternative approach saw [this line](https://github.com/urbit/arvo/blob/master/sys/vane/gall.hoon#L600) changed to set a 0 bone to be the bone that was used for booting the app, but that's silent and doesn't tell the developer bone 0 is semantically incorrect.

Prevents #501 and similar cases from happening. Note that this does nothing to reverse the damage that has been done.

### DIFF
--- a/sys/vane/gall.hoon
+++ b/sys/vane/gall.hoon
@@ -782,7 +782,7 @@
       ?^  -.q.vax  :_(+>.$ [%| (ap-suck "move: invalid move (bone)")])
       ?@  +.q.vax  :_(+>.$ [%| (ap-suck "move: invalid move (card)")])
       =+  hun=(~(get by r.zam) -.q.vax)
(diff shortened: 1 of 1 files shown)


Fang-:
RFC.

PR #539, opened 2018-01-09
Unit Testing / Partial New Library Work

eglaysher:
I think this is now in a state where we can now discuss merging this.

Functionally, this adds a generator `+test` which will depends on all tests under `%/tests` using ford. When the generator is invoked, it will run all tests whose paths start with an optional path takes as an argument to `+test`, running all tests on the current desk if no such filter is passed.

This also puts my container library in `%/lib/new-hoon.hoon`, which probably needs a better name. All the tests I wrote have become the sample tests for `+test`.


eglaysher:
@cgyarvin I added some documentation to this patch, describing how to use the system.

PR #540, opened 2018-01-11
Fix clay %next to provide current data

ohAitch:


sys/vane/clay.hoon
```diff
@@ -3456,34 +3447,40 @@
   ^-  (unit @da)
   ~
 ::
-++  load
+++  load  !:
```
belisarius222: looks like accidental !: left in

PR #541, opened 2018-01-12
adds AddressSanitizer to CI build

ohAitch:
tests urbit/urbit#915 on an actual fakezod

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,11 @@ node_js:
   - 4
 before_install:
   - cd .travis # keep main directory clear
-  
-  - wget -i pin-urbit-release.url -O ./urbit.deb
-  - sudo apt install ./urbit.deb
-  
+  - >
+      wget $(cat ./pin-urbit-release.url)/urbit -O ./urbit &&
+      chmod +x ./urbit;
+      echo "FIXME downloaded raw urbit binary releaseinstead of .deb";
+      echo "FIXME used full pinned url instead of tag name"
       
 before_script: bash get-or-build-pill.sh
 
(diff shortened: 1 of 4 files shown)


joemfb:
Setting ASAN_OPTIONS (which is read at runtime) to ignore the shadow gap (see 0173c86df) should get you past this error.

joemfb:
Update: that option is only successful if the binary was built with clang -- I still can't get it to work with gcc. I've updated urbit/urbit#915

belisarius222:
This broke our CI, so let's close this until we have time to get back to work on fixing that.

ohAitch:
s|broke our CI|broke in CI| :P

joemfb:
As far as I can tell, the only thing that works is building vere with clang 5, and then setting some runtime options. Getting that to work will require some more back-and-forth between the urbit/urbit and urbit/arvo CI configurations. Idk how to do that without a) being a major hassle for you or @ohAitch or b) setting up my own copies of the CI infrastructure.

ohAitch:
How far can you get on b) in 15 minutes? It should be a matter of
- auth'ing in to http://travis-ci.org
- enabling it for your fork of the urbit and arvo repos
- gem install travis to grab their cli tool
- running  travis setup releases -f  to swap the .travis.yml  deploy:
section to point at joemfb/urbit, and pushing a tag with that change
- if this successfully dumps a binary in
https://github.com/joemfb/urbit/releases, fork off this PR with
https://github.com/*joemfb*/urbit/releases/download/*tagname*
in.travis/pin-urbit-release.url,
and push that to a branch

PR #575, opened 2018-01-15
Clay multi-%next: %mult (and fix for %next)

Fang-:
This is the simplest possible implementation of a %next for multiple simultaneous files.  
If the subscription includes files in the past, or on other ships, then it is ended immediately. If one or more of the watched files change, then all those changes are sent in one response.

Not yet ready for merge. Lobes currently [aren't dealt with at all](https://github.com/Fang-/arvo/blob/afb1836f4477c8f35d5bc862efa3a1926354a8b3/sys/vane/clay.hoon#L604), as I'm not entirely sure how to do that here. It appears as if `++blab` calls out to ford to build the file into data for us, which won't happen for our restricted set of supported cases. Correct y/n?

(Also I should maybe change existing uses of `(unit (each cage lobe))` into `cach` (and probably rename that into like re`sult` or something), for brevity.)

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -270,9 +270,15 @@
 ::  Like a ++rave but with caches of current versions for %next and %many.
 ::  Generally used when we store a request in our state somewhere.
 ::
+++  cach  (unit (each cage lobe))                       ::  cached result
 ++  rove                                                ::  stored request
           $%  {$sing p/mood}                            ::  single request
               {$next p/mood q/(unit (each cage lobe))}  ::  next version
+              $:  $mult                                 ::  next version of any
+                  p/care                                ::
+                  q/(jug case spur)                     ::
+                  r/(map case (map spur cach))          ::
+              ==                                        ::
               {$many p/? q/moat r/(map path lobe)}      ::  change range
           ==                                            ::
 ::
@@ -318,6 +324,7 @@
           {$note p/@tD q/tank}                          ::  debug message
           {$ogre p/@tas}                                ::  delete mount point
           {$writ p/riot}                                ::  response
+          {$wris p/(set rant)}                          ::  responses
       ==                                                ::
 ++  note                                                ::  out request $->
   $%  $:  $a                                            ::  to %ames
@@ -587,6 +594,16 @@
         %f  %exec  our  ~  [her syd q.mun]  (lobe-to-silk:ze r.mun p.dat)
     ==
   ::
+  ++  blas
+    |=  {hen/duct das/(map mood (each cage lobe))}
+    ^+  +>
+    =-  (emit hen %give %wris -)
+    %-  ~(run in das)
+    |=  {mun/mood dat/(each cage lobe)}
+    ~|  %lobe-unsupported-tmp
+    ?>  ?=($& -.dat)
+    [[p.mun q.mun syd] r.mun p.dat]
+  ::
   ::  Give next step in a subscription.
   ::
   ++  bleb
@@ -621,6 +638,7 @@
   ::
   ++  blub-all  (duct-lift |=({a/duct $~} (blub a)))    ::  lifted ++blub
   ++  blab-all  (duct-lift blab)                        ::  lifted ++blab
+  ++  blas-all  (duct-lift blas)                        ::  lifted ++blas
   ++  balk-all  (duct-lift balk)                        ::  lifted ++balk
   ++  bleb-all  (duct-lift bleb)                        ::  lifted ++bleb
   ::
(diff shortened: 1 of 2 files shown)


Fang-:
Appending a fix for clay immediately closing a %next on a not-yet-existing file to this PR because it also involves changing the state mold (through `++rove`) and is tangentially related anyway.

sys/vane/clay.hoon
```diff
@@ -788,25 +842,81 @@
...
+        q:(snag 0 `(list (pair path (unit cach)))`ole)
+      ::
+      ++  respond                                       ::  send changes
+        |=  res/(map mood (each cage lobe))
+        ^+  ..start-request
+        ?:  ?=($mult -.rav)  (blas hen res)
+        ?>  ?=({* $~ $~} res)
+        (blab hen n.res)
+      ::
+      ++  know  |=({p/path c/(unit cach)} ?=(^ c))      ::  know awout file
```
belisarius222: awout -> about

sys/vane/clay.hoon
```diff
@@ -788,25 +842,81 @@
...
+      ::
+      ++  read-all-at                                   ::  initialize cache
```
belisarius222: Isn't this more general than just for initializing a cache? It's really more like:
`try to read all paths at case, producing a list of (possibly blocked) results`, or something like that.

sys/vane/clay.hoon
```diff
@@ -1673,26 +1787,105 @@
...
+          (store p.res)
+      |-  ::  so that we can retry for the next aeon if possible/needed.
+      ::  if we don't have an aeon yet, see if we have one now.
+      ?~  yon
+        =+  aey=(case-to-aeon:ze q.mol)
+        ::  if we still don't, wait.
+        ?~  aey  |+rov
+        ::  if we do, update the request and retry.
+        $(rov [-.rov mol `+(u.aey) ~ ~])
+      ::  if old isn't complete, try fillin in the gaps.
```
belisarius222: fillin -> filling

sys/vane/clay.hoon
```diff
@@ -1673,26 +1787,105 @@
...
+      ++  respond                                       ::  send changes
+        |=  res/(map mood (each cage lobe))
+        ^+  ..wake
+        ::NOTE  want to use =-, but compiler bug?
+        ?:  ?=($mult -.vor)
+          ^^$(xiq t.xiq, ..wake (blas-all q.i.xiq res))
+        ?>  ?=({* $~ $~} res)
+        ^^$(xiq t.xiq, ..wake (blab-all q.i.xiq n.res))
+      ::
+      ++  know  |=({p/path c/(unit cach)} ?=(^ c))      ::  know awout file
```
belisarius222: awout -> about

Fang-:
Okay, apparently some of the efforts herein were a bit misguided.

Really the only thing that currently uses `%next` is ford. And while `%next` sends a `cage` containing the new version of the changed thing in its response, that isn't even used.

`%next` should only send the case and the paths to the things that changed at that case in its response, and `%next` really should just support multiple paths instead of requiring a different request (`%mult`) for that.

Fang-:
Have tested and confirmed that `%mult` works  as expected, even when asking about cases in the past.

I have left `%next` in without changing its interface for temporary backwards compatibility. Once ford has been updated to use `%mult`, we should rename it to `%next` and completely get rid of old, one-path-and-full-change `%next`.  
(Or, if you ford devs don't care, I can go and change that right now.)

sys/vane/clay.hoon
```diff
@@ -318,6 +325,7 @@
           {$note p/@tD q/tank}                          ::  debug message
           {$ogre p/@tas}                                ::  delete mount point
           {$writ p/riot}                                ::  response
+          {$wris p/(set rant)}                          ::  responses
       ==                                                ::
```
ohAitch: Is there a reason this arm exists separately from zuse? `++move` directly above uses the zuse version instead, so it's not like this copy of the definition is ever referenced. (In particular, the `%wris` definition is outdated and would not otherwise compile)

belisarius222:
More specifically, (set rant) should be (set path)

sys/vane/clay.hoon
Fang-: Hmm yeah, I missed this for the reason you mentioned. Originally updates it because "hey, this is here, might as well keep it consistent", but you're right that there's no reason for it to be there in the first place. Can confirm clay compiles just fine without it.

Similarly, replacing `++note`'s body with `note-arvo` (and making sure to say `%b` where we currently say `%t`) works fine too. Probably best to not re-define these interfaces locally, I imagine?

ohAitch:
Note makes sense to me, since it's defining a subset of the API that clay
is actually using(and mirrors ++sign, which we definitely don't want to
open up); it should also be typechecked against zuse however.

ohAitch:
Is there an internal clay reason the api is `[care case (set path)]`, and not `[case (set [care path])]`? The ford keys are all `[beam care]`, and while currently it's always %z anyway, at some point we want to revert to using canonical `%x`/`%y`. In that case, if you have e.g. a tree page serving `docs/*`, deleting `docs/intro.md` would produce notifications on all of `%y docs/`, `%y docs/into`, `%y docs/intro/md`, and `%x docs/intro/md`.

Ideally, those would all be in the same event, so that ford could have a full blacklist and treat all other requests to that desk as unchanged.

PR #578, opened 2018-01-16
Fixed weird code for updating /circles subscriptions.

Fang-:
No longer incorrectly uses the subscriber's ship name when getting the affected story.

(The real fix is the change from `who.qer` to `nom.det` on line 2328. The rest is just cleanup and comments to make this block easier to follow, since it had me struggling for a little while when revisiting it just now.)

cc @c-johnson

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -2323,17 +2323,25 @@
     ::REVIEW  this could be considered leaky, since it
     ::        doesn't check if {who} ever knew of {nom},
     ::        but does that matter? can't really check..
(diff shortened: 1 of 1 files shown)


belisarius222:
I struggled to follow this code too. It looks like it should work, though. @Fang- merge at your leisure.

Fang-:
I'll push this live alongside a fix for urbit/talk#46, which I should get to soon.

c-johnson:
thank you!

PR #597, opened 2018-01-23
Clay %mult multiple cares

Fang-:
As suggested by @ohAitch in #575, the interface for `%mult` now allows you to set a care *per path* rather than for all the paths in the request.

Also removes needless `++gift` mirror from its internal structures.


Fang-:
Should note that this assumes previous changes haven't gone onto the livenet yet. If they have, this'll  need a state adapter because of the changes to `++rove`.

belisarius222:
I did not push code to the network recently. Unless @keatondunsford did, I think we're fine.

PR #610, opened 2018-02-08
Clay permissions

Fang-:
As discussed internally, permissions for the filesystem.

Permissions are set for all revisions of a node. If a node doesn't have permissions set, those of the parent node are used. If no permissions are set all the way up to root, we default to fully private.

The clay interface sees some additions to support basic permissions management.

`++task`:
- `{$cred our/ship nom/@ta cew/crew}` for setting permission groups.
- `{$crew our/ship}` for getting permission groups.
- `{$perm our/ship des/desk pax/path rit/rite}` for setting permissions for a node.
- A `%p` `++care` for reading permissions set for a node on a local desk.

`++gift`:
- `{$cruz cez/(map @ta crew)}`, the response to `%crew` requests.
- `{$rule red/dict wit/dict}`, the response to `%p` file requests.

~And a notable change that was required for making permission checks actually functional:  
The interface for `%warp` file request now includes a `ship` which has to be set to the requesting ship.  
When doing a `%warp` for local data, you just set it to `our`, as you do for many other tasks. When doing a `%west` for foreign data, the target ship automatically turns it into a `%warp` with the original sender specified.~  
Read requests from foreign ships come in as `%west`s. They now get re-sent to the local clay as `%werp`s, which get dealt with as `%warp`s that require permission checks. No checks are done for write requests, because those only originate locally right now.

Eventually, the above will be more elegantly solved from within arvo itself, passing vanes a pair of recipients and responsible ships for each incoming event. Not to mention making a global `our` available once the new boot sequence lands. Changing this to make use of that shouldn't be difficult.

Aside, having to take original requester into account when deduping requests likely completely obliterates any use that still saw. We're fine to keep it in, but at this point it might be costing more performance than it'll ever gain us.

Need to do more thorough testing and implement a state adapter, but it boots (off an updated pill at least) and seems to work as advertised so far, so I'd say that's good enough for review/discussion.

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -3536,7 +3536,7 @@
 ::
 ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
 =|                                                    ::  instrument state
-    $:  $3                                            ::  vane version
+    $:  $4                                            ::  vane version
         ruf/raft                                      ::  revision tree
     ==                                                ::
 |=  {now/@da eny/@ ski/sley}                          ::  activate
@@ -3767,44 +3767,55 @@
 ::
 ++  load
   =>  |%
-      +=  rove-2
-        $%  {$sing p/mood}
-            {$next p/mood q/(unit (each cage lobe))}
-            {$many p/? q/moat r/(map path lobe)}
+      +=  wove-3  rove
+      ++  cult-3  (jug wove-3 duct)
+      ++  dojo-3
+        $:  qyx/cult-3
+            dom/dome
+            dok/(unit dork)
+            mer/(unit mery)
         ==
-      ++  cult-2  (jug rove-2 duct)
-      ++  dojo-2  (cork dojo |=(a/dojo a(qyx *cult-2)))
-      ++  rede-2  (cork rede |=(a/rede a(qyx *cult-2)))
-      ++  room-2  (cork room |=(a/room a(dos (~(run by dos.a) dojo-2))))
-      ++  rung-2  (cork rung |=(a/rung a(rus (~(run by rus.a) rede-2))))
-      ++  raft-2
-        %+  cork  raft
-        |=(a/raft a(fat (~(run by fat.a) room-2), hoy (~(run by hoy.a) rung-2)))
-      ++  axle    $%({$2 ruf/raft-2} {$3 ruf/raft})
+      ++  rede-3
+        $:  lim/@da
+            ref/(unit rind)
+            qyx/cult-3
+            dom/dome
+            dok/(unit dork)
+            mer/(unit mery)
+        ==
+      ++  room-3  (cork room |=(a/room a(dos (~(run by dos.a) dojo-3))))
+      ++  rung-3  (cork rung |=(a/rung a(rus (~(run by rus.a) rede-3))))
+      ++  raft-3
+        $:  fat/(map ship room-3)
+            hoy/(map ship rung-3)
+            ran/rang
+            mon/(map term beam)
+            hez/(unit duct)
+        ==
+      ++  axle    $%({$3 ruf/raft-3} {$4 ruf/raft})
       --
   |=  old/axle
   ^+  ..^$
   ?-  -.old
-    $3  ..^$(ruf ruf.old)
-    $2  =/  rov
-          |=  a/rove-2  ^-  rove
-          ?+  -.a  a
-              $next
-            ?~  q.a  a
-            a(q `q.a)
-          ==
+    $4  ..^$(ruf ruf.old)
+    $3  =/  wov
+          |=  a/wove-3  ^-  wove
+          [~ a]
         =/  cul
-          |=  a/cult-2  ^-  cult
+          |=  a/cult-3  ^-  cult
           %-  ~(gas by *cult)
-          (turn ~(tap by a) |=({p/rove-2 q/(set duct)} [(rov p) q]))
+          (turn ~(tap by a) |=({p/wove-3 q/(set duct)} [(wov p) q]))
         =/  rom
-          =+  doj=|=(a/dojo-2 a(qyx (cul qyx.a)))
-          |=(a/room-2 a(dos (~(run by dos.a) doj)))
+          =+  doj=|=(dojo-3 [(cul qyx) dom dok mer ~ ~])
+          |=(a/room-3 a(dos (~(run by dos.a) doj)))
         =/  run
-          =+  red=|=(a/rede-2 a(qyx (cul qyx.a)))
-          |=(a/rung-2 a(rus (~(run by rus.a) red)))
-        =+  r=ruf.old
-        $(old [%3 r(fat (~(run by fat.r) rom), hoy (~(run by hoy.r) run))])
+          =/  red
+            |=  rede-3
+            =+  [[/ %black ~] ~ ~]
+            [lim ref (cul qyx) dom dok mer - -]
+          |=(a/rung-3 a(rus (~(run by rus.a) red)))
+        =+  ruf.old
+        $(old [%4 [(~(run by fat) rom) (~(run by hoy) run) ran mon hez ~]])
   ==
 ::
 ++  scry                                              ::  inspect
@@ -3836,7 +3847,7 @@
   ?:  ?=($& -.u.u.-)  ``p.u.u.-
   ~
 ::
-++  stay  [%3 ruf]
+++  stay  [%4 ruf]
 ++  take                                              ::  accept response
   |=  {tea/wire hen/duct hin/(hypo sign)}
   ^+  [p=*(list move) q=..^$]
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -329,7 +329,8 @@
       $:  $c                                            ::  to %clay
   $%  {$info p/@p q/@tas r/nori}                        ::  internal edit
       {$merg p/@p q/@tas r/@p s/@tas t/case u/germ}     ::  merge desks
-      {$warp p/ship q/sock r/riff}                      ::
+      {$warp p/sock q/riff}                             ::
+      {$werp p/ship q/sock r/riff}                      ::
   ==  ==                                                ::
       $:  $d                                            ::
   $%  {$flog p/{$crud p/@tas q/(list tank)}}            ::  to %dill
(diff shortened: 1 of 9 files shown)


Fang-:
Changing the `%warp` interface was giving problems, new-style `%warp` events were being sent before clay itself got properly updates to handle them.  
Settled on adding `%werp` for incoming external read requests and leaving `%warp` unchanged. This makes upgrading easier, doesn't break compatibility, and keeps the internal file request interface clean.

sys/vane/clay.hoon
```diff
@@ -2314,6 +2360,66 @@
...
+          $p
+        =(who our)
+      ::
+          ?($y $z)
+        =+  tak=(~(get by hit.dom) yon)
+        ?~  tak  |
+        =+  yak=(tako-to-yaki u.tak)
+        =+  len=(lent pax)
+        =-  (levy ~(tap in -) |=(p/path (allowed-by who p per.red)))
+        %+  roll  ~(tap in (~(del in ~(key by q.yak)) pax))
```
belisarius222: This seems likely to be a duplicate of the usual %y / %z logic. If so it might be good to make a little helper function.

sys/vane/clay.hoon
```diff
@@ -2314,6 +2360,66 @@
...
+      |=  {who/ship pax/path pes/(map path rule)}
+      ^-  ?
+      =+  rul=rul:(read-p-in pax pes)
+      =-  =(0 -)
+      %+  mix  ?=($black mod.rul)
+      %-  ~(rep in who.rul)
+      |=  {w/whom h/?}
+      ?:  h  &
+      ?:  ?=($& -.w)  =(w &+who)
+      (~(has in (fall (~(get by cez) p.w) ~)) who)
```
belisarius222: Could `(~(get by cez) p.w)` ever fail? I feel like that should be a `~(got by ...)`

sys/vane/clay.hoon
```diff
@@ -3457,6 +3567,23 @@
       $boat
     :_  ..^$
     [hen %give %hill (turn ~(tap by mon.ruf) head)]~
+  ::.
+      $cred
+    =.  cez.ruf
+      ?~  cew.q.hic  (~(del by cez.ruf) nom.q.hic)
```
belisarius222: `q.hic` could really use an alias. Comparing `p.q.hic` and `p.q.q.hic` a little later in this function is cruel and unusual punishment.

sys/vane/clay.hoon
```diff
@@ -3576,19 +3703,33 @@
...
+      =+  den=((de now hen ruf) [. .]:our.q.hic des.q.hic)
+      abet:(perm:den pax.q.hic rit.q.hic)
+    [mos ..^$]
+  ::
+      ?($warp $werp)
+    =^  for  q.hic
+      ?:  ?=($warp -.q.hic)
+        [~ q.hic]
+      :_  [%warp q.q.hic r.q.hic]
+      ?:  =(p.q.hic p.q.q.hic)  ~&([%huh-this-west-may-be-weird p.q.hic] ~)
```
belisarius222: What's weird about it?

sys/vane/clay.hoon
```diff
@@ -3576,19 +3703,33 @@
...
+      `p.q.hic
+    ?>  ?=($warp -.q.hic)
     =^  mos  ruf
       =+  den=((de now hen ruf) p.q.hic p.q.q.hic)
-      ::  =-  ~?  ?=([~ %sing %w *] q.q.q.hic)
+      ::  =-  ~?  ?=([~ %sing %w *] q.r.q.hic)
       ::        :*  %someones-warping
-      ::            rav=u.q.q.q.hic
+      ::            rav=u.q.r.q.hic
```
belisarius222: There's no way I can reasonably review `u.q.r.q.hic`. I realize this was already the style in this part of the code, but it's not fit for human consumption.

sys/vane/clay.hoon
```diff
@@ -3635,44 +3776,60 @@
...
-            ?~  q.a  a
-            a(q `q.a)
-          ==
+    $4  ..^$(ruf ruf.old)
+    $3  =/  wov
+          |=  a/wove-3  ^-  wove
+          [~ a]
         =/  cul
-          |=  a/cult-2  ^-  cult
+          |=  a/cult-3  ^-  cult
```
belisarius222: I think this would be cleaner as a `|^` since it has so many internal functions.

sys/vane/dill.hoon
```diff
@@ -280,7 +281,7 @@
                   (sync %home our %base)
                 (init-sync %home our %base)
         =.  +>  ?.  ?=(?($duke $king $czar) can)  +>
-                (sync %kids our %base)
+                (show %kids):(sync %kids our %base)
```
belisarius222: Can you explain why this change was necessary?

sys/vane/clay.hoon
```diff
@@ -367,6 +375,7 @@
 ::
 ::  --  current time `now`
 ::  --  current duct `hen`
+::  --  foreign requester `for`, if any
```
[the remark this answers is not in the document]
Fang-: Lingering trace of experimentation. Will delete.

sys/vane/clay.hoon
```diff
@@ -2193,7 +2238,7 @@
     ::  Gets a map of the data at the given path and all children of it.
     ::
     ++  lobes-at-path
-      |=  {yon/aeon pax/path}
+      |=  {for/(unit ship) yon/aeon pax/path}
```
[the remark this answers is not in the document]
Fang-: It should be used, since this is how `%many` requests get their data. Going to add a permission check here.

sys/vane/clay.hoon
Fang-: It totally can. Either I added a group which I haven't defined yet, or I deleted a group and there's still references to it lingering.

sys/vane/clay.hoon
Fang-: Nothing but the fact that I left a debug statement in it. Cleaning up.

sys/vane/clay.hoon
Fang-: This particular comment can probably just get removed, but I agree that I should probably add a few `=*` here.

sys/vane/dill.hoon
Fang-: Since all local desks are fully private by default, we need to make our `%kids` desk public explicitly if we want people to be able to sync with it.

sys/vane/clay.hoon
```diff
@@ -421,6 +430,8 @@
             dom=*dome
             dok=~
             mer=~
+            per=[[/ %black ~] ~ ~]
```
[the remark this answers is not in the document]
Fang-: These are the defaults for local copies of foreign desks. Those are not accessible by outsiders over the network. In a previous iteration of the PR setting these to be readable was required.
That said, the current version always says "no permission checks" for all internal requests, so there's no need to set permissions here explicitly anymore.

sys/vane/clay.hoon
Fang-: Fair, on it!

sys/vane/clay.hoon
Fang-: You're right, even just aliasing `q.hic` to `req` makes a world of difference.
I also made a commit that adds actual faces to the `++task`. Not sure if that's boyscouting a bit too aggressively.

sys/vane/clay.hoon
Fang-: It's not an exact match. `%y` wants just the path, `%z` wants lobes too, and they both want only the path tails, where the permission check just wants the full paths.  
The logic it still fairly transferable though, the data structures here are easy enough to work with. Having one function for "get child nodes with tail paths" seems to give a net gain to simplicity.

Fang-: Actually, I'm having trouble finding a helper function here that isn't in some way terrible to use for two out of three cases. Problem is that if we do a dir listing for `/one` with the following data stored, then we still want to see `./two` show up, but we can't provide a `lobe` alongside that, unlike for `./three`

```
/one/two/x
/one/two/y
/one/three
```

belisarius222: Ok, no need to force it if it doesn't feel right.

sys/vane/dill.hoon
belisarius222: I think a comment here would be helpful.

sys/vane/clay.hoon
```diff
@@ -3621,6 +3621,27 @@
...
+      ?~  des  [[hen %give %croz rus]~ ..^^$]
+      =+  per=(filter-rules per.q.i.des)
+      =+  pew=(filter-rules pew.q.i.des)
+      $(des t.des, rus (~(put by rus) p.i.des per pew))
+    ::
+    ++  filter-rules
+      |=  pes/regs
+      ^+  pes
+      =-  (~(gas in *regs) -)
+      %+  murn  ~(tap by pes)
```
belisarius222: Looks like this could be a ++skip instead of a ++murn.

belisarius222:
Looks good to me. How's the testing coming?

Fang-:
As you might deduce from the commits, testing pointed to a few issues that needed solving. Everything now works as expected.  
Both upgrading from livenet, booting using old pill and booting using new pill all work fine and don't break syncs.

For the record, and for the future, [here](https://gist.github.com/Fang-/a585bd75eabda765d49c5bfa8baadd59) are the beginnings of a clay testing lib. Should work nearly identically for other vanes. Also see [this](https://github.com/ohAitch/arvo/blob/21db0da/lib/test/jael.hoon) bigger one for jael.  
Didn't do much actual testing with this because the testing framework wasn't playing well with it and I just wanted to get this PR finished.

Included is a pair of generators for setting the readability of desks (or "directories" therein). `|public` and `|private %desk /optional/path`.

Unless there's anything other than the `%kids` desk that needs to remain public (`web.plan` maybe?), I'll call this fine to be merged. @belisarius222, want to take a final look at this?

belisarius222:
Re-reviewed. Looks good to me.

Fang-:
I lied about the booting btw (thanks, last-minute additions), kiln *does* complain about not finding one of the new clay types if you don't boot with a new-clay pill. So we'll want to make sure to push an up-to-date latest.pill when this goes onto the network.

belisarius222:
Does that mean this needs to be breached in?

Fang-:
It doesn't need to be breached in, but I just double-checked, and the kiln changes did break the updating. Particularly, it says it can't find `++rite`, a new clay type.

We could do like a partial update that doesn't include the offending kiln changes, and then push out the second half separately.  
I think the slightly better solution is to do [what we did](https://github.com/urbit/arvo/blob/master/lib/hood/kiln.hoon#L83) when we saw similar behavior with `++riot` (also used by kiln), and just include the relevant structures in kiln directly. We can then (after this has been pushed to live) remove the structure definitions from kiln for cleanliness. That won't have to be pushed out immediately. Sound good?

(I guess ultimately this is an issue of kiln not getting recompiled alongside a zuse change...)

PR #618, opened 2018-02-12
Replace $-(* *) with gate.

jfranklin9000:
If you want these changes I'll update the Source hoons in the library docs.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -387,7 +387,7 @@
   (b u.a)
 ::
 ++  bind                                                ::  argue
-  |*  {a/(unit) b/$-(* *)}
+  |*  {a/(unit) b/gate}
   ?~  a  ~
   [~ u=(b u.a)]
 ::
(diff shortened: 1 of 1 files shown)


sys/hoon.hoon
```diff
@@ -12651,16 +12651,16 @@
          ==                                            ::
 ++  desk  @tas                                          ::  ship desk case spur
 ++  cage  (cask vase)                                   ::  global metadata
-++  cask  |*(a/$-(* *) (pair mark a))                   ::  global data
```
ohAitch: this looks like it should obviously be `mold`, not `gate`? (similarly with `++pair` calls and such)

jfranklin9000:
And `++hypo` too.  I changed these to `gate` because they matched the existing "type signature".
I've tested changing `++cask` and `++hypo` to `mold` and all the vanes compile, but I'm no expert.
Someone will have to confirm that these should be changed to `mold`.

PR #627, opened 2018-02-19
Fix outdated use of "span" in hoon.hoon

Fang-:
Guess we should've waited for the CI check on #417 after all.


ohAitch:
I notice CI is still broken, and \~unrelatedly~, if I try to boot a latest.pill master fakezod it crashes `-find.rite` in kiln. What gives?

belisarius222:
It was from the merges we did on Friday. I have it on my list of todos to
check on the CI and revert any problematic changes before we deploy master
to the network. I can look into this more tomorrow.

Issue #611, opened 2018-02-09
Crash caused by sync causes sync to break permanently


Fang-:
Issue #501 was caused by this. See that for past discussion and an outdated repro case.  
That particular root cause was fixed in #538, but probably didn't change the fact that a crash caused by the effects of a sync seems to break the sync/it's subscription permanently.

Crashes shouldn't be breaking upstream subscriptions as aggressively as they are.

Maybe we should just separate the notification %clay sends after file change more strongly.

jtobin:
I was going to leave this one open, but remembered a great line from the 1998 film *Ronin*:

> If there's any doubt, there is no doubt.

I think it needs to be updated, or more detail needs to be provided, in order for it to be useful.

joemfb:
> Maybe we should just separate the notification %clay sends after file change more strongly.

This was just done in @philipcmonk's clay transaction rewrites.

Issue #509, opened 2017-12-18
Old version of application stays running after update


Fang-:
I (on ~palfun-foslup) and others (like ~tonlur-sarret) are running into an issue where we have received the latest version of talk, but don't seem to be running it. We can see changed/added application logic in the hoon file (in both base and home), but talk acts as if that weren't there.

We've tried editing the file manually, to make it re-sync. And it does, but still doesn't seem to affect the running code.

May or may not be related to #501.

belisarius222:
I have some questions about this:
1) Do these ships think they're mid-merge in Clay anywhere? Just want to check that this isn't some sort of downstream bug from the Clay bug we've been seeing.
2) Have you checked whether the app tried to reload? I wonder if its `++prep` arm errored out or something like that. We could try adding a printf whenever it reloads ... if we could get it to reload the code containing the printf ...
3) Has anyone seen this sort of behavior before? @cgyarvin @ohAitch

cgyarvin:
This seems like a %gall bug.  Can you construct a "tame" example case?  Or
even just save a pier?

Fang-:
Re @belisarius222:

1. `|syncs` on ~palfun shows that everything's set up properly, doesn't tell me anything's currently running.
2. Just tried, a printf in ++prep doesn't print...

I'll try and see if I can reproduce this on a fakenet. If not, we may just have plenty of ships to pick from...

joshuareagan:
Same thing is happening on `~taglux-nidsep`.

PR #632, opened 2018-02-21
Revert changes to hall circle configs (and thus the hall protocol)

Fang-:
The stars are currently drowning in `%diff-bad-ack` messages. They're trying to send out urbit-meta updates so subscribers, but not all of those subscribers are running the latest arvo with the recent hall changes. This results in a mismatch in the expected message type.

In its current state, the network clearly isn't ready to receive userspace protocol updates. Not all ships are getting/running updates (see also #509 and #611), and comets weren't in the first place. There is currently no mechanism for implementing application interface versioning that would allow this to be overcome elegantly, so I think our safest bet here is to just keep interfaces consistent for now.

The changes this reverts come from #621. These were made to support the currently still in development Collections application/web interface, so nothing on the livenet depends on them. After this PR, up-to-date ships will be able to chat with out-of-date ships again.

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -117,7 +117,7 @@
 :>  #
 :>    functional cores and arms.
 ::
-|_  {bol/bowl:gall $1 state}
+|_  {bol/bowl:gall $2 state}
 ::
 :>  #  %transition
 :>    prep transition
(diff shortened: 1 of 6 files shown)


Fang-:
Got the go-ahead by @belisarius222 to merge this and push it live.

PR #636, opened 2018-02-22
escapes "{" in tape print (++dash:us)

joemfb:
closes urbit/urbit#630

I've done the whole song-and-dance to incrementally update the `++dash:us` signature. As near as I can tell, this would have to a minimum of three separate syncs to be safely released over-the-air.


cgyarvin:
Ha I believe that!

belisarius222:
Well, we are breaching tomorrow ...

Fang-:
Works fine when including in a ship's boot directly. Into the breach it goes!

PR #637, opened 2018-02-22
Pre-breach cleanup

Fang-:
- Make federation more performant by just sending 100 messages of backlog initially. (short-term fix for #582 in a federation context)
- Put #621 back in.
- Clean up app (hall, talk) and vane (clay, dill, eyre, ford, gall) states and state adapters to be state version `%0` again.
- Remove deprecated code in kiln.

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -2192,9 +2192,9 @@
     ::  only auto-federate channels for now.
     ?.  ?=($channel sec.con.shape.s)  ~
     :+  ~  n
-    ::  share no more than 2k messages at once, for performance reasons.
-    :+  ?:  (lte count.s 2.000)  grams.s
-        (slag (sub count.s 2.000) grams.s)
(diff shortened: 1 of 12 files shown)


app/hall.hoon
```diff
@@ -2192,9 +2192,9 @@
     ::  only auto-federate channels for now.
     ?.  ?=($channel sec.con.shape.s)  ~
     :+  ~  n
-    ::  share no more than 2k messages at once, for performance reasons.
-    :+  ?:  (lte count.s 2.000)  grams.s
-        (slag (sub count.s 2.000) grams.s)
+    ::  share no more than the last 100, for performance reasons.
+    :+  ?:  (lte count.s 100)  grams.s
```
belisarius222: I feel like this should be a named constant, but this is better than it was so I approve.

belisarius222:
This all looks good to me. Please merge once tested.

PR #639, opened 2018-02-22
Bump ames version for breach

Fang-:
Also removes trailing whitespace from ames, something we really should do for the whole repo sometime.


cgyarvin:
The time to do mass clerical conversions like whitespace removal is when the codebase is quite stable — it wreaks hell with merges.

PR #642, opened 2018-02-22
fix typo

cxk280:


keatondunsford:
@cxk280 Thanks!

@ohAitch You bet.

PR #651, opened 2018-03-02
Add keccak hashing family into zuse

Fang-:
This implements the keccak family of hashing functions (keccak, sha3 (FIPS 202), shake and rawshake). These are required for my ongoing Ethereum integration, but are worth their own PR.

Invoke as such:

```
=,  keccak:crypto
=/  input-a=@  (swp 3 'some string')
=/  input-b=@  0xabcd.1234
(keccak-256 input-a)
(sha3-512 input-b)
(shake-128 256 input-a)
(rawshake-224 64 input-b)
::  outputs, in order, cast from @ to @ux:
::  0x83c7.37ad.570e.9f3e.71e0.d280.0958.e447.70d8.12e9.2db2.c175.8626.613d.1e6b.a514
::  0x1cae.3332.3396.c8d7.ddb8.cfbb.3d86.b66f.2273.fe62.0f8a.2d9f.869a.b6ca.cdae.a4cb.4669.b080.f3d4.bb80.71e0.4630.7d8a.ce9b.87ee.bd9a.1391.4d08.7d04.10ea.ed12.eae6
::  0x971d.f044.e825.bed0.dcbe.0650.342d.ebab.1732.5fe0.bd41.16a4.3b8b.388d.44cf.5e34
::  0x9dd2.45ee.bef0.e9a3
```

Shake has variable output length, so you specify that manually (in bits).  
Number atoms can be hashed as-is, but we have to reverse the byte order in strings because endianness.

For the record, these are the resources that helped me understand and implement the mechanics behind keccak. Byte ordering had me confused for longer than I care to admit.

1. [Keccak official reference document](https://keccak.team/files/Keccak-reference-3.0.pdf)
2. [Pseudocode summary of the spec](https://keccak.team/keccak_specs_summary.html)
3. [This textbook chapter](https://pdfs.semanticscholar.org/8450/06456ff132a406444fa85aa7b5636266a8d0.pdf)
4. [A Python implementation](https://github.com/ctz/keccak/blob/master/keccak.py)
5. [Padding and byte ordering explanations #1](https://www.reddit.com/r/crypto/comments/1w5xig/can_somebody_explain_the_bytebit_ordering_in/)
6. [Padding and byte ordering explanations #2](https://www.reddit.com/r/crypto/comments/6287my/explanations_on_the_keccaksha3_paddingbyte/)
7. [Easily finding expected output of arbitrary input strings](https://emn178.github.io/online-tools/keccak_256.html)

The code should contain enough comments to match lines of code against parts of the spec.

`++sponge` function isn't generic currently, it assumes a permutation state of 5x5 words. This is fine for keccak, but may not be appropriate for other algorithms that want a sponge construction. I consider that out of scope for now though, we can always touch it up later.

Before this gets merged, I still want to write a test file for these so you won't have to take my word for it that these actually produce correct output. I've verified is actually does for empty input, some random numbers, and a paragraph of lorem-ipsum, but, you know.

In the mean time, I'm sure there's people who have strong opinions about the interface of this thing and how I put it into zuse. Let's discuss!

### DIFF
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -3015,6 +3015,12 @@
     ::  core
     ::
     ++  hash
+      :>  per:  permutation function with configurable width.
+      :>  pad:  padding function.
+      :>  rat:  bitrate, size in bits of blocks to operate on.
+      :>  cap:  capacity, bits of sponge padding.
+      :>  out:  length of desired output, in bits.
+      :>  inp:  input to hash.
       |=  $:  per=$-(@ud $-(@ @))
               pad=$-([octs @ud] @)
               rat=@ud
@@ -3051,11 +3057,11 @@
     ++  sponge
       :>  sponge construction
       ::
-      :>  premute:  permutation function with configurable width.
+      :>  preperm:  permutation function with configurable width.
       :>  padding:  padding function.
       :>  bitrate:  size of blocks to operate on.
       :>  capacity:  sponge padding.
-      |=  $:  premute=$-(@ud $-(@ @))
+      |=  $:  preperm=$-(@ud $-(@ @))
               padding=$-([octs @ud] @)
               bitrate=@ud
               capacity=@ud
@@ -3064,7 +3070,7 @@
       ::  preparing
       =+  bitrate-bytes=(div bitrate 8)
       =+  blockwidth=(add bitrate capacity)
-      =+  permute=(premute blockwidth)
+      =+  permute=(preperm blockwidth)
       ::
       |=  [input=octs output=@ud]
       |^  ^-  @
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -3022,7 +3022,7 @@
       :>  out:  length of desired output, in bits.
       :>  inp:  input to hash.
       |=  $:  per=$-(@ud $-(@ @))
-              pad=$-([octs @ud] @)
+              pad=$-([octs @ud] octs)
               rat=@ud
               cap=@ud
               out=@ud
@@ -3045,11 +3045,12 @@
       |=  dsb=@ux
       ?>  (lte dsb 0xff)
       |=  [inp=octs mut=@ud]
-      ^-  @
+      ^-  octs
       =.  mut  (div mut 8)
       =+  pal=(sub mut (mod p.inp mut))
       =?  pal  =(pal 0)  mut
       =.  pal  (dec pal)
+      :-  (add p.inp +(pal))
       ::  padding is provided in lane bit ordering,
       ::  ie, LSB = left.
       (cat 3 (con (lsh 3 pal dsb) 0x80) q.inp)
@@ -3062,7 +3063,7 @@
       :>  bitrate:  size of blocks to operate on.
       :>  capacity:  sponge padding.
       |=  $:  preperm=$-(@ud $-(@ @))
-              padding=$-([octs @ud] @)
+              padding=$-([octs @ud] octs)
               bitrate=@ud
               capacity=@ud
           ==
@@ -3076,18 +3077,19 @@
       |^  ^-  @
         ::
         ::  padding
-        =/  padded=@  (padding input bitrate)
+        =.  input  (padding input bitrate)
         ::
         ::  absorbing
         =/  pieces=(list @)
           ::  amount of bitrate-sized blocks.
-          =+  i=(div (met 3 padded) bitrate-bytes)
+          ?>  =(0 (mod p.input bitrate-bytes))
+          =+  i=(div p.input bitrate-bytes)
           |-
           ?:  =(i 0)  ~
           :_  $(i (dec i))
           ::  get the bitrate-sized block of bytes
           ::  that ends with the byte at -.
-          =-  (cut 3 [- bitrate-bytes] padded)
+          =-  (cut 3 [- bitrate-bytes] q.input)
           (mul (dec i) bitrate-bytes)
         =/  state=@
           ::  for every piece,
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -3136,7 +3136,7 @@
         %+  can  3
         =-  (turn - |=(a=@ [1 a]))
         =+  (flop (rip 3 a))
-        (weld - (reap (sub 8 (lent -)) 0x0))
+        (weld (reap (sub 8 (lent -)) 0x0) -)
       --
     ::
     ++  keccak-f
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -2966,6 +2966,298 @@
       ::  alice decrypts with same key
       `@t`(dy.ali secret-key crypted-msg)
     --  ::test
+  ::                                                    ::
+  ::::                    ++keccak:crypto               ::  (2b7) keccak family
+    ::                                                  ::::
+  ++  keccak
+    |%
+    ::
+    ::  keccak
+    ::
+    ++  keccak-224  |=(a=@ (keccak 1.152 448 224 a))
+    ++  keccak-256  |=(a=@ (keccak 1.088 512 256 a))
+    ++  keccak-384  |=(a=@ (keccak 832 768 384 a))
+    ++  keccak-512  |=(a=@ (keccak 576 1.024 512 a))
+    ::
+    ++  keccak  (cury (cury hash keccak-f) padding-keccak)
+    ::
+    ++  padding-keccak  (multirate-padding 0x1)
+    ::
+    ::  sha3
+    ::
+    ++  sha3-224  |=(a=@ (sha3 1.152 448 224 a))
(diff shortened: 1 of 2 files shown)


sys/zuse.hoon
```diff
@@ -2966,6 +2966,298 @@
...
+      =?  pal  =(pal 0)  mut
+      =.  pal  (dec pal)
+      ::  padding is provided in lane bit ordering,
+      ::  ie, LSB = left.
+      (cat 3 (con (lsh 3 pal dsb) 0x80) inp)
+    ::
+    ++  sponge
+      :>  sponge construction
+      ::
+      :>  premute:  permutation function with configurable width.
```
belisarius222: premute -> permute

belisarius222: oh I see, nevermind. Maybe rename that, though.

sys/zuse.hoon
```diff
@@ -2966,6 +2966,298 @@
...
+    ++  rawshake-224  |=([o=@ud i=@] (rawshake 1.088 512 o i))
+    ::
+    ++  rawshake  (cury (cury hash keccak-f) padding-rawshake)
+    ::
+    ++  padding-rawshake  (multirate-padding 0x7)
+    ::
+    ::  core
+    ::
+    ++  hash
+      |=  $:  per=$-(@ud $-(@ @))
```
belisarius222: These args deserve individual comments.

belisarius222: Also, does the input need to be supplied with a specific endianness, or other requirements?

Fang-:
Okay, so I did a bit more digging and experimenting, and every Urbit atom that gets passed in needs to be made big-endian for processing.

This... gets a bit hairy wrt input length though, and how that's currently treated. (ie `0b1000` (4 bits) becomes `0b0001` aka `0b1` (1 bit).) We can't really take input with lots of leading zero bits, because then it'll just register as shorter input. So, we need to be able to explicitly specify the amount of bits our input has.  
I think the best solution there is to make the interface not `input=@` but `input=?(@ [@ud @])`, so that the input length (bits) can be specified if needed. If not explicitly specified, we should assume it's `(mul 8 (met 3 input))` aka bytes without leading null bytes, which is a good fit for the most common use case: hashing a string.

This ties into a small issue I discovered that causes this to not fully adhere to the spec: it can't take input that isn't a multiple of 8 bits. More precisely, the padding function operates on bytes, and so our input gets treated as bytes rather than bits. This makes processing input like `0b1101` incorrect.

Should be a fairly easy fix, will have that pushed soon.

cgyarvin:
You run into this frequently when implementing crypto algos from languages that don’t understand big numbers. It *should* be just an atom but it’s something more complicated...

ohAitch:
you definitely don't want ?(@ [@ @]), just ++octs. Which I think is how our
existing external-crypto functions work?

Fang-:
Oh yeah, almost forgot about that. But judging by its data type (`@t`), name, and `++as-octs` (`(met 3 ...`) that's a pair of byte-length and content, not bit-length and content. Of course the types still work out, but it's a semantic mismatch.

(also imo it's nice to have the option of a simpler interface (just `@`), but this may lead to unexpected behavior in uncommon cases so, uh, I concede)

cgyarvin:
Good catch, Anton.

You’ll only be using keccak in standard ways on standard widths, all of which should fit ++octs...

Fang-:
Pushed fixes and tests. For keccak, we're taking a handful of inputs from the known answer tests. For all others, since they only differ from keccak in their parameters, we test just a single null input

Test file is a tad ugly, I couldn't get the logic in the keccak test arms to work without adding a face to core state. Maybe I'm just stupid though.

As for performance, using a 255-byte input it takes about 40 seconds to generate 100 keccak-512 hashes.  
`(turn (gulf 1 100) |=(a=@ (keccak-512:keccak:crypto 255 (add input a))))`  
That's under half a second per hash, which isn't too shabby for Hoon code without new jets. And in most of my upcoming cases (Ethereum integration), inputs will be significantly shorter. (For 100 bytes, the above takes about 21 seconds. For 60 bytes, a mere 11.)

As far as I'm concerned, this is ready to be merged after final review.

Fang-:
Actually, are there even any use cases where you want the hash value instead of just the string? Thinking we should maybe just output a `tape` or `@t` instead of an `@`...

cgyarvin:
That’s going too far!  But I can’t say why :-)

Fang-:
At the very least we want helper functions *somewhere* for pretty-printing `@ux` in a real-world-usable format. That is, without the `.`s and with leading zeroes where appropriate. We'll want that for `@ub`, `@ud`, `@sd` too, but this would all be a separate PR since it's not strictly related.

Current status: still ready for merge afaic.

cgyarvin:
We already have @i but need more of that.  Easy to add.

belisarius222:
I’ll review this PR today.

I had the same feeling as @cgyarvin about returning a string. I think the
feeling might come from some idea that the lack of use of the formal
numeric output is only incidental. If we didn’t provide a public interface
for it, a use case that appears later would require us to change this code.

Better to implement the convenient semantics on top of the formal ones.

belisarius222:
@eglaysher @Fang- I wonder if you need a face on the testing library because of @ohAitch's change to `++mold` in https://github.com/urbit/arvo/pull/644 ?  I feel like that's not the case, but I'm not sure.

tests/zuse/crypto/keccak.hoon
```diff
@@ -0,0 +1,333 @@
...
+::  answer for rawshake-256 was found here:
+::  https://github.com/maandree/libkeccak/blob/master/test.c
+::  for rawshake-128, the implementation was assumed to be correct at the time
+::  these tests were written.
+::
+::
+/+  tester
+=,  keccak:crypto
+::
+=/  keccak-inputs=(map @ud @)
```
belisarius222: Please document what the keys of this map represent.

tests/zuse/crypto/keccak.hoon
```diff
@@ -0,0 +1,333 @@
...
+      d79d.c266.f6a5.be6b.b0e4.a92e.ceeb.aeb1
+  ==
+::
+|_  tst=tester-type:tester
+::
+::  keccak
+::
+++  test-keccak-224
+  ^+  tst
+  =/  answers=(list (pair @ud @))
```
belisarius222: Please document what the `@ud` is here.

tests/zuse/crypto/keccak.hoon
```diff
@@ -0,0 +1,333 @@
...
+        d6d8.0d32.1906.31e2.f06c.22ad
+    ::
+      :-  255
+      0x5af5.6987.ea9c.f11f.cd0e.ac5e.bc14.b037.
+        365e.9b11.23e3.1cb2.dfc7.929a
+    ==
+  |-
+  ?~  answers  tst
+  =.  tst
+    %^  expect-eq.tst
```
belisarius222: I think this would be easier to read with aliases for `p.i.answers` and `q.i.answers`.

tests/zuse/crypto/keccak.hoon
```diff
@@ -0,0 +1,333 @@
...
+        eef3.a0ef.94ca.65d4.448a.cf1e.892b.8ebb.
+        0887.6318.04dd.64e1.53ad.41fa.e012.7a85
+    ::
+      :-  255
+      0x6bff.1c84.05a3.fe59.4e36.0e3b.ccea.1ebc.
+        d509.310d.c79b.9e45.c263.783d.7a5d.d662.
+        c678.9b18.bd56.7dbd.da15.54f5.bee6.a860
+    ==
+  |-
+  ?~  answers  tst
```
belisarius222: This code repeats for each keccak length. It should be a helper function, and we might want to collect the inputs and answers into a single map with the keccak length as the key. Not sure about the latter, but we should not have code repeating like this.

ohAitch:
> At the very least we want helper functions *somewhere* for pretty-printing `@ux` in a real-world-usable format. That is, without the `.`s and with leading zeroes where appropriate. We'll want that for `@ub`, `@ud`, `@sd` too, but this would all be a separate PR since it's not strictly related.

Pull `x-co` `d-co` etc out of `++co`, maybe? Like, we mostly have the parsers and printers for all of these, they're just in cores in hoon.hoon

-1 on doubling the `@u` namespace in the manner of `@ui`, these definitely do not belong as valid hoon literals.

cgyarvin:
Think Anton’s right here...

tests/zuse/crypto/keccak.hoon
Fang-: Very fair. Also not sure about grouping answers for different hashing functions, so I opted to leave them under their own arms for now.

All done!

Fang-:
Forgot to mention the tests depend on #659, but that will be in soon.

PR #658, opened 2018-03-07
Get last x messages from hall

### LINKED ISSUE #643: Can't query hall for "last x messages"
Original title: **web talk hangs if there are no messages**

The page chrome loads, but the spinner goes indefinitely. Instead, it should display something like "No messages in the last 6 hours."

Fang-:
With this change, it becomes possible to specify a relative range in hall subscriptions to refer to the last x messages.

For example:  
`/circle/inbox/grams/-5` gets the last 5 messages from the inbox, and updates on all that follow.  
`/circle/inbox/grams/-5/-0` gets the last 5 messages from the inbox, nothing more.  
`/circle/inbox/grams/0/-5` gets all messages from the inbox, up to but not including the last five.

Subscriptions with a relative range start are always active, they will receive updates until their range end is reached.  
Subscriptions with a relative range end end immediately.

Tested this to update cleanly from the previous hall version, and to not cause connectivity issues when talking to out-of-date versions.  
Subscriptions to older ships that make use of this new functionality/interface actually go through somehow, but count as rangeless subscriptions. Once the out-of-date ship updates, the subscriptions get seen and dealt with as appropriate.

Closes #643.  
cc @vvisigoth

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -1210,6 +1210,7 @@
               ?=(^ tal.u.ran.src)
             ::
               ?-  -.u.tal.u.ran.src
+                $sd   &
                 $da   (gte now.bol +.u.tal.u.ran.src)
                 $ud   ?&  ?=(^ seq)
                           (gte u.seq +.u.tal.u.ran.src)
@@ -1230,7 +1231,13 @@
       ::  fill in empty ranges to select all grams.
       =.  ran
         ?~  ran  `[[%ud 0] `[%ud count]]
-        ?~  tal.u.ran  `[hed.u.ran `[%ud count]]
+        =*  hed  hed.u.ran
+        =?  hed  ?=($sd -.hed)
+          [%ud (sub count (min count (abs:si +.hed)))]
+        ?~  tal.u.ran  `[hed `[%ud count]]
+        =*  tal  u.tal.u.ran
+        =?  tal  ?=($sd -.tal)
+          [%ud (sub count (min count (abs:si +.tal)))]
         ran
       ::  never fails, but compiler needs it.
       ?>  &(?=(^ ran) ?=(^ tal.u.ran))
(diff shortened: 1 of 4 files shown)


ohAitch:
Should this be {date}-{n} or such? Being able to make two identical
subscriptions and get different results seems not in line with how gall
subscriptions usually work; tho I guess a bunch of one-offs do have
implicit current dates in this manner.

@cgyarvin I expect you to have more detailed architectural ideas than me
here

vvisigoth:
Thanks @Fang- !

c-johnson:
This is excellent. Thank you!

vvisigoth:
@ohAitch Copy and pasting Mark and CGY's prior correspondence about this for posterity:

FROM: CGY

TDLR: go ahead.

A lot of the stuff we're doing in the %gall layer feels like it will
need to be rearchitected, but that's fine -- we're developing the
collective experience needed to do it right.

(In particular, I am really not happy with the concept of a %gall
"application" -- I think it is very misleading to use the word "app,"
which implies a 1:1 connection with some kind of UI.  These things are
daemons.  They possibly could be considered actors, but they are
really a subset of actors, because the classical actor model lets
actors spawn more actors.  Which is a complication I really don't
think we need in a personal server.  I always love to remove unneeded
generality.  However, if %gall daemons keep multiplying as we grow,
consistent with Conway's law, we might need this complication!  In
general, I feel that all state in %gall daemons needs to be understood
carefully to think about how it might be data rather than metadata,
and if so find a deeper home in a vane, probably %clay.)

I think you're sort of thinking of referential transparency in the
wrong way -- it's a sort of "folk referential transparency."  Actually
I'm not quite sure what you mean *literally* by RT in this message,
but I sort of understand what you mean *intuitively*.  Intuitively, I
think what you are thinking of is almost something more like a CRDT or
some subset of it.  Let me explain my own thinking.

The purpose of a subscription is to let the client create and maintain
a synchronized view of a changing resource on, or query against, the
server.  Like "the last 10 messages."  Since this is an inherently
small query, it is particularly simple, as we don't need to
distinguish deltas from replacements.

As far as I know or rather last time I looked, the :hall queries we
normally use (all messages since some date) are also, of course,
mutable.  That is, they are lists we add messages to.  The deltas have
this CRDT-like property, which is nice but which we don't really
depend on -- they are additions.

And of course, we never actually build the data structure we're
nominally updating.  We never build the full query result.  We just
send the new messages into a message database.  Any mailbox is
inherently conflict-resilient.

So in a sense, it's really the "last 10 messages" type of query that's
the simple, common case, and what we're doing with message
subscriptions now is the unusual and difficult case.  (Curious to hear
exactly what you mean by "And really, hall isn't very well-equipped
currently to deal with date-based queries effectively.")

Fang-:
Merging because this has been in use by frontend devs for a while now and is certified Fine.

PR #659, opened 2018-03-07
Make +test only run arms starting with ++test-

Fang-:
My expectation was that this is how it worked, because the testing documentation mentions that testing arms start with `++test-` (or `++check`, not sure why). This isn't the case though, as adding a `++helper` into the main core causes it to try running that as a test anyway and crashing.

This PR makes sure that only arms starting with `++test-` are actually run when doing tests.

### DIFF
--- a/gen/test.hoon
+++ b/gen/test.hoon
@@ -20,7 +20,8 @@
   ?.  matches
     ~
   ?-  -.q.b
-    %&  (run-test [p.b pax] eny p.q.b)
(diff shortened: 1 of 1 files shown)


eglaysher:
You probably expected this because I said that it did this. I said that it did this because I thought I implemented this.

Apparently I forgot. Sorry. You'll also want to run ++check- arms, which are quickcheck style tests.

PR #661, opened 2018-03-08
fix #660: run +test in CI

### LINKED ISSUE #660: CI should run +test
...and fail when `FAILED` or `CRASHED` gets seen in output.

This will get more useful as we get more tests, but we already have *some* so why not put them to good use?

ohAitch:


gen/test.hoon
```diff
@@ -10,12 +10,15 @@
 ++  test-runner
   :>  run all tests in {a} with a filter.
   =|  pax=path
-  |=  [filter=path eny=@uvJ a=tests:tester]
+  |=  [filter=path [defer=? eny=@uvJ] a=tests:tester]
```
belisarius222: The `defer` argument needs a comment explaining its affect on behavior.

.travis/test.ls
```diff
@@ -1,32 +1,42 @@
...
+<- on-next /%got-dojo/
+
+exit-code = 0
+
+urbit.write "+test, =defer |, =seed `@uvI`(shaz %reproducible)\r"
+on-next /(FAILED|CRASHED)/ ->
+  console.log "\n\n---\nnode: detected error\n---\n\n"
+  exit-code := 2
+
+urbit.write "%tested\r"
```
belisarius222: There should be a comment here explaining that this is how we know the tests have  completed.

ohAitch:
PR #666 of DOOM cuts this down from 4 minutes [down to 3.5](https://travis-ci.org/ohAitch/arvo/builds/351596182), which is a slightly more impressive 30% cut if you factor out that [2.5 of those minutes are spent just starting urbit](https://travis-ci.org/urbit/arvo/builds/351489106) according to master.

I should probably put in tracking for how long each test section takes.

PR #672, opened 2018-03-13
add ++lest, the non-empty list mold builder

frodwith:


ohAitch:
Is there a reason it's defined corecursively with `++list`, instead of just having the structure in both definitions?

keatondunsford:
@ohAitch @frodwith If you guys think this is a good thing to have in Hoon and agree on the bad corecursive definition, want to make the structure change, then I'll merge this?

frodwith:
Does the corecursive definiton cause a problem?

keatondunsford:
I personally don't think it matters.

ohAitch:
It causes you to have to look up the other definition to visually parse the mold; but the other definition is approximately right there even with the doccord gap, so probably just merge it.

PR #698, opened 2018-04-04
3-fixed (unary sample) %cnsg

joemfb:
forward-ports and stages #381

/cc @ohAitch

### DIFF
--- a/.travis/pin-parent-pill-pier.url
+++ b/.travis/pin-parent-pill-pier.url
@@ -1,1 +1,1 @@
-https://ci-piers.urbit.org/zod-ccaffc55e6cd2f244e6fd1710479c05e1019c167.tgz
+https://ci-piers.urbit.org/zod-807652e1ab88dd7e7edc7ca8593a2ba4fb1f8919.tgz
--- a/.travis/pin-parent-pill-pier.url
+++ b/.travis/pin-parent-pill-pier.url
@@ -1,1 +1,1 @@
-https://ci-piers.urbit.org/zod-807652e1ab88dd7e7edc7ca8593a2ba4fb1f8919.tgz
+https://ci-piers.urbit.org/zod-b5aeee11e463f9338909ede5cee13d30883ea807.tgz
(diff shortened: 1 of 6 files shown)


ohAitch:
you'll want to push the intermediate commits so the pilling chain actually goes through, and after the first one bump `.travis/pin-parent-pill-pier.url` two times to refer to the commits they respectively compile under
do you have permissions to push tags to **urbit/arvo**? I can probably take this over if not

joemfb:
If you could, that would be great! I don't think I have any access to directly manipulate this repository.

ohAitch:
(alternately, the staging can be avoided entirely if you just push a `hoon.hoon` with `=- ~(gas by -)` instead of the two uses of tall form `%~`, which I maybe could have thought of before  you wrote these sorry)

joemfb:
It's no problem, these weren't exactly difficult changes. Feel free to cherry-pick whatever's useful here.

ohAitch:
Thanks, I grabbed ++expn into db7120d which [at this rate](https://travis-ci.org/urbit/arvo/builds/362296092) looks like it'll succeed :3

ohAitch:
huh it occurs to me that this *is* totally necessary for deployment to live network, hmm

ohAitch:
@keatondunsford this is now deployable I think, though make sure to go through all the commits individually. (in theory, you could do just the three checkmark ones, but I remember there being issues where a reboot and app-reload in the same commit would be sequenced incorrectly)

keatondunsford:
@ohAitch Cool. I'll get to this and the other PRs soon when I finish up a couple of other things. Thanks for checking on this.

hoclun-rigsep:
Ah, very cool. I regret not having seen this through myself.

keatondunsford:
@hoclun-rigsep Thanks for catching the original issue! ~zod's work continues.

hoclun-rigsep:
On April 11, 2018 7:02:38 PM EDT, Keaton Dunsford <notifications@github.com> wrote:
>@hoclun-rigsep Thanks for catching the original issue! ~zod's work
>continues.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/urbit/arvo/pull/698#issuecomment-380622464

You haven't heard the last of me...
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #721, opened 2018-04-25
adds lane-change subscriptions (%tend/%rove) to ames

joemfb:
This PR extends ames to support subscriptions that receive notifications of any route (`+lane` changes on child ships. This passes a basic local smoketest, but I still need to setup a fake-net across multiple machines to test this properly.

Does the `%tend` subscription move need to be paired with an acknowledgement upon successful 
`(~(put in *(set duct) hen)`?


vvisigoth:
Where did we land on doing a multi-machine test for this? @joemfb Is this the one @cgyarvin was referring to when he said not to test it into the ground?

cgyarvin:
I would test this as a component of whatever it’s coupled to.  Otherwise the harness is as much work as the (pretty simple) feature.

joemfb:
superceded by #752

Issue #674, opened 2018-03-15
%bad-text, and %ap-lame %hood error when activating link in Talk


joshuareagan:
In the terminal I tried to activate the Ford UP 4 link, and got a %bad-text crash, a stack trace, and then an `[%ap-lame %hood %diff-sole-effect-drum-phat]` message.  After that I couldn't send Talk messages on any channel, until I unlinked and then relinked Talk.  Then I could rejoin /urbit-meta and message normally.  I've been able to reproduce this error locally.

joshuareagan:
I tried activating the same link with a comet, got some %bad-texts, but the fora message went through.

keatondunsford:
I think this is because we were working on this in a Google Doc, and originally I copied and pasted the raw Google Doc content which naturally uses smart quotes and other non-ASCII characters. So I manually made those changes and the file seems fine now, but perhaps the activation link of the Talk messages still points to the old revision of that Fora post.

Closing this for now since I'm pretty sure activation of other Talk Fora notifications work, but perhaps these kinds of errors should be dealt with in the future.

Thanks!

belisarius222:
It's possible for new Fora posts to include Unicode and break people's Talk
apps, so I think this should be reopened until we either:

a) prevent unicode in Fora posts, or
b) get our terminal to handle printing unicode code points, or
c) sanitize Fora posts before trying to print them on the terminal

Fang-:
Thanks Ted, I was about to post the same.

I'd also like to offer the additional option of sanitizing unclean `@t`s before printing them, rather than going "bad text" and not printing anything. Slightly broader version of your point c I guess.

Your option a is a no-go as far as I'm concerned. Let's not artificially limit our systems like that.

Fang-:
Also, drum doesn't explicitly crash on `%bad-text`s, they just get no-opped. So this is crashing somewhere down the event flow, and it probably shouldn't.

cgyarvin:
(d) build a time machine, go back in time and  kill the baby that will grow up to invent unicode 

Unicode in markdown shouldn’t actually be a problem.  Would be interesting to find out what’s actually crashing and how.

joemfb:
The terminal already prints unicode codepoints! The drum `%bad-text` printf is on text that is not `(sane %t)`: invalid utf-8.

cgyarvin:
The question is how we got bad Unicode in...

joemfb:
[...]

PR #725, opened 2018-04-26
updates +win:re to wrap lines safely with regards to unicode

joemfb:
(by converting the `tape`s in `+tank` to (list @c) before wrapping, then back)

This is a fix for unsafe unicode handling in general, and #674 specifically.

The "printable" structure `+tank`, designed for stack traces and type printing and slated for replacement, is heavily used in printing to the console. The renderer `+re` converts `+tank` to `+wall`, optionally wrapping long lines to a specified width. But since `+tank` consists of various combinations of `+tape`, which is a linked list of utf8 bytes, it's possible for `+re` to split multi-byte unicode codepoints, resulting in `%bad-text`.

For example:

```
> (~(win re [%leaf "abc’d"]) 0 5)
[%bad-text "[60 60 34 92 92 47 97 92 92 47 34 32 34 32 32 98 34 32 34 32 32 99 34 32 34 32 32 226 34 32 34 32 32 128 34 32 34 32 32 153 34 32 34 32 32 100 34 32 34 92 92 47 32 92 92 47 34 62 62 0]"]
```

This PR is a straightforward hack to make line-wrapping safe: first we make a copy of the `+tank`, converting all the tapes within to lists of codepoints, then we wrap and render the lists of codepoints, then we convert the output back to actual tapes. It's a hack because a) it's not very efficient and b) we have to lie to the type system. But it has the advantage of working without touching all the different parts of urbit that create a `+tank`, which, again, is slated for removal.

```
> (~(win re [%leaf "abc’d"]) 0 5)
<<"abc’d">>
```

These changes are all internal to `+re`, so there shouldn't be any issues pushing this out to the live network. I still need to reproduce the exact fora post scenario from #674 and confirm that no other issues are involved.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -4118,17 +4118,17 @@
       (weld ram(tac i.q.tac) ?~(t.q.tac voz (weld p.p.tac voz)))
     ==
   ::
-  ++  gyp                                               :: horrible hack
+  ++  ug                                                ::  horrible hack
     |%
-    ++  hew
-      |=(a=tape `tape``(list @)`(tuba a))
+    ++  ace                                             ::  strip ctrl chars
+      |=  a=tape
+      ^-  tape
+      ?~  a  ~
+      ?:  |((lth i.a 32) =(127 `@`i.a))
+        $(a t.a)
+      [i.a $(a t.a)]
     ::
-    ++  fix
-      |=  wol=wall
-      %+  turn  wol
-      |=(a=tape (tufa `(list @c)``(list @)`a))
-    ::
-    ++  act
+    ++  act                                             ::  pretend tapes
       |=  tac=tank
       ^-  tank
       ?-  -.tac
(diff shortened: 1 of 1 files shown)


joemfb:
I've tested this fix against the fora post in question (the Ford UP), and it's still crashing. But now it's because of the presence of a tab character.

I think we should just strip all control characters out of cords and/or tapes that enter from the Outside. Or we could just escape them with `+mesc`. Either way, should this just be changed specifically in the fora-post code path, or should sanitation be generally applied in existing conversions (such `+to-wain:format`, `+tuba`, `+trip` and/or others)?

cgyarvin:
Yes but call it something else -- cores here are two-letter, so ++ug?

I like "the Outside."  Agree on stripping control characters.  I also consider \10 a control character.  You have to be 32 or taller to go on this ride.

joemfb:
I've renamed `+gyp:re` to `+ug:re`, and updated it to strip all ctrl characters from the tapes in the tank before handling the unicode conversions. This now passes my local tests with some arbitrary bad tapes, as well as the original Ford UP.

I think we'll need to review the general approach to handling cords/tapes/(list @c), etc. There should be some blessed, simple path for getting untrusted text into a safe, reliable state.

(And I can't figure how how to mark the requested changes as complete on this PR)

vvisigoth:
@cgyarvin are Joe’s changes good to go?

cgyarvin:
Merged, thanks.

Issue #681, opened 2018-03-20
decrement-underflow while accessing foreign clay node


vvisigoth:
Foreign clay node access seems to work (at least I see a lot of camping going on domestic ship) fine.
 
This error arises when you attempt to access a non-existent node on foreign clay using the `<domestic ship>/<foreign beam>/path/to/node`. Once you've attempted this, the following error occurs every time you try to access _any_ foreign node.

*given:* two ships, ~wanzod and ~binzod, run `|public %home` on both
*reproduction*
- attempt to access a non-existent foreign node, eg `http://wanzod/~binzod/==/404.hoon`
- you should see the error below on the domestic ship
- you should see this error whenever you attempt to access a foreign node, even a known-good one. 

```
[%fail-ack 0]
crude: all delivery failed!
  [%swim-call-vane %c %west ~]
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.543 3].[3.784 5]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.544 3].[3.784 5]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.557 3].[3.784 5]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.558 3].[3.784 5]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.753 5].[3.770 37]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.754 5].[3.770 37]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.755 5].[3.770 37]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.764 5].[3.770 37]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.765 5].[3.770 37]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.766 5].[3.770 37]>
  /~bidled-witwex/home/~2018.3.15..22.40.41..99e0/sys/vane/clay:<[3.767 5].[3.770 37]>
[...]

PR #726, opened 2018-04-30
Foreign ship %next requests

Fang-:
This fixes the symptom displayed in #681, but doesn't really address the underlying issue. This PR should probably be considered not ready to be merged until that has been resolved properly.

Currently, this PR only fixes the runtime error that would occur when trying to check revision zero for changes. It would try to look at the previous revision, but `(dec 0)` crashes. Now, it detects this and generates "file does not exist" results for all relevant files.

The underlying cause here is that the known latest revision for foreign ships (relevant [when converting a `%next` into a `%mult`](https://github.com/Fang-/arvo/blob/9f94497e87c28f8355c7a8628843b8f6bc02634e/sys/vane/clay.hoon#L1865) for code sharing purposes) is blindly used for `%next` requests, even when they're asking for changes on higher revisions.

That seems like something we probably want to fix. And that's doable, but also raises a question about `%next` requests. Was is ever supposed to work with requests to foreign ships? If we look at [the old `++wake` code for `%next`](https://github.com/urbit/arvo/blob/3f9223094a76a0fc6bc125191faecee0395cab57/sys/vane/clay.hoon#L1676-L1696) we can see that it always tries to read out the current version. If it can't do so synchronously, it prints `%oh-noes` and moves on.  
Unless I'm misunderstanding, it's not uncommon for it to go "oh noes" there on foreign files, since we're not guaranteed to have the latest version around. This creates problems, because `++wake` logic for `%next` (used to, and still) assumes that the check on the previous revision already happened.

Proposed solution is to extend the `++rove` for `%next` with a "next revision to check for change" variable. This has the tiny additional benefit of bringing it closer to `%mult`, implementation-wise.

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -1886,6 +1886,12 @@
         $(rov [-.rov mol `+(u.aey) ~ ~])
       ::  if old isn't complete, try filling in the gaps.
       =?  old  !(complete old)
+        ::  if we're looking for changes in revision zero, then the previous
+        ::  revision only has "file does not exist" results.
+        ?:  =(u.yon 0)
+          %-  ~(gas by *(map (pair care path) cach))
+          %+  turn  ~(tap in q.mol)
+          |=(r=(pair care path) [r [~ ~]])
         (read-unknown mol(p [%ud (dec u.yon)]) old)
       ::  if the next aeon we want to compare is in the future, wait again.
       =+  aey=(case-to-aeon:ze [%ud u.yon])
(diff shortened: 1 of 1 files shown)


Fang-:
Implemented the solution proposed in the OP. This should keep `%next` requests from checking against older revision numbers, and makes sure they don't skip any revisions.  
The only way in which `%next` and `%mult` handling differs now is the way in which they're stored.

Also removed the band-aid fix made in the original commit, because that case can't happen anymore.

belisarius222:
@Fang- Is this ready to review now?

Fang-:
Sorry, should've explicitly mentioned: yes, ready for review. (:

PR #728, opened 2018-05-02
Change %w care to return all related cases

Fang-:
Since this builds on #726's `++load` changes instead of adding a new clay version altogether, we'll want to wait for this to merge before pushing the master branch to the network.

This PR changes the result returned for `%w` clay requests from `@ud` to `[ud=@ud da=@da tas=(set @tas)]`, to aid in New Ford work.

As discussed with @belisarius222, this is not ready to be merged until either he or @eglaysher updates the places where the `%w` care is used. It seems this is just kiln and ford.  
Not to mention, I haven't taken care of the `++take-foreign-update` case. I'll get to that tomorrow.

I haven't been able to test this properly yet, since ford apparently comes into play (and trips over the changed `%w` result) even for apps that have already been built. But maybe that has to do with zuse also getting changed, idk, I'll try a bit harder tomorrow. Fairly confident this works as advertised though.

### DIFF
--- a/lib/hood/kiln.hoon
+++ b/lib/hood/kiln.hoon
@@ -412,7 +412,7 @@
           leaf+"bad %writ response"
         (render "on sync" sud her syd)
       ~
-    =.  let  ?.  ?=($w p.p.u.rot)  let  ((hard @ud) q.q.r.u.rot)
+    =.  let  ?.  ?=($w p.p.u.rot)  let  ud:((hard cass) q.q.r.u.rot)
     %-  blab  ^-  (list move)  :_  ~
     :*  ost  %merg
         [%kiln %sync syd (scot %p her) sud ?:(reset /reset /)]
--- a/lib/hood/kiln.hoon
+++ b/lib/hood/kiln.hoon
@@ -417,7 +417,7 @@
     :*  ost  %merg
         [%kiln %sync syd (scot %p her) sud ?:(reset /reset /)]
         our  syd  her  sud  ud+let
-        ?:  =(0 .^(* %cw /(scot %p our)/[syd]/(scot %da now)))
+        ?:  =(0 ud:.^(cass:clay %cw /(scot %p our)/[syd]/(scot %da now)))
           %init
         %mate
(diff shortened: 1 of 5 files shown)


Fang-:
Fixed some stupid error in my code, fixed usage in `++take-foreign-update` (it's practically unchanged, just more semantically correct now), and fixed ford to play nicely enough with this to actually allow me to test it.

I have special-cased requests about the `0` case to return the "no result ever" result, since it crashed, and we won't ever have info for it anyway.

Fang-:
It seems `[~ ~]` for `%w` requests is not supported, so for the `0` case we now return ``[0 `@da`0 ~]`` instead, which seems sane enough.

ohAitch:
Aaaaaa, -1 on it returning labels, .^ is supposed to be reverentially
transparent while labels are retroactive.

belisarius222:
That's a good point, @ohAitch. I think we should take the labels out of the
%w response, and maybe later we can add a request to retrieve the labels
for a commit.

I suspect supporting requests to resolve the label case is the reason for
%w never responding with a [~ ~]. Just because a label doesn't map to a
commit now doesn't mean it won't in the future. This should be
referentially transparent as long as labels are immutable.

ohAitch:
I'm also getting a CI error in kiln:
https://github.com/urbit/arvo/blob/7a914e4d5536eaf8830876f68a60cab7842578ba/lib/hood/kiln.hoon#L415
presumably that should be `ud:((hard cass:clay) q.q.r.u.rot)`?

Fang-:
Good call about the referential transparency of the labels, @ohAitch. I'll take the labels out of the response.

@belisarius222 had agreed to chase down the changes in non-clay files, but since these all seem like one-line fixes I'll pick them up real quick.

eglaysher:
@belisarius222 is out today. If you need help, reach out to me.

Fang-:
This should do it. I'll close this and open a PR with cleaner commits though, I've been a bit too fast-and-loose here.

PR #730, opened 2018-05-07
Vague proposal: a backwards list.

eglaysher:
Somebody in the office vaguely proposed this on Friday. (~pittyp?
~rovnys? ~sorreg?) This is a sketch that shows its possible.

Something like this should prevent confusion about whether you have
to flop a list.

Deliberate wide audience on reviewers since I'd like broad feedback on whether this is even a good idea.

### DIFF
--- a/lib/new-hoon.hoon
+++ b/lib/new-hoon.hoon
@@ -159,6 +159,59 @@
   ::  used in other files: bond, drop (but only once)
   ::  unusued: clap
   --
+++  slit
+  ::  null terminated backwards list
+  ::
+  ::    normal lists are cheap to push items to the front, but we also want a
+  ::    type which is cheap to push items to the back. one common problem is
+  ::    forgetting to reverse a list where we were pushing the last item to the
+  ::    front. +slit lets us rely on the type system so we have a different list
+  ::    with a different shape for back insertion lists
+  ::
+  |*(a=mold $@($~ [t=(slit a) i=a]))
+::
+++  sl
+  ::  backwards list
+  ::
+  |%
+  ::  +homo: homogenize +slit
+  ::
+  ++  homo
+    |*  a=(slit)
+    ^+  =<  $
+      |%  +-  $  ?:(*? ~ [t=$ i=(nags 0 a)])
+      --
+    a
+  ::  +size: length of the slit
+  ::
+  ++  size
+    |=  a=(slit)
+    =|  b=@u
+    ^-  @u
+    |-
+    ?~  a
+      b
+    $(a t.a, b +(b))
+  ::  +nags: index into list from the back
+  ::
+  ++  nags
+    |*  {a/@ b/(slit)}
+    |-  ^+  ?>(?=(^ b) i.b)
+    ?~  b
+      ~_  leaf+"nags-fail"
+      !!
+    ?:  =(0 a)  i.b
+    $(b t.b, a (dec a))
+  ::
+  ++  to-list
+    |*  a=(slit @ud)
+    =>  .(a (homo a))
+    =+  b=`(list _i.+.a)`~
+    ^+  b
+    |-
+    ?~  a  b
+    $(a t.a, b [i.a b])
+  --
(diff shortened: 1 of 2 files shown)


vvisigoth:
@belisarius222 @ohAitch @cgyarvin Y'all have any feedback on this?

eglaysher:
I put in a backwards `+snag` because it's needed as part of the machinery to implement backwards-`+homo`. And you need to have `+homo` because wet gates won't automatically homogenize things in their sample.

belisarius222:
@joemfb you might want to weigh in on this too.

I think it's a clever idea, but I'm not sure it's a good idea for the proposed use case (list of moves), and I'm generally skeptical of schemes that aim to improve usability by encoding something backward. I wonder whether we'd be better served by a standard library function that appends an item to a list (+push, perhaps). We could even add runes for adding elements to the front and back of lists, which might look something like this: 
```
=/  liz=(list @)  ~[2 3 4]
::
;-  liz 1
;+  liz  5
::
liz  ::  ~[1 2 3 4 5]
```

There's a theoretical performance cost to using normal lists and appending to them, but I doubt that cost merits swapping the order for lists of moves -- I even doubt whether any of Urbit's performance problems come from list traversal. If there were lots of use cases for backward lists, and they were supported as well as normal lists, then maybe they would be worthwhile? I feel like I'd be happier with this if we had an iterator pattern in Hoon already, since then we could drop this in as one of many kinds of iterable sequences with different semantics and performance characteristics.

By "iterators," I mean things like these:
Rust iterators: https://doc.rust-lang.org/std/iter/trait.Iterator.html#examples
Haskell Traversable: https://wiki.haskell.org/Foldable_and_Traversable

joemfb:
I like the idea in theory -- it's very convenient to push things onto the head of a list, and it can be a bit of a pain to keep track of which lists should be flopped and which shouldn't. And when I first saw this, I thought re-using the list faces would allow `+slit` to work with the existing wet gates in the stdlib.

Unfortunately, it doesn't play well with `+redo` face stacks:

```
> =foo |%  ++  slit  |*(a=mold $@($~ [t=(slit a) i=a]))  --
> ? `(slit:foo @)`[[[~ 3] 2] 1]
  ^#1.?({t/#1 i/@} $~)
[t=[t=[t=~ i=3] i=2] i=1]
> (|*(a=(list) ?~(a ~ i.a)) `(slit:foo @)`[[[~ 3] 2] 1])
t=[t=[t=~ i=3] i=2]
> (|*(a=(list) ?~(a ~ t.a)) `(slit:foo @)`[[[~ 3] 2] 1])
i=1
```

Of course, if all the list gates end up namespace'd in an `+ls` core, and then duplicated (in reverse form) in an `+sl` core, that's still pretty useable. I was just hoping for a kind of magic that isn't very hoonish ...

The existence of `+slit` would also create questions for use-cases that involve both prepending and appending (somewhat similar to the current choice between using tapes and cords). I guess I'm torn; it's a cool thing to have, but I'm not sure if it's worth the cognitive overhead.

ohAitch:
You need different nock to traverse a head-list than a tail-list, and wet gates are all about reusing nock, right.

For a deque, you'd probably want to use something like `qeu`: actually, right now the asymptotics of that are the same if you use it from both sides, the methods just aren't implemented.

cgyarvin:
I share the generally negative opinion.  To me a list is a list, and its
"reversedness" is a matter of semantic interpretation.  I don't see any
particular reason to use a rightward list to represent whether a list is a
stack, a sequence, etc.

I also just think it's way too early to redesign the standard library.
When I see a problem where (a) nothing is blocking on it, and (b) the
solution will come out better if we solve it later, I am inclined to
procrastinate productively.

In this case, we need more people who both know Hoon well and have deep FP
experience.  I don't feel we have enough of a critical mass at present, and
I worry we'll just do the job badly.

vvisigoth:
@eglaysher I move to close this. Are you satisfied with the feedback you got?

PR #649, opened 2018-02-23
fixes and documents ++spin/++spun (forwardport #385)

joemfb:
This forwardports #385, with a few additional changes:

- doccords (are these in the correct style? clear descriptions?)
- corrected typecast
- changed `++spin` argument order

The previously commented-out typecast reliably dumps core. Here's a representative backtrace: https://gist.github.com/joemfb/2884e2edf0e910beac630bef5b0c1ad0

I haven't been able to create a reproduction smaller than this:

```
|*  [a=(list) b=_|=(^ [** +<+]) c=*]
?<  ?=($-([_?<(?=($~ a) i.a) _c] [* _c]) b)
a
```

While that obviously shouldn't crash a ship (somehow not being virtualized?), I think fishing for a core like that is supposed to crash in the Hoon sense. I've switched to the common `=>  %=` typecast style.

As for the argument order, I think it's best for the gate to be last, particularly when it comes to wide form hoons.

Ex:

```
> (spin (ly 3 3 3 ~) ~(. og eny) |=([a=@ b=_og] (rads:b a)))
[p=~[2 2 0] q=<4.wmp {a/@uvJ <54.tyv 119.aiq 30.hhq 1.jmk $143>}>]
> (spun (ly 3 3 3 ~) |=([a=@ b=_~(. og eny)] (rads:b a)))
~[1 2 0]
```

These gates are unused, so there shouldn't be a problem pushing this out live.

/cc @chc4

PR #738, opened 2018-05-15
fixes +spin to preserve the type of the head of the product

joemfb:
This resolves an issue that I introduced when I refactor `+spin` and `+spun` in #649. The typecast of the sample gate `c` in `+spin` was throwing away the type of the head of the product, which is somewhat unfortunate:

```
> ? (spun "foo" |=([a=@tD b=@ud] [[b a] +(b)]))
  it(*)
~[[0 102] [1 111] [2 111]]
```

With this fix in place:

```
> ? (spun "foo" |=([a=@tD b=@ud] [[b a] +(b)]))
  it({@ud @tD})
~[[0 'f'] [1 'o'] [2 'o']]
```

My apologies for insufficiently testing the previous PR.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -650,7 +650,7 @@
   :>  c: gate from list-item and state to product and new state
   ~/  %spin
   |*  [a=(list) b=* c=_|=(^ [** +<+])]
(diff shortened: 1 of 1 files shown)


keatondunsford:
Seems legit.

PR #742, opened 2018-05-25
Transition ~marzod's Ask duties to a dedicated ship

keatondunsford:
This commit allows Ask app responsibility/load to get taken off of
~marzod and assigned to a dedicated Ask ship: currently, ~socden-malzod,
a child of Tlon's ops star, ~dopzod.

### DIFF
--- a/lib/hood/helm.hoon
+++ b/lib/hood/helm.hoon
@@ -107,8 +107,8 @@
 ::
 ++  poke-send-ask
   |=  mel/cord  =<  abet
-  %^  emit  %poke  /helm/ask/(scot %p ~marzod)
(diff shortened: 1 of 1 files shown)


cgyarvin:
Why?  This must be negligible load.  It’s one more server to watch and no one will tell us if it goes down...

keatondunsford:
@cgyarvin I plan on setting up ping notifications for it. One problem I had towards the end of last era, as ~marzod's load increased, was barely being able to send out invites. It's not that the load is negligible to ~marzod; it's that ~marzod's load is non-trivial for an always-on Ask ship.

cgyarvin:
I see.

Issue #749, opened 2018-06-14
Clay permissions scry is too opaque


ixv:
The `+dict` structure returned by scrying for clay permissions is too opaque a structure.

The issue is inside `+rule`, it refers to a `(set whom)` where `+whom` is `(each ship @ta)`, and the `@ta` is a key in a map to permissions groups, internal to clay. The values of this map are ignored in a scry, ie: when you scry for permissions, you can get these black box `@ta` keys returned to you.

Fang-:
You can send a `[%crew our=ship]` move to clay to get back a `[%cruz cez=(map @ta crew)]` (where `crew` is `(set ship)`).

If we were to send back "unfolded" permissions here, then there'd be no way to get the relevant permission groups from a rule.

PR #750, opened 2018-06-18
clay permissions issue #749

ixv:
Fixes https://github.com/urbit/arvo/issues/749

Since dict is only used once in clay, in the place that gets permissions for a scry, I changed its definition to:
```
++  dict  {src/path rul/real}
```
where real is
```
++  real  {mod/?($black $white) who/(set ship)}
```
And made the necessary changes in clay to handle this. 

Perhaps I should return all the group information rather than reducing it to a `(set ship)`. If anyone has any objections I can change it to `(set (each ship (map @ta crew)))`, or something else if you have a better suggestion.

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -2370,10 +2370,19 @@
       (read-p-in pax pew.red)
     ::
     ++  read-p-in
+      !:
       |=  {pax/path pes/regs}
       ^-  dict
-      =+  rul=(~(get by pes) pax)
-      ?^  rul  [pax u.rul]
+      =/  rul=(unit rule)  (~(get by pes) pax)
+      ?^  rul
+        :+  pax  mod.u.rul
+        %-  ~(rep in who.u.rul)
+        |=  {w/whom out/(set @p)}
+        ?:  ?=([$& @p] w)
+          (~(put in out) +.w)
+        =/  cru=(unit (set @p))  (~(get by cez.ruf) +.w)
+        ?~  cru  out
+        (~(uni in out) u.cru)
       ?~  pax  [/ %white ~]
       $(pax (scag (dec (lent pax)) `path`pax))
     ::
@@ -2407,14 +2416,12 @@
     ++  allowed-by
       |=  {who/ship pax/path pes/regs}
       ^-  ?
-      =+  rul=rul:(read-p-in pax pes)
-      =-  ?:(?=($black mod.rul) !- -)
-      %-  ~(rep in who.rul)
-      |=  {w/whom h/_|}
-      ?:  h  &
-      ?:  ?=($& -.w)  =(p.w who)
-      (~(has in (fall (~(get by cez) p.w) ~)) who)
-    ::
+      =/  rul=real  rul:(read-p-in pax pes)
+      =/  in-list=?  (~(has in who.rul) who)
+      ?:  =(%black mod.rul)
+        !in-list
+      in-list
+    ::  ::
     ::  Checks for existence of a node at an aeon.
     ::
     ::  This checks for existence of content at the node, and does *not* look
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -2376,8 +2376,8 @@
       ?^  rul
         :+  pax  mod.u.rul
         %-  ~(rep in who.u.rul)
-        |=  {w/whom out/(pair (set @p) (map @ta crew))}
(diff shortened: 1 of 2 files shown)


Fang-:
See also my comment on #749, it's maybe a non-issue.

`(set (each ship (map @ta crew)))` sounds like a nice change though, I must admit. I'd probably review/merge that.

sys/vane/clay.hoon
```diff
@@ -2372,9 +2372,17 @@
...
       |=  {pax/path pes/regs}
       ^-  dict
-      =+  rul=(~(get by pes) pax)
-      ?^  rul  [pax u.rul]
-      ?~  pax  [/ %white ~]
+      =/  rul=(unit rule)  (~(get by pes) pax)
+      ?^  rul
+        :+  pax  mod.u.rul
+        %-  ~(rep in who.u.rul)
+        |=  {w/whom out/(pair (set @p) (map @ta crew))}
```
Fang-: Might want to use `ship` in place of `@p`, like the structure definition in zuse does.

sys/vane/clay.hoon
```diff
@@ -2407,14 +2415,18 @@
...
-      =-  ?:(?=($black mod.rul) !- -)
-      %-  ~(rep in who.rul)
-      |=  {w/whom h/_|}
-      ?:  h  &
-      ?:  ?=($& -.w)  =(p.w who)
-      (~(has in (fall (~(get by cez) p.w) ~)) who)
-    ::
+      =/  rul=real  rul:(read-p-in pax pes)
+      =/  in-list=?  
+        ?|  (~(has in p.who.rul) who)
```
Fang-: This would be better for readability:

```
?|  (~(has in p.who.rul) who)
  ::
    %-  ~(rep ...
```

sys/vane/clay.hoon
```diff
@@ -2372,9 +2372,17 @@
...
       ^-  dict
-      =+  rul=(~(get by pes) pax)
-      ?^  rul  [pax u.rul]
-      ?~  pax  [/ %white ~]
+      =/  rul=(unit rule)  (~(get by pes) pax)
+      ?^  rul
+        :+  pax  mod.u.rul
+        %-  ~(rep in who.u.rul)
+        |=  {w/whom out/(pair (set @p) (map @ta crew))}
+        ?:  ?=([$& @p] w)
```
Fang-: You're mixing old and new syntax here, as well as other places in the pr. Clay in general still uses old syntax, you might want to stick to that for consistency, but since this change is fairly self-contained (in clay) I see no harm in applying new syntax for everything you touch. Just be consistent about it.

sys/vane/clay.hoon
```diff
@@ -2407,14 +2415,18 @@
...
+        ?|  (~(has in p.who.rul) who)
+            %-  ~(rep by q.who.rul)
+            |=  [[@ta cru=crew] out=_|]
+            ?:  out  &
+            (~(has in cru) who)
+        ==
+      ?:  =(%black mod.rul)
+        !in-list
+      in-list
+    ::  ::
```
Fang-: Extra `::` can be removed.

ixv:
Updated as per your suggestion.

The reason I think its an issue is that it's weird to have to make a synchronous scry combined with an asynchronous call to clay in order to know for sure if someone should or should not have permissions to a thing. It comes up when your in a position where you can't make any async calls at all.

ixv:
Thanks Fang-,  changes made! I did test it yeah.

sys/vane/clay.hoon
```diff
@@ -2407,13 +2415,18 @@
...
-      |=  {w/whom h/_|}
-      ?:  h  &
-      ?:  ?=($& -.w)  =(p.w who)
-      (~(has in (fall (~(get by cez) p.w) ~)) who)
+      =/  rul=real  rul:(read-p-in pax pes)
+      =/  in-list=?  
+        ?|  (~(has in p.who.rul) who)
+          ::
+            %-  ~(rep by q.who.rul)
+            |=  {{@ta cru=crew} out/_|}
```
Fang-: Going by your changes you want to stick to old syntax. Friendly reminder that `=` isn't old syntax, `/` is. (Here, and every `=/` you used.)

(In case it was the previous usage with `=+` that threw you off, the `=` for `=+` is different ("give value a face") from the `=` for `=/` ("give face a type", sort of).)

ixv:
argh, thanks for being patient Fang

PR #752, opened 2018-06-21
adds dns app, http(s) server-config and proxy notification moves

joemfb:
This PR includes:

- new `%eyre` moves for the reverse reverse TCP proxy (`%wise` and `%that`)
- new `%eyre` moves for http(s) server config (`%form` and `%live`)
- new `%ames` moves for monitoring child-ship IP address changes (`%rove` and `%tend`)
- the dynamic DNS app and related components

These changes are all additive -- they do not necessarily require a breach. They are properly paired with an updated vere, however (I'll link those changes shortly). If run on a current-or-older vere, server configuration and reverse-proxy notification effects will simply be ignored.

### DIFF
--- a/app/dns.hoon
+++ b/app/dns.hoon
@@ -5,6 +5,7 @@
 +=  move  (pair bone card)
 +=  poke  $%  [%dns-bind for=ship him=ship target]
               [%dns-bond for=ship him=ship turf]
+              [%dns-authority authority]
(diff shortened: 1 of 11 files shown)


joemfb:
Related vere changes pushed to https://github.com/urbit/urbit/tree/https-proxy-config

PR #754, opened 2018-06-21
Cookies fix

ixv:
This changes the annoying behavior where eyre attempts to log you in as a foreign ship if it detects you have cookies set for that ship.


joemfb:
Should this be against the `collections` branch? This would disable SSO across the network, breaking fora and anything similar.

ixv:
Oh yeah, good point

PR #755, opened 2018-06-27
removes dead +mz:nl, updates +bif syntax

rmariani:
`+mz:nl` was determined by ~sorreg to be dead code, and there doesn't seem to be any possible `a` that will nest in it.

Also, `+bif:in` and `+bif:by` had the weird syntactical construct of `=<  [+< +>]`. I simplified it to the more straightforward `=< +`.


ohAitch:
That seems to be being over-enthusiastically clear about producing a pair, yeah

PR #765, opened 2018-07-19
adds letsencrypt client :acme, integrates with :dns app and %eyre

joemfb:
This PR adds the new LetsEncrypt client app `:acme`, and builds on the `:dns` app and `%eyre` changes in #752.


joemfb:
These changes require a binary built from https://github.com/urbit/urbit/tree/https-proxy-config.

PR #766, opened 2018-07-20
crash `++ rad` when arg of `0` is used

joshuareagan:
(I messed up a previous version of the PR, so I closed that and am trying again with this one.)

@rmariani ran into an endless loop when he tried passing `0` to `rad:og`: https://github.com/urbit/urbit/issues/1016#ref-pullrequest-342508381

This edit causes `rad:og` to crash with a `rad-zero` when the sample is `0`.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -3418,6 +3418,7 @@
   |_  a/@
   ++  rad                                               ::  random in range
     |=  b/@  ^-  @
+    ~_  leaf+"rad-zero"
     =+  c=(raw (met 0 b))
     ?:((lth c b) c $(a +(a)))
   ::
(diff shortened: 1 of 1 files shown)


sys/hoon.hoon
```diff
@@ -3418,6 +3418,7 @@
   |_  a/@
   ++  rad                                               ::  random in range
     |=  b/@  ^-  @
+    ~_  leaf+"rad-zero"
```
joemfb: This looks to be missing the assertion that was present in the earlier PR.

(BTW, you can force-push to a branch and arbitrarily change a pull-request. You can even switch to a different target branch. You shouldn't need to open a new PR to replace the contents.)

joshuareagan: Yikes, that was an unfortunate omission.  Fixed now.

PR #777, opened 2018-08-09
Collections redesign

ixv:
Just finished testing it and it seems okay. 

Note that currently the app continues to check for permissions before it allows a post to go through, which clay defaults to rejection. We can set perms from the frontend but we currently do not do so, just a heads up if your wondering why cross-ship posting isn't working.


vvisigoth:
OK (re collections). Is fixing that on front end on your list (just stubbing in open perms for everything)?

PR #778, opened 2018-08-10
fix for travis pill generation issue

ixv:
Old access key stopped working, so here is a new one


joemfb:
@ixv, could you put this on the collections branch too?

PR #784, opened 2018-08-28
updates CI build to compile vere at a specified commit

joemfb:
Also fixes the compilation testing in the `:test` app (note that there are plenty of test failure). 

@eglaysher, @belisarius222, could one of you give some pointers on how the renderer testing could be made to work? The old implementation gave a dummy path, but when I try to duplicate that I get scry failures.


app/test.hoon
```diff
@@ -23,69 +23,90 @@
...
-  :-  -.bem
-  :~  %tabl
-    [`silk`[%bake a fake-fcgi bem] [%$ %cont !>(b)]]
-  ==
+  :-  %build
+  :^    a-rend+/[a]
+      our
+    live=|
+  ^-  schematic:ford
+  :: XX what should the rail be?
```
belisarius222:
```
=/  bem=beam  (need (de-beam %/example))
=/  =rail  [[p q] s]:bem
```

belisarius222:
I left a comment suggesting how you can get a `+rail` from a `+beam`. The rest of the arguments to the `%bake` build look like they should work.

app/test.hoon
joemfb: That makes sense, but I get the same errors regardless (the only difference being `%/app/test/hoon/example/hoon` in place of `%/example/hoon`):

```
> :test [%renders /]
[%findining-hoons under=/ren]
> [/ren/collections %skipped "temporarily disabled"]
> [/ren/css %skipped "not meant to be called outside /web/pack"]
> [/ren/js %skipped "not meant to be called outside /web/pack"]
/ren/rss-xml
> [/ren/run %skipped "not meant to be called except on a (different) hoon file"]
/ren/test-gen
/ren/tree/index
> [/ren/urb %skipped "temporarily disabled"]
> [/ren/x-collections-json %skipped "temporarily disabled"]
> [/ren/x-collections-snip %skipped "temporarily disabled"]
> [/ren/x-htm %skipped "temporarily disabled"]
> [/ren/x-urb %skipped "temporarily disabled"]
>> [%ren %rss-xml]
>> [%ren %test-gen]
{%alts: all options failed}
as-mark
{/! failed:   ford: %hood failed for /~zod/home/0/app/test/hoon/example/hoon:  scry failed for  %cx /~zod/home/~2018.8.28..07.53.16..ba0a/app/test/hoon/example/hoon}
as-renderer
ford: %bake %test-gen on /~zod/home/0/app/test/hoon/example failed:
>> [%ren %tree-index]
{%alts: all options failed}
as-mark
{/tree-elem/ failed:   ford: %bake %tree-elem on /~zod/home/0/app/test/hoon/example failed:  %alts: all options failed}
as-renderer
ford: %bake %tree-index on /~zod/home/0/app/test/hoon/example failed:
%rens-tested
```

I wonder if the `%bake` semantics have changed in case of a scry failure, or if this is a more general change across ford. In any case, given the number of renderers that are already excluded from testing, it might be fine to just disable them all until a more robust approach can be developed for them.

belisarius222: Oh, I think what happened here is that this tried to run the test renderer at ren/test-gen.hoon. That renderer tries to compile .hoon file full of tests at its path to render, but the example file is a .umd file, so when the test-gen renderer tries to render the example path, it doesn't find a hoon file there.

We don't need to test the test renderer. Quis probabit ipsos probatores?

PR #785, opened 2018-08-28
fixy uno~by, other vor-based code duplication

joemfb:
Capturing https://github.com/urbit/arvo/commit/d5d8715b3c9325cdb89ea3f95cec9a1be4b62d75 here for review -- I didn't want it to be lost in a much larger merge.

I don't have any particular insight into these changes, but it's noteworthy that they appear to make `+uno:by` type-safe in addition to deduplicating the hash comparisons.

/cc @ohAitch

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -1427,17 +1427,12 @@
         a
       ?~  a
         b
-      ?:  (vor n.a n.b)
-        ?:  =(n.b n.a)
-          [n.b $(a l.a, b l.b) $(a r.a, b r.b)]
-        ?:  (hor n.b n.a)
-          $(a [n.a $(a l.a, b [n.b l.b ~]) r.a], b r.b)
-        $(a [n.a l.a $(a r.a, b [n.b ~ r.b])], b l.b)
-      ?:  =(n.a n.b)
-        [n.b $(b l.b, a l.a) $(b r.b, a r.a)]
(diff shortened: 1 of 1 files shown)


belisarius222:
I move to reject this PR until it has unit tests. We're not ready to require that with every PR, but bugs with this code could hide out for a long time.

ohAitch:
I claim that the refactoring is mechanically straightforward, so any bugs the unit tests catch are already present on the network ^-^

ohAitch:
(except for the type checking, which should maybe be a separate one-line commit, but at any rate seems unlikely to make things *worse* esp at runtime)

cgyarvin:
This doesn't currently compile according to CI.  What problem were we fixing here?  Just code duplication?

joemfb:
There's also the `+uno:by` type-safety, which should probably be separated. These commits where on the `research-ames` branch from which the ethereum changes were originally staged. I didn't want them to sneak in unreviewed ...

ohAitch:
Interesting. Has mar/plan been touched at all recently, or is this
reflective of general compiler changes? This definitely compiled at some
point P:

joemfb:
Closing this in favor of #972, which is just the added type-checking. I like this refactoring, but it's a low priority and can be revived when there's time for a general cleanup pass over `+in` and `+by`.

PR #789, opened 2018-08-28
cleans up %ames and %jael in anticipation of greater ethereum integration

joemfb:
- improves debug logging for boot errors
- removes `+crua`
- simplifies `+acru`
- updates %jael to use `+pass` and accept incremental events from the contract


cgyarvin:
Beast mode

PR #790, opened 2018-08-28
a clay test passes

belisarius222:


tests/sys/vane/clay.hoon
```diff
@@ -0,0 +1,208 @@
...
+          ^-  tang
+          %+  welp
+            (expect-eq !>([/inserting/~nul/home/~1111.1.1 wire]))
+          ::
+          =/  note=note:clay-gate  q.card
+          ::
+          ?.  ?=([%f %build *] note)
+            [%leaf "bad move, not a %build: {<move>}"]~
+          ::
+          ^-  tang
```
eglaysher: The only real thing that jumps out is the repeated `^-  tang` lines, which seem superfluous.

PR #791, opened 2018-08-29
Redo +expect-eq and consolidate tests

eglaysher:
This patchset changes how `+expect-eq` works so that it takes two vases (instead of taking one vase and then manually splitting the vase apart) and then adds nesting checks to ensure that the expected value nests in the actual value--many tests were already doing this manually and this significantly shortened the ford unit tests.

This also consolidates all existing tests in the system. It merges Joe's patchset that moves the acme tests into their own tests files, moves a remaining generator style test into the formal testing system, and makes sure that the `%/tests` hierarchy mirrors the rest of the filesystem.


app/acme.hoon
```diff
@@ -1,103 +1,10 @@
-/+  tester
+/-  asn1
+/+  base64, der, primitive-rsa, *pkcs, *jose, tester
 =,  eyre
-::
-::::  %zuse additions
+=*  rsa  primitive-rsa
+::  XX move these to %zuse and dedupe libs and tests
```
belisarius222: This looks done?

lib/tester.hoon
```diff
@@ -70,17 +70,30 @@
...
+    =|  result=tang
+    ::
+    =?  result  !=(q.expected q.actual)
+      %+  weld  result
+      ^-  tang
+      :~  [%palm [": " ~ ~ ~] [leaf+"expected" (sell expected) ~]]
+          [%palm [": " ~ ~ ~] [leaf+"actual" (sell actual) ~]]
+      ==
+    ::
+    =?  result  !(~(nest ut p.actual) | p.expected)
```
belisarius222: This checks whether expected nests in actual. Don't we want to check the other way? Expectations should generally be broader than specific results.

belisarius222: It does depend on how we expect people to use this. If users all just write `!>(expected)`, then the type of the expected vase will generally be pretty specific, and this check makes sense. It's oddly contravariant, though.

app/acme.hoon
eglaysher: Removed TODO.

PR #792, opened 2018-08-29
complete +expect-schematic in ford test lib

belisarius222:
Fills out all the other tags of `+schematic`. Not all of them are tested, but this library compiles and a few of the schematics are used so far in /tests/sys/vane/clay/hoon

### DIFF
--- a/lib/test/ford.hoon
+++ b/lib/test/ford.hoon
@@ -84,21 +84,30 @@
   ::
       %bake
     ::
+    %+  weld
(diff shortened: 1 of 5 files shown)


eglaysher:
FYI, this is showing up as conflicting with release-candidate. You probably want to resolve these conflicts so the CI can run.

belisarius222:
Closing in favor of an upcoming non-conflicting PR.

PR #793, opened 2018-08-29
+expect-schematic in ford test lib

belisarius222:
Wrote an `+expect-schematic` function in /lib/test/ford/hoon and used it to simplify the Clay tests.

### DIFF
--- a/tests/sys/vane/clay.hoon
+++ b/tests/sys/vane/clay.hoon
@@ -84,7 +84,7 @@
             (expect-eq !>(%.n) !>(live.note))
           ::
           %-  expect-schematic:test-ford
-          :_  [-:!>(*schematic:ford) schematic.note]
+          :_  !>(schematic.note)
           ^-  schematic:ford
           :+  %pin  ~1111.1.1
           :-  %list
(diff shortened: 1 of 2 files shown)


tests/sys/vane/clay.hoon
```diff
@@ -83,67 +83,19 @@
...
-          =/  cast2=schematic:ford  +.s2
-          ::
-          ?.  ?=([%cast [%~nul %home] %noun *] cast2)
-            [%leaf "bad move, wrong cast2: {<move>}"]~
-          ::
-          =/  lit2=schematic:ford  input.cast2
-          ::
-          (expect-eq !>([%$ %noun %noun 'file2']) !>(lit2))
+          %-  expect-schematic:test-ford
+          :_  [-:!>(*schematic:ford) schematic.note]
```
joemfb: Is there a reason that this needs to not be `!>(schematic.note)`? Not a style critique, just curious if there's type wizardry happening.

belisarius222: Oh, maybe not. I had tried that to fix an error, but it might not be necessary. I'll see if I can pull it out.

PR #803, opened 2018-09-04
Ethereum transaction signing

Fang-:
This adds a function into zuse's `++ethereum` arm for signing raw transactions with a private key.

Also implements supporting logic (`++rlp` encoding), an RPC call for sending such a signed transaction to a node, cleans up existing Ethereum code a little bit, and removed deprecated files.


sys/zuse.hoon
```diff
@@ -4188,15 +4201,15 @@
...
+        :*  v=(mix (end 0 1 y) ?:(big-s 1 0))
             r=r
             s=?.(big-s s (sub ^n s))
         ==
       ::
       ++  ecdsa-raw-recover                             ::  get pubkey from sig
         |=  [has=@uvI sig=[v=@ r=@ s=@]]
         ^-  pont
-        ?>  ?&((lte 27 v.sig) (lte v.sig 34))
+        ?>  (lte v.sig 7)
```
Fang-: @belisarius222 seems sane, right?
Do you remember why this checks for such a large range, even though line 4204 means `v` can only ever be `0` or `1` (previously `27` or `28`)? It's not like the logic ever uses anything but the last bit from `v` anyway.

belisarius222: I don't recall.

sys/zuse.hoon
```diff
@@ -7185,6 +7198,77 @@
   =,  mimes:html
   =,  ethe
   |%
+  ++  sign-transaction
```
eglaysher: Are there test vectors for signature transactions that could be added to the test suite?

Fang-: I haven't come across any official test data, but I'll definitely add in tests against results provided by web3.js

Fang-:
I put in tests for both the RLP encoding by itself, and the transaction signing as a whole. Couldn't find official test cases for the latter, so I threw some data I had on hand into web3.js and used that as test case and result.

eglaysher:
Thank you for adding tests!

PR #804, opened 2018-09-04
Be explicit in having hoon file suffixes in security driver paths.

eglaysher:
`%core` schematics now require a real path, instead of assuming a `hoon` suffix.

### DIFF
--- a/sys/vane/eyre.hoon
+++ b/sys/vane/eyre.hoon
@@ -1967,7 +1967,7 @@
       ^-  schematic:ford
       :+  %mute
         ^-  schematic:ford
-        [%core [[our %home] (flop %_(dom . sec+dom))]]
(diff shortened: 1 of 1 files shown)


eglaysher:
With this fix, using a security driver will now dump this to the console:

```
eyre: execute [%se p=%filter-request q=[~. <|com googleapis|>]]
ford: %call execution failed:
%oauth-no-keys
Run |init-oauth2 /com/googleapis
If necessary, obtain client keys configured for a redirect_uri of
  http://localhost:8080/~/ac/googleapis.com/_state/in
```

Instead of failing silently.

ohAitch:
Presumably this API change has already been reflected in gall and :test,
and nothing else uses %core?

eglaysher:
Yes, this was the last place that wasn't updated.

eglaysher:
Just remembered Joe is out today, redirecting to Ted.

PR #807, opened 2018-09-05
Enqueue second Clay write during an Arvo event

belisarius222:
All but the first write request during an Arvo event gets enqueued, to be run in a later event triggered by a Behn timer. There's a test on this behavior, which passes.

This doesn't entirely fix referential transparency for date-base read requests, but it should at least make it less likely for referential transparency to be violated. Fixing this problem in its entirety is actually a bit tricky, and we will need to think it through in more detail. This is just a quick patch.

The failing case is that someone does a read request at the current date, then does a write request in the same Arvo event, then performs the same read request a second time, which now has a different answer.

I still think it's worth adding this to see if we can unblock the collections integration, even if it doesn't get us all the way to correctness. We should not use this solution indefinitely, though.

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -187,7 +187,7 @@
 ::      location).
 ::  --  `hez` is the unix duct that %ergo's should be sent to.
 ::  --  `cez` is a collection of named permission groups.
-::  --  `cue` is a queue of request to perform in later events.
+::  --  `cue` is a queue of requests to perform in later events.
 ::  --  `tip` is the date of the last write; if now, enqueue incoming requests.
 ::
 ++  raft                                                ::  filesystem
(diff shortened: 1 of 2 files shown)


sys/vane/clay.hoon
```diff
@@ -195,6 +195,8 @@
       mon/(map term beam)                               ::  mount points
       hez/(unit duct)                                   ::  sync duct
       cez/(map @ta crew)                                ::  permission groups
+      cue/(qeu [duct task:able])                        ::  queued requests
+      tip/@da                                           ::  last write date
```
Fang-: These should maybe get comments above along with the rest of the `raft`. There's room up there for a short explanation of what this gets used for/solves.

belisarius222: Ok, I'll add some more docs and then merge this.

PR #817, opened 2018-09-13
Fix +grad case in +make-join

eglaysher:
This fixes a bug in +make-join where following the +grad arm will cause a build to add itself as a dependency. This adds a test to ensure this doesn't regress.

This also adds several assertion that ensure that a client is not in the list of its sub builds, which were useful in tracking down this bug.

review: @belisarius222 
cc: @ixv


belisarius222:
LGTM. Note that we also memoized `+build-to-tape` in this PR.

Issue #819, opened 2018-09-20
[RC] ford error when trying to sync a clay desk


ixv:
stack trace here:
```
[%bad-softing 'crud']
[%bad-softing 'warn']
crude: all delivery failed!
  ####
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.231 3].[6.326 5]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.234 3].[6.326 5]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.237 3].[6.326 5]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.240 3].[6.326 5]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.243 3].[6.326 5]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.249 3].[6.326 5]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.249 7].[6.259 24]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.251 7].[6.259 24]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.252 9].[6.255 11]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.253 26].[6.253 39]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.263 7].[6.290 54]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.265 7].[6.290 54]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.266 7].[6.290 54]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.267 7].[6.290 54]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.269 7].[6.290 54]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.277 7].[6.290 54]>
  /~zod/home/~2018.9.18..17.46.35..0b43/sys/vane/ford:<[6.281 7].[6.290 54]>
[...]

PR #820, opened 2018-09-24
fix bug in +move-root-to-cache

belisarius222:
This should fix the bug encountered by @ixv on the testnet where Ford crashed due to a missing sub-build during `+copy-build-tree-as-provisional`.

The bug was triggered by `+move-root-to-cache` performing a double mutation of a sub-build in the case of a triangle build, where build A depends on B and C, and build B also depends on build C. The fix was to add a check to no-op if we'd already mutated that sub-build.

This PR also contains a minor optimization that removes some vestigial code from `+remove-single-build`, in addition to a cosmetic cleanup in `+add-client-to-sub`.

Closes https://github.com/urbit/arvo/issues/819.

### DIFF
--- a/sys/vane/ford.hoon
+++ b/sys/vane/ford.hoon
@@ -1561,6 +1561,11 @@
       |=  =build-status
       %_    build-status
           clients
+        ::  if we've already encountered :i.subs, don't overwrite
+        ::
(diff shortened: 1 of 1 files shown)


vvisigoth:
mrg it boyz
—
~poldec-tonteg
http://urbit.org

PR #821, opened 2018-09-24
Fix for CI vere usage error

joemfb:
I don't know how this would've worked before, and yet it did. I've added some printfs that should make it clear if argument handling in the livescript vere wrapper somehow breaks in the future.

We should probably replace the livescript test runner with something a little more familiar. Or task someone (not me) with learning livescript.


joemfb:
Also runs the test framework before the app.

ohAitch:
Ideally, urbit itself!

In lieu of fixing urbit "daemon mode", probably the closest more-familiar
language is going to be tcl-based `expect`: iirc the main reason I pulled
in node was the pty.js library, which was able to provide the
requisite terminal faking in a way I was not able to get working in pure
bash.

joemfb:
I can't imagine how to do something like this in bash. Node seems like a good fit, I just don't want to learn compile-2-js languages ... ;)

The more of this we can move into urbit, the better. Up to a point, anyway -- I expect we'll always want some kind of external orchestration, given that we're developing the runtime and Arvo at the same time.

ohAitch:
There might be anything interesting in .travis/ git history; agree that
even if I checked it in it wasn't pretty!

In the long term, esp with the urbit-noun-worker vs the
urbit-event-IO-translator process separation, I imagine it will be quite
doable; though probably involving downloading the latest stable release as
a host environment, if we want any detailed tests on revisions that do not
straightforwardly boot.

joemfb:
Vere arguments: before https://travis-ci.org/urbit/arvo/builds/432696187#L1171 and after https://travis-ci.org/urbit/arvo/builds/432705702#L1055. The presence of `'./in-urbit.ls'` was the problem. Thanks!

PR #823, opened 2018-09-28
nock 12 (4K)

joemfb:
This PR is the dual of urbit/urbit#1027, notably integrating @frodwith's addition of the nock 12 instruction. Additionally, this resolves battery duplication in `+pfix`/`+sfix` and adds some missing %fast hints.


sys/hoon.hoon
```diff
@@ -6409,6 +6409,7 @@
               {$9 p/@ q/nock}                           ::  select arm and fire
               {$10 p/$@(@ {p/@ q/nock}) q/nock}         ::  hint
               {$11 p/nock q/nock}                       ::  grab data from sky
+              {$12 p/{p/@ q/nock} q/nock}               ::  edit
```
eglaysher: I thought we intended to switch $12 and $11, so that edit was $11 while scrying became $12; otherwise there's a hole in the nock spec proper since scrying isn't part of nock.

cgyarvin:
We do *intend* to do that.  We have not, however, *done* it :-)

cgyarvin:
I should point out that modifying nock itself is a considerable challenge
although I guess you guys may be up to it.  I was going to do this.  You
can give it a try of course.  I won't be surprised if you come back howling
like toddlers though.

cgyarvin:
FWIW, I think there's enough that's endearingly whimsical about Urbit.
But then I would wouldn't I.

I think having edit come after hint is baggage enough.  Of course, why
not change that while you're at it!  That would force us to go to Nock
3, considering the shirts we just sent off.  But what the hell, live
dangerously.  (Also having a Nock 4 shirt would be like owning a
Revenge of the Jedi poster.)

joemfb:
closed in favor of #856

PR #824, opened 2018-09-28
Add @q, phonetic data representation

Fang-:
It's `@p`, except never scrambles, doesn't care about even number of bytes, and doesn't use `--` in large values.  
This is needed for tickets, and useful for anything else that may want to use the `@p` look-and-feel to display data.

Syntax for literals is `.~`. A `.~` by itself is invalid, you need at least one syllable. They follow byte ordering as in the `@ux` case, so:

```hoon
@ux       @q
0x1       .~nec
0x201     .~binnec
0x3.0201  .~wes-binnec
```

Note that leading zeroes are valid, but won't change the value. `.~dozzod-dozzod` is equivalent to `.~zod`.

I have included some tests! Pretty sure this is how you test auras: just using them.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -5293,6 +5293,7 @@
 ++  dem  (bass 10 (most gon dit))                       ::  decimal to atom
 ++  dit  (cook |=(a/@ (sub a '0')) (shim '0' '9'))      ::  decimal digit
 ++  dog  ;~(plug dot gay)                               ::  .  number separator
+++  dof  ;~(plug hep gay)                               ::  - @q separator
 ++  doh  ;~(plug ;~(plug hep hep) gay)                  ::  --  phon separator
 ++  dun  (cold ~ ;~(plug hep hep))                      ::  -- (stop) to ~
 ++  duz  (cold ~ ;~(plug tis tis))                      ::  == (stet) to ~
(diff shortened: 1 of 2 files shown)


ohAitch:
What is the use case for diverging on scrambling, saving on a syllable of zero padding for odd numbers of bytes(>1), non-roundtripping "valid" inputs, _or_ removing the long value visual chunking? Can't tickets have the `@p` look and feel, by being literally `@p`?

Fang-:
They could, but it wouldn't be right. `@p` is strictly about ship names, and as far as I'm concerned all existing usage of it that isn't literally about representing/storing *ship names* is incorrect. From a data representation point of view, having `~palfun-foslup` and `~palfun-fosnup` differ in more than just a single byte is super weird.

I'm not yet 100% set on the removal of visual chunking for longer values by the way, waiting to hear @cgyarvin's thoughts.

ohAitch:
I don't think that matches up with historical usage? Ship names are `@p` yes, but so are like all currently issued tickets, various hashes that get printed throughout the system, eyre one-time codes, etc.
`@p` was very much designed to be a generic phonetic representation, with address space merely as the driving central use case

Fang-:
Most of the things you listed are going away in the near term anyway. (^:  
History was maybe wrong here. @cgyarvin and I agree that `@q` is more correct for such uses moving forward.

vvisigoth:
"history was wrong"
-- fang
—
~poldec-tonteg
http://urbit.org

Fang-:
@cgyarvin ping!

mikolajpp:
Also pinging. The new parser wants to know whether or not we are supporting this ;-)

cgyarvin:
Yes, we must support!

Fang-:
fyi, @cgyarvin, I will just like merge this sometime during the coming week and assume this is as clean and correct as it gets.

PR #825, opened 2018-09-28
Update keygen

Fang-:
...to match the latest keygen-js structure and logic, as per the revised spec. This includes delegate -> voting, having those and management proxies per ship, rather than per owner, and restructuring of the wallets to include an Ethereum address.

This doesn't yet include sharding functionality, which I believe @jtobin offered to implement? (Please confirm/deny.) This is not ready to be merged until that is in.

The tests here are largely copied over from keygen-js, and test the core operations. Type safety gives us plenty guarantees about the structure of the full wallet output, though it wouldn't hurt to eventually add tests for the contents of that as well.

This makes implicit use of #824, but works fine without it.

Unrelated, but this also removes an old, outdated version of the BIP32 library that somehow never got properly removed, or snuck its way back in at some point.


Fang-:
Uh yeah the test that uses argon2 is gonna be way too slow if argon2 is not jetted. I... thought we had jets for that in already in latest rc vere, but apparently we don't? They're on the crypto-additions branch, which iirc is still waiting on urbit/urbit#1020 to be merged before merging into release-candidate.

I'll go ahead and make a PR for crypto-additions to be merged into release-candidate as-is, argon2 is by far the most important jet to have, and this PR can't pass without it.

lib/keygen.hoon
```diff
@@ -29,44 +29,55 @@
...
+  ^-  vault
+  =/  owner-seed=byts
+    seed-size^(argon2u ticket seed-size)
   =+  dr=~(. sd pass)
-  =+  cn=|=([s=byts m=meta] (child-node-from-seed s m pass))
+  =/  cn
+    |=  m=meta
+    (child-node-from-seed owner-seed m pass)
+  ::
+  :-  `@q`dat.ticket
```
jtobin: Oh yeah, I'll add this feature to ob-js tomorrow.

Fang-: This really should just be that simplified implementation you did with just the star names (though this also supports uneven byte counts, ie `~fun-witbud`).  
ob-js may or may not be the right place for it, but I think it's the best place we have for JS utils relating to urbit name-like types?

Fang-: (See also #824 I guess.)

Fang-:
Superseded by #877.

Issue #826, opened 2018-10-01
[RC] crude: punted upon copying a large number of files into a pier


ixv:
@belisarius222 @eglaysher 
I've got another error that looks very similar to https://github.com/urbit/arvo/issues/819 but slightly different.

I did the following things
- I copied around 600 files into a pier
- I see my ship succeed in syncing them
- I shut down my ship
- on restarting it I get this stack trace:
https://pastebin.com/fVMXsm5Q

subsequent restarts produces the same stack trace.

I can give you a ship that reproduces this.

belisarius222:
Hm. The stack trace has a bunch of mentions of line 1671 in Ford. The most
recent release-candidate has this as that line:
https://github.com/urbit/arvo/blob/release-candidate/sys/vane/ford.hoon#L1671

Does that match the code that you're running? Seems like a strange place to
die.


—
~rovnys-ricfer
https://urbit.org

ixv:
closed by https://github.com/urbit/arvo/pull/827

PR #827, opened 2018-10-01
removed zapcol from ford

ixv:
lack of tail-call optimization in one of the functions was blowing the stack, causing issue: https://github.com/urbit/arvo/issues/826

removing the !: at the top of the file fixed this.


ohAitch:
3:

Putting tactful !.s on the actual inner loops would go a lot better here,
assuming you want any future bug reports to still have line numbers of any
kind.

PR #828, opened 2018-10-02
restores previously-crashing =, in &hall-action

### LINKED ISSUE #440: =, tiscom causes bail: fail
I've been seeing weird behavior when using `=,`. This appears to be a compile-time one, it occurs even when the compilation should nest-fail.

See [this](https://github.com/urbit/arvo/blob/master/mar/hall/action.hoon#L13) file for my workaround. Just uncommenting the `=,` on lines 13 or 41 causes the issue to appear, printing the following whenever you attempt to use the mark:

```
> &talk-action [%convey ~]
bail: fail
bailing out
```

joemfb:
closes #440

### DIFF
--- a/mar/hall/action.hoon
+++ b/mar/hall/action.hoon
@@ -10,7 +10,7 @@
   |%
   ++  noun  action:hall                                 :::  from %noun
   ++  json                                              :::  from %json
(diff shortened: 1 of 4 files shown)


joemfb:
Fixed this in the rest of the hall marks, too.

PR #830, opened 2018-10-02
changed json -> txt conversion to work with new mark conversion

ixv:
Some of the changes to mark conversion in ford broke some conversions, this PR just fixes one problem I encountered.

### DIFF
--- a/mar/json.hoon
+++ b/mar/json.hoon
@@ -12,8 +12,8 @@
 ::
 ++  grow                                                ::  convert to
   |%
-  ++  mime  [/application/json (as-octs:mimes txt)]     ::  convert to %mime
-  ++  txt   (crip (en-json jon))
+  ++  mime  [/application/json (as-octs:mimes -:txt)]   ::  convert to %mime
+  ++  txt   [(crip (en-json jon))]~
   --
 ++  grab
   |%                                                    ::  convert from
--- a/mar/txt.hoon
+++ b/mar/txt.hoon
@@ -13,7 +13,6 @@
   |%
   ++  mime  |=({p/mite:eyre q/octs:eyre} (to-wain q.q))
   ++  noun  wain                                        ::  clam from %noun
-  ++  json  (corl to-wain so:dejs:format)
   --
 ++  grow
   =>  v=.


belisarius222:
Can you explain what this broke? I'm not sure I understand what Ford-turbo does differently that relates to these changes.

ixv:
Actually I think you're right, I'm not sure why this was broken, or rather I'm not sure why `(corl to-wain so:dejs:format)` ever worked for converting json to txt in the first place.

ohAitch:
It's not encoding *arbitrary* json as text, that's &mime. This is a partial
function that accepts only json string values, and gets you the (list of
lines)text value of those.

belisarius222:
Hmm, well the mark engine tries to run `+grab` first, then tries to run `+grow`, but I think if `+grab` errors out, the whole build errors out. This means that a `+grab` written as a partial function will break the system. I'm not sure we need to support that feature ...

belisarius222:
(the 'grab as partial function' feature, I mean)

belisarius222:
I've always thought of `+grab` as the thing you write when you know this mark will need to convert data *from* a preexisting mark, and `+grow` as the thing you write when you know your new mark will need to convert *to* a preexisting mark.

It is certainly not obvious to me that partial functions are a good idea in the case of the 'talk-command' example. It seems odd to me that we'd have a well-typed data structure that would fail mark conversion...

Do we ever need both a grab and a grow? Maybe it would make more sense to error out if both are defined. That way no one will be surprised when they write a grow arm for their mark, but the result ends up coming out different from what they expected since what actually ran was the grab arm from the target mark.

I'm going to approve this as-is, since it fixes the problem at hand, and we should keep this discussion in mind in case we start to see a pattern of similar issues.

ohAitch:
Agree that this fixes the problem at hand, given the problem is that there
is both a grab and grow :P

(and that it would be helpful to, if not outright forbid that, at least add
a test suite that fails any pair of marks that go both ways, with some
whitelist for cases that's been determined to be useful somehow)

It is my understanding that the purpose of grab is to *filter out* data
that is not well typed. Talk-command is a bit complicated as an example,
but suppose you have a :pong app, with a +=pong-msg type of $%([%ping @t]
[%pong @t]), serialized for the web as e.g. {"ping":"yo"} and
{"pong":"whattup"} respectively. By "partial function", I mean that if the
frontend sends json containing [420] (or in hoon terms, [%a ~[%n '420']]),
there is not a well-typed &pong-msg this can be converted to, and the
conversion should in fact fail. (For a more controversial case, see: tab
characters in md files received via UNIX sync; or bad UTF8 for that matter)

PR #844, opened 2018-10-05
adds pre-boot validation

joemfb:
This PR adds the `|dawn` core, to be used by vere to generate requests, parse responses, and validate what will become the boot event.

I've opened this PR now because I've already fought some regressions and incompatibilities, and would like this test-suite to be enabled going forward. More to come soon.

### DIFF
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -362,7 +362,7 @@
     ++  contracts
       |%
       ++  ships
-        0x6c22.3e68.fefe.b0d2.aaa1.3d5a.3b12.f50f.8b6b.498a
+        0x84b3.7fbc.6188.da8a.e866.1eae.322a.f4d9.2db4.5ecc
       --
     ::
     ::  hashes of ship event signatures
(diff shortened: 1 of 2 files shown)


sys/zuse.hoon
```diff
@@ -7942,4 +7936,148 @@
       --
     --
   --
+:: |dawn: pre-boot request/response de/serialization and validation
```
Fang-: Is `|something` new comment syntax for a core arm? Are single spaces for comments considered good style these days? How far have we fallen? (Also still strongly not a fan of the whitespace between the comment and the thing, combined with no whitespace between the previous thing and the comment, but maybe that's a losing battle.)

sys/zuse.hoon
```diff
@@ -7942,4 +7936,148 @@
...
+      ::
+      +$  live  (unit [=life breach=?])
+      :: +seed: private boot parameters
+      ::
+      +$  seed  [who=ship lyf=life key=ring sig=(unit oath:pki:jael)]
+      --
+  |%
+  :: |give:dawn: produce requests for pre-boot validation
+  ::
+  ++  give
```
Fang-: Maybe `=, ethereum, html, mimes:html` in this?

sys/zuse.hoon
```diff
@@ -7942,4 +7936,148 @@
...
+      |=(gal=@ [`(scot %ud gal) tract ['getKeys(uint32)' [%uint gal]~]])
+    :: +hull:give:dawn: Eth RPC for ship's contract state
+    ::
+    ++  hull
+      |=  who=ship
+      ^-  octs
+      %-  as-octt:mimes:html
+      %-  en-json:html
+      (read-request:ethereum [~ tract ['ships(uint32)' [%uint `@`who]~]])
+    :: +turf:give:dawn: Eth RPC for network domains
```
Fang-: Comment style consistency please! (:

sys/zuse.hoon
```diff
@@ -7942,4 +7936,148 @@
...
+    =,  dejs:format
+    |%
+    :: +czar:take:dawn: parse galaxy table
+    ::
+    ++  czar
+      |=  rep=octs
+      ^-  (map ship [=life =pass])
+      =/  jon=json  (need (de-json:html q.rep))
+      =/  res=(list [@t @t])
+        ((ar (ot id+so result+so ~)) jon)
```
Fang-: You might want to look at `/mar/json/rpc` and pull things out into the stdlib as appropriate.

sys/zuse.hoon
```diff
@@ -7942,4 +7936,148 @@
...
+      %-  as-octt:mimes:html
+      %-  en-json:html
+      %-  batch-read-request:ethereum
+      %+  turn  (gulf 0 2)
+      |=(idx=@ [`(scot %ud idx) tract ['dnsDomains(uint256)' [%uint idx]~]])
+    --
+  :: |take:dawn: parse responses for pre-boot validation
+  ::
+  ++  take
+    =,  dejs:format
```
Fang-: Probably also `=, constitution:ethereum`.

sys/zuse.hoon
```diff
@@ -7942,4 +7936,148 @@
...
+  :: +veri:dawn: validate keys, life, discontinuity, &c
+  ::
+  ++  veri
+    |=  [=seed =hull:constitution:ethe =live]
+    ^-  $%  [%& ^seed (unit ship)]
+            [%| rank:ames @tas]
+        ==
+    =/  rac  (clan:title who.seed)
+    =/  cub  (nol:nu:crub:crypto key.seed)
+    ?-  rac
```
Fang-: This could do with a brief comment per case explaining how the validation for that case works.

sys/zuse.hoon
```diff
@@ -7942,4 +7936,148 @@
...
+        =/  loy  (com:nu:crub:crypto pass.u.net.hull)
+      =/  hax  (shaf %earl (sham lyf.seed pub:ex:cub))
+      ?.  =((some hax) (sure:as:loy u.sig.seed))
+        [%| rac %invalid-sig]
+      :: XX revisit for rekey
+      ?^  live
+        [%| rac %already-booted]
+      [%& seed ~]
+    ::
+        *
```
Fang-: Isn't this what `?+` is for?

tests/sys/zuse/dawn.hoon
```diff
@@ -0,0 +1,265 @@
...
+    ==
+  %+  expect-eq
+    !>  oct
+    !>  turf:give:dawn
+::
+++  test-take-czar
+  =/  oct
+    %-  as-octs:mimes:html
+    %+  rap  3
+    :~  '[{"id":"0","jsonrpc":"2.0","result":"0xb69b6818b17b7cc22f8e0a2291f58'
```
Fang-: For these big result blobs, maybe consider splitting them as 32 bytes per line, for readability of the test data.

sys/zuse.hoon
joemfb: I believe this is proper space-rated style. @belisarius222, can you confirm?

sys/zuse.hoon
joemfb: Likewise, `|foo` for a library core. `+bar:foo` for a gate in said core (`+bar` vs `+bar:foo` is still subject to debate, I believe).

tests/sys/zuse/dawn.hoon
joemfb: You wouldn't prefer to maximize the available 80 chars? I'm not sure anything will look good here ...

Worth noting, several those blobs contains the Ships contract address, so they'll need to change at least once more.

sys/zuse.hoon
joemfb: Sure, but it'd be a big first-backstep, very unbalanced.

sys/zuse.hoon
Fang-: Here I was referring to the lack of below-comment whiteline specifically, not the `+` vs `|`. I should've been more explicit!

sys/zuse.hoon
joemfb: I'm torn when it comes to adding lots of `=,`. I also think we need a review of the boundaries between the different ethereum-related cores (`|ethereum` vs `|ethe`) in particular).

tests/sys/zuse/dawn.hoon
Fang-: 32 bytes as in 64 characters, and we already use 8 for left-hand padding. I think being able to comfortably read the data is more valuable than 8 "wasted" characters.  
To clarify, I mean to say that doing 32 bytes per line allows you to separate the arguments to be actually readable. Everything but variable-length data is just a 32 byte line.

Ships contract address should maybe be interpolated in from zuse then?

ohAitch: I mean if the goal is to read the data, you probably want to construct a hoon json literal and compare against _that_.

sys/zuse.hoon
joemfb: I'll be happy to comment/review these in detail when I PR code that actually uses them. For now, the error symbols at least label the failure cases.

tests/sys/zuse/dawn.hoon
joemfb: Reformatted.

@ohAitch, this goal here is to completely encapsulate en/de-serialization, so vere only needs to only make very straightforward function calls. I'll tag you in when I open that PR.

sys/zuse.hoon
joemfb: Fixed

sys/zuse.hoon
joemfb: Agreed. I'm also not sure that jael should be making marked requests, as opposed to parsing itself. When we're "done", we should have a larger review of all the eth/jael stuff together, Imo.

sys/zuse.hoon
joemfb: As above, I expect this to need some refactoring.

ohAitch:
That's reasonable, but we can still express json as (en-json {some-hoon})
in the tests; and trust the standard library tests that the "++json <->
json text" transcoder to catch any errors there.

sys/zuse.hoon
belisarius222: Yes, `|arm-name` for arms that define cores that aren't gates. And yes, space-rated style has comment, then blank comment line, then target of the comment, with no blank line above the comment.

There should be two spaces after the colcol, however.

sys/zuse.hoon
Fang-: I generally try to =, any namespace I enter more than like twice, but a rule for this should definitely be part of the style standard.

Strongly agreed about `ethereum` and `ethe`. I split it up like that because types vs functions, which seemed to be a zuse convention, but it *is* a bit unwieldy as-is.

sys/zuse.hoon
Fang-: Yes please! Strong agree on the big review.

sys/zuse.hoon
Fang-: > space-rated style has comment, then blank comment line, then target of the comment, with no blank line above the comment.

What's the reasoning for doing a blank line between the comment and the thing it's about, and/but not between the thing it's not about and the comment? I brought this up when this style first surfaced, I think it's not particularly pleasant to read.

tests/sys/zuse/dawn.hoon
```diff
@@ -0,0 +1,282 @@
...
+    !>  &
+    !>  ?=(^ zar)
+::
+++  test-give-hull
+  =/  oct
+    %-  as-octs:mimes:html
+    %+  rap  3
+    :~  '{"jsonrpc":"2.0","method":"eth_call","params":[{"data":"'
+        '0xb322069400000000000000000000000000000000000000000000000000000000'
+          '00000000'
```
Fang-: For calls like this and the ones in turf, it makes sense to put the function identifier (first 4 bytes) on a separate line from the first argument, so that you still end up with 32-byte lines for the arguments.

Sorry for being a pain about this, and thanks for the changes. The others look super good now!

sys/zuse.hoon
joemfb: @belisarius222, I don't know how I got the single-space stuck in my head. This might be wrong in `:acme` and it's libs too. Two is definitely better.

tests/sys/zuse/dawn.hoon
joemfb: Thanks!

joemfb:
@ohAitch, we certainly can. Any particular reason you think I should here? To me, those gates should produce mostly constant blobs, varying only by the ship name in question. If those change, I want to know.  (I guess JSON serialization defaults could change without invalidating those blobs. But I'd argue that the shouldn't.)

PR #847, opened 2018-10-05
fixes userspace :test compilation failures

### LINKED ISSUE #814: RC branch breaks urb
Testing out the RC branch for urbit and arvo and urb seems to have borked 

```
dkerschner@swampert:~/Work/examples$ /home/dkerschner/build/urb/urb -d '(add 2 2)'
WARNING <module> 363 - No pier or port specified, looking on port 12321
{ ford: %core on /~zod/home/0/mar/lens/command/hoon failed:
  ford: %plan failed:
  ford: %ride failed to compute type:
  ford: %slim failed:
  -find.command
  /~zod/home/0/mar/lens/command/hoon::[5 9].[5 21]>
  /~zod/home/0/mar/lens/command/hoon::[5 5].[5 21]>
  /~zod/home/0/mar/lens/command/hoon::[5 1].[63 3]>
  /~zod/home/0/mar/lens/command/hoon::[4 1].[63 3]>
  /~zod/home/git-b4631e9438/sys/vane/ford:<[4.794 18].[4.794 59]>
}
while compiling %lens-command:
ford: no mark path from %json to %lens-command
ford: %cast %lens-command on [ship=~zod desk=%home] failed:
[%eyre-lens-fail %coup]

dkerschner@swampert:~/Work/examples$
```

### LINKED ISSUE #845: curl generator broken in release-candidate
```
> +curl "http://www.google.com"
/~zod/home/0/gen/curl/hoon::[8 1].[14 20]>
/~zod/home/0/gen/curl/hoon::[9 1].[14 20]>
/~zod/home/0/gen/curl/hoon::[10 1].[14 20]>
/~zod/home/0/gen/curl/hoon::[10 11].[14 20]>
/~zod/home/0/gen/curl/hoon::[11 1].[14 20]>
/~zod/home/0/gen/curl/hoon::[12 1].[14 20]>
nest-fail
ford: %slim failed: 
ford: %ride failed to compute type:
ford: %plan failed: 
ford: %core on /~zod/home/0/gen/curl/hoon failed:
```

joemfb:
- fixes `+curl` generators by fixing `+sole-request` (close #845)
- fixes `mar/lens/command` (fixes #814?)
- fixes the +ivory pill
- fixes +deco (by commenting out core chapters)
- fixes various web hoons

/cc @baudtack


gen/deco.hoon
```diff
@@ -99,7 +99,7 @@
     ::
     ::  arms producing molds are introduced with {+$}.  the
     ::  compiler will copy the arm decoration onto its product
-    +|
+    :: +|
```
joemfb: @cgyarvin, are core chapters temporarily disabled? Slated for removal?

mar/lens/command.hoon
```diff
@@ -2,13 +2,13 @@
 /+    old-zuse
 ::
 =,  old-zuse
-|_  com/command:lens
+|_  com/command:^^lens
```
joemfb: @belisarius222, I found this by trial and error. I can't make heads or tails of it. Any idea what's happening here?

web/pack/css/codemirror-fonts-bootstrap-tree.hoon
```diff
@@ -8,4 +8,4 @@
         /:/===/web/lib/css/bootstrap:/css/
         /:/===/web/tree/main:/css/
   ==
-(rap 3 -.-)
+(rap 3 -)
```
joemfb: @ixv, does this look right? Is this file still used?

web/pack/js/tree-urb.hoon
```diff
@@ -6,4 +6,4 @@
   /.    /:/===/web/tree/main:/js/
         /:/===/web/lib/js/urb:/js/
   ==
-(rap 3 -.-)
+(rap 3 -)
```
joemfb: @ixv, this one too.

mar/lens/command.hoon
ixv: Could be a result of changes to how =, works. Me and Ted noticed we had to do `..^is` in a jet hint because there was an additional copy included by =,

belisarius222: Looks like old-zuse defines `lens`, and then we're skipping over the imported sur/lens, to grab `lens` from the standard library? I'm still reading through this.

belisarius222: There's no `+lens` in zuse or `old-zuse`. There's a `+lens` in hoon.hoon, but it doesn't have a `+command` in it.

I suspect `=,  old-zuse` exposes two copies of the standard library, each of which has the `lens` from hoon.hoon.

A downside of the way organic dairy imports work is that they can expose multiple copies of the standard library, especially if you `=,` them. Arguably, Ford should try to convert imported cores to lead (opaque) cores, although that could break other things, and Hoon files aren't constrained to producing cores, so that functionality would break if a library produces, say, a cell of an atom and a core. Another alternative would be making `=,` of a core convert the core to a lead core, although that would suffer from the same issue. The last option I can see would be a convention to do `=,  ^?(lens)`.

In my own code I generally only do `=,`s inside functions, since each one raises the risk of some nasty namespace issue. Also, when I see a symbol, I like to see where it came from. This is a bit of a cop-out, though.

belisarius222: @eglaysher you might have some ideas here

belisarius222: Oh, there is yet another option, which would be to have a convention to `^?` cores produced by libraries if they're not supposed to expose the standard library. The system would need to be resilient against the inevitable lack of compliance with this convention, though.

web/pack/css/codemirror-fonts-bootstrap-tree.hoon
joemfb: Thanks, removed.

mar/lens/command.hoon
joemfb: Thanks for checking. It's worth noting that `command:lens` doesn't work in this case, and neither does `command:^lens`.  IMO, old-zuse should really be removed on release-candidate. That would certainly make this sort of thing less likely. Do we need a separate issue to track namespace oddities?

belisarius222: Maybe we should add a `^?` to `old-zuse`. Wanna give that a shot?

belisarius222: I'm not opposed to having a catch-all issue for dealing with namespace weirdnesses, if you think that's a good way to go.

belisarius222: Removing old-zuse entirely is not a bad idea. It's tedious, but it would be a good task for someone who's just learning Hoon.

PR #855, opened 2018-10-10
added sunk note to jael

ixv:
Emits a %sunk note to vanes when notified that a ship has lost continuity.


ixv:
Let's merge after.

PR #858, opened 2018-10-12
ethereum-integrated boot and networking

joemfb:
This PR includes a variety of changes and new features to support our ethereum integration.

- pre-boot validation

Already added in #844, refined and commented here.

- new single-homed boot

While we still boot through %dill, the boot event is now routed to %jael instead of %ames. `%dawn` is a validated, live, ethereum-integrated boot - including our private keys, galaxy table, ames domains, and sponsor. `%fake` does what is says on the can. Everything related to the old multi-homed comet-upgrading boot process has been removed.

Private key files for this boot can be generated with `+key` or `+moon` (but note that moons are not yet wired into the boot process).

Arvo is practically single-homed (it will always only have one identity), but formally still multi-homed, as the old Arvo/vane storage structures and interfaces are still present.

- %ames/%jael integration

The old inline PKI in ames has been removed (with allowances for moons and comets), and all PKI updates are sent from %jael to %ames. %jael itself gets these updates by subscribing to its sponsor, or directly polling an ethereum node.

This changes the %ames packet acceptance criteria in some subtle ways, but the gist of the typical path is that %ames drops packets from unfamiliar ships until it receives the appropriate keys from %jael.

- individual continuity breaches

Sunken ships are no more! Or rather, they can be *discontinuously* re-keyed in the Ships contract. When %jael learns of such an occurrence, it sends a `%sunk` event to all (relevant) vanes, which can in turn purge state and subscriptions related to the ship in question. The most important such vane is %ames, which must reset sequence numbers to be able to communicate with a breached ship. The other vanes currently treat this event as a no-op. Thanks @ixv!

There's much more to discuss and document in these changes, and all will be revealed in due time. Work on the remaining features mentioned above is ongoing.

This PR is the dual of urbit/urbit#1037, and should be merged at the same time.


sys/vane/ames.hoon
```diff
@@ -1438,17 +1077,23 @@
...
-              =+  rol=(~(get by ryl.bah) cha)
-              ?~(rol *rill u.rol)
-          =+  sex=sed.rol
+          =/  rol=rill  (fall (~(get by ryl.bah) cha) *rill)
+          =/  sex=@ud  sed.rol
           ::  ~&  [%tx [our her] cha sex]
+          ::  if we don't have a public key for :her,
+          ::  request keys and proceed (skin will be %open)
+          ::
+          ::    XX update state so we only ask once?
```
Fang-: Yes, a "waiting on keys" is nice to have. In the "my sponsor doesn't have keys for her either" case, we just get the keys *eventually*, right?

sys/zuse.hoon
```diff
@@ -95,6 +95,8 @@
   %-  unit                                              ::  ~: unknown
   %-  unit                                              ::  ~ ~: invalid
   (cask vase)                                           ::  marked cargo
+::
+++  turf  (list @t)                                     ::  domain, tld first
```
Fang-: You might actually be overriding the `++turf` in `hoon.hoon` with this?

sys/vane/jael.hoon
```diff
@@ -644,35 +647,103 @@
...
+      ::  set initial domains
+      ::
+      =.  tuf.urb  turf.tac
+      ::
+      =.  moz
+        %+  weld  moz
+        ::  order is crucial!
+        ::
+        ::    %dill must init after %gall (and after %give intended for unix)
+        ::    %jael init must be deferred post %eyre init
```
Fang-: Brief reasoning here might be insightful.

sys/vane/jael.hoon
```diff
@@ -644,35 +647,103 @@
...
+      %^  cute  hen  our.tac  =<  abet
       (~(look et our.tac now.sys etn.lex) src.tac)
     ::
     ::  create promises
     ::    {$mint p/ship q/safe}
     ::
         $mint
-      %+  cure  our.tac
+      ~|  %fake-jael
+      ?<  fak.urb
```
Fang-: Wait, why wouldn't you be able to do non-PKI operations?

sys/vane/jael.hoon
```diff
@@ -852,9 +937,9 @@
   ::  urb: absolute urbit state
   ::  sub: relative urbit state
   ::
-  =*  urb  ->-
-  =*  sub  ->+<
-  =*  etn  ->+>
+  =*  urb  ->+<
+  =*  sub  ->+>-
+  =*  etn  ->+>+
```
Fang-: Every time I see a PR touch this I cringe. This feels pretty stupid, but I guess there's no better way of having both direct access to the contents, and a face for reference of the whole?

sys/vane/jael.hoon
```diff
@@ -1142,11 +1241,25 @@
...
-        ..file(hul.eth (~(put by hul.eth) who hel))
+        ?:  ?=(%& -.new)
+          :-  (file-keys who p.new)
+          ..file(hul.eth (~(put by hul.eth) who hel))
+        ?:  p.new  
+          :-  kyz
+          ..file(hul.eth (~(put by hul.eth) who hel))
+        :-  (file-discontinuity who)
+        %=  ..file
+          hul.eth  (~(put by hul.eth) who hel)
```
Fang-: Probably just `=.  hul.eth` right after the `=-`.

sys/vane/jael.hoon
```diff
@@ -1673,11 +1786,69 @@
...
+      ~
+    =/  who  (slaw %p i.tyl)
+    ?~  who  ~
+    ?:  =(u.who p.why)
+      ``[%atom !>(lyf.own.sub.lex)]
+    =/  pub  (~(get by kyz.puk.sub.lex) u.who)
+    ?~  pub  [~ ~]
+    ``[%atom !>(life.u.pub)]
+  ::
+      %earl
```
Fang-: Any reason we're not deviating from archaic terminology and just saying `moon` instead?

sys/zuse.hoon
```diff
@@ -564,10 +577,9 @@
         olz/(map flap cape)                             ::  packets completed
         wab/(map ship bath)                             ::  relationship
     ==                                                  ::
-  ++  deyd  {p/@ q/step r/?}                            ::  sig stage fake?
+  +$  deed  [=life =pass oath=(unit oath:pki:jael)]     ::  stage/pub/sig
```
Fang-: "stage" for `life` is confusing here, I haven't seen that word used for it anywhere else.

gen/key.hoon
```diff
@@ -0,0 +1,29 @@
...
+?>  =('b' mag)
+=/  bod=@  (rsh 3 1 pub)
+=/  cry=@  (rsh 8 1 bod)
+=/  sgn=@  (end 8 1 bod)
+%+  print  leaf+"  authentication: 0x{(render-hex-bytes:ethereum 32 sgn)}"
+%+  print  leaf+"  networking:     0x{(render-hex-bytes:ethereum 32 cry)}"
+%+  print  leaf+"ethereum public keys:"
+::
+=/  sed=seed:able:jael
+  [who 1 sec:ex:cub ~]
```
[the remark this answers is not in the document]
joemfb: Good catch! This was written before discontinuity was implemented. In the long run, we won't be generating keys this way. But, in the meantime, this should be a named argument with a default of 1 or something.

gen/moon.hoon
```diff
@@ -15,11 +12,18 @@
         ~
         ~
     ==
-=/  ran  (clan p.bec)
+=*  our  p.bec
+=/  ran  (clan:title our)
 ?:  ?=({?($earl $pawn)} ran)
   %-  produce
   :-  %tang  :_  ~
   leaf+"can't create a moon from a {?:(?=($earl ran) "moon" "comet")}"
```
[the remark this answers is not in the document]
joemfb: We have in the past, I don't see why we shouldn't continue. @cgyarvin?

gen/moon.hoon
```diff
@@ -15,11 +12,18 @@
...
-=/  mon  (mix (lsh 5 1 (end 5 1 eny)) p.bec)
-=/  tic  .^(@ /a/(scot %p p.bec)/tick/(scot %da now)/(scot %p mon))
-(produce [%tang leaf+"ticket: {<`@p`tic>}" leaf+"moon: {<`@p`mon>}" ~])
+=/  mon=ship  (mix (lsh 5 1 (end 5 1 eny)) our)
```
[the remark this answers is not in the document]
joemfb: Yeah, good point.

gen/moon.hoon
```diff
@@ -15,11 +12,18 @@
...
+=/  cub=acru:ames  (pit:nu:crub:crypto 512 eny)
```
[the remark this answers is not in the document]
joemfb: "First" or "last" is a matter of perspective, of course. In any case, `eny` *is* 512 bits.

But in general, `+pit:nu:crub:crypto` takes a bitwidth and seed, converts the bitwidth to bytewidth and uses that for a length-adjusted SHA-512 (`+shal`) of the seed. `+shal` in turn truncates its input to the specified bytewidth. So that would be the "first" bytes semantically, from the little end (or "last" visually, with %ux for instance).

sys/vane/jael.hoon
```diff
@@ -1673,11 +1786,69 @@
   ^-  (unit (unit cage))
   :: XX security
```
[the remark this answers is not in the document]
joemfb: I've never implemented before a `+scry` before! The interface itself seems to support things that don't appear accessible through `.^`. So I simply don't know. I kept it as minimal as possible.

sys/vane/jael.hoon
```diff
@@ -1673,11 +1786,69 @@
...
-  %-  some
-  ?.  =(%$ ren)  ~
-  %+  bind  (~(scry of [now eny] lex) syd tyl)
-  =-  ~!  -  -
-  |=(a/gilt [-.a (slot `@`3 !>(a))])
+  :: %-  some
+  ?.  =(%$ ren)  [~ ~]
+  :: =-  ~!  -  -
+  :: %+  bind  (~(scry of [now eny] lex) syd tyl)
+  :: |=(a/gilt [-.a (slot `@`3 !>(a))])
```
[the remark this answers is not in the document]
joemfb: As above. `+scry:of` seems like a pattern that should be followed. %ames also has a distinction between `+perm` and `+temp` that makes some sense. I'd like more guidance on a what a good `+scry` interface looks like for any vane other than `%clay`. @cgyarvin @belisarius222

lib/hood/womb.hoon
```diff
@@ -694,13 +691,6 @@
   ::  (emit /rekey %next sec:ex:(pit:nu:crub 512 (shaz (mix %next (shaz eny)))))
   ~&  %rekey-stub  .
```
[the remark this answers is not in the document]
joemfb: That seems plausible to me, but I don't know for sure. Who could make that call?

sys/vane/ames.hoon
```diff
@@ -423,46 +69,69 @@
...
-          (com:nu:crub:crypto r.q.i.lew.wod.dur)
+          ^-  [lyf=life cub=acru]                       ::  client crypto
+          ?>  ?=(^ lew.wod.dur)
+          [life.u.lew.wod.dur (com:nu:crub:crypto pass.u.lew.wod.dur)]
         ::
         ++  clon
           ^-  life
-          ?~(lew.wod.dur 0 p.p.q.i.lew.wod.dur)
+          ::  if we don't have a +deed for :her, we guess 1
+          ::  XX should we keep guessing 0? is this used for crypto?
```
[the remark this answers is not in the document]
joemfb: The comment can be removed. `+life` used to be 0-indexed, is now 1-indexed.

sys/vane/ames.hoon
```diff
@@ -961,59 +604,40 @@
...
+      ?>  (~(has by vein) life)
+      ?>  =(life (roll ~(tap in ~(key by vein)) max))
+      ::
+      ~|  %vein-stub
+      ?<  (~(has by urb.ton.fox) our)
+      ::
+      =/  can=(list ship)
+        %-  flop
+        |-  ^-  (list ship)
+        ?:((lth our 256) ~ =+(seg=(sein:title our) [seg $(our seg)]))
```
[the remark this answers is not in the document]
joemfb: `seg` was just the convention in the file. I assume "seigniorage" or some cognate.

sys/vane/jael.hoon
joemfb: Abandon reason all ye who touch the boot sequence!

sys/zuse.hoon
joemfb: Indeed, and unfortunately, I am. I got attached to the name (and used it events/effects) before I realized. I think "turf" is a really good flw for domain, and remain attached to it. There's only one use of `^turf` in all of userspace. Maybe we can rename "utf8 to utf32"? And good catch, I meant to call this out explicitly.

sys/vane/jael.hoon
joemfb: We don't want to let fake ships rekey. This is a heavy handed way to do that. It'll blow up if we ever start using %jael for "promises". Which we aren't doing.

sys/vane/jael.hoon
joemfb: It'd probably be possible to achieve the desired affect in clearer ways. This vane predates `=,` (and `+*` in `|_`).

sys/vane/jael.hoon
joemfb: Consistency over clarity.

sys/vane/ames.hoon
```diff
@@ -627,50 +296,62 @@
...
-            =+  yig=sen
-            ::  =+  bil=`wyll`(pare wyl.dur law.saf)    ::  XX not set
-            =+  bil=law.saf                             ::  XX send whole wyll
-            =+  hom=(jam ham)
+            =/  bil=(unit deed)
+              ?.  ?|  ?=(?(%earl %pawn) (clan:title our))
+                      ?&  =(our (sein:title her))
+                          ::  XX is this necessary?
+                          ::
+                          !=(our her)
```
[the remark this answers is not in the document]
joemfb: That context is "if we're not a moon or a comet, we only include our deed if we're using asymmetric crypto with our sponsor". We only use asymmetric crypto until we've established a symmetric key. I could probably add a comment to that effect.

So that additional `!=(our her)` is just trying to avoid sending our deed to ourselves (which we only be possible if we were self-sponsored, ie a galaxy). Not particularly important either way, just needs to be consistent with `+deng`. I'll double check and comment accordingly.

As a side note, it's kind of weird that self communication goes all the way through the whole packet encryption mechanism. I guess it's the simplest way to make sure sequence numbers are maintained, message ordering within channels are preserved, etc.

sys/vane/ames.hoon
joemfb: We don't currently set that state, so we re-request on every packet. But we request on the same duct, so the subscription is never duplicated. At least that used to be the case. With the more recent ames boot changes, the duct is reset on every restart. So this should be fixed.

gen/key.hoon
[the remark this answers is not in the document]
Fang-: Don't you just scry into jael to get the current life? Or am I missing something about what this is supposed to be used for?

gen/moon.hoon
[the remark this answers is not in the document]
Fang-: Pretty sure that means "tl;dr no entropy leakage". Good!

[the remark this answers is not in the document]
Fang-: Actually that tl;dr is wrong, since "`eny` *is* 512 bits" means all of the data gets sent into `shal`. 32 of those bits are known, lowering the amount of security bits we have here.

Assuming that *is* correct, we should first use `eny` to get the `cub`, then `shal` it to get a new `eny` to use for `mon`. Seems sane?

sys/zuse.hoon
Fang-: I'm in favor of renaming the utf conversion. The flw is much more appropriate here, and there's only very little code that uses it.

sys/vane/jael.hoon
Fang-: > It'll blow up if we ever start using %jael for "promises".

My point. Can't we just ban `%jewel`s or whatever and call it a day?

Also, to be sure: we don't want fake ships to rekey because they technically only have one key (generated from shipnum), correct? What if we just incremented the key seed (ie shipnum) for every new life? Maybe too much work to be worth it.

sys/vane/jael.hoon
Fang-: One of these days someone will make a PR for something like "gaxy, star, lant, moon, come". It may or may not be me. We may or may not spend a couple months bikeshedding about various four-letter versions and just using the full names.  
imo the `rank` terminology needs to go sooner rather than later, but the consistency point is fair, and this is not a discussion that needs to happen in/for PR.

ohAitch:
@cgyarvin wrote the spec, I only implemented it; or in this case, very
clearly didn't even do that :P

ohAitch:
Yeah I have hope for something like  =, [urb sub etn]=->+  or maybe
actually putting the names on in the sample and then immediately =, ing
them back off again

ohAitch:
Yup, short of dumping the ~entire network reparenting state~ into zuse
somehow, I don't think you *can* implement the mutable version without .^
So vanes will have to do more explicit acrobatics if they want to care
about non-hierarchical parents.

gen/key.hoon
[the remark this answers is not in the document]
joemfb: This is for generating new keys, which is why it prints the `+pass` as separate authentication and networking keys, formatted for ethereum. I've been using it for local testing. It uses the same passphrase-based derivation that `+pope` used to, but should be probably adjusted to be more similar to the final derivations. But, again, this is just temporary a solution.

sys/vane/ames.hoon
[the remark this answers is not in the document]
joemfb: And you're right, i'll replace it with `(tail (saxo our))`, where `saxo` is a jael scry.

sys/vane/jael.hoon
```diff
@@ -624,11 +626,29 @@
...
+    ^-  ship
+    ::  XX save %dawn sponsor in .own.sub, check there
+    ::
+    =/  hul  (~(get by hul.eth.sub) who)
+    ?:  ?&  ?=(^ hul)
+            ?=(^ net.u.hul)
+            ?=(^ sponsor.u.net.u.hul)
+        ==
+      u.sponsor.u.net.u.hul
+    (^sein:title who)
```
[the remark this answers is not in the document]
Fang-: That is not correct. If the sponsor is unset (but the ship is otherwise active/initialized, ie it has an owner (for our purposes here saying "it has network keys" is also fair, since anything without network keys does not matter to the urbit network)) then that means the ship has no sponsor at all.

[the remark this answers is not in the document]
Fang-: To be clear, "`+hull` as a representation of the on-chain state" is exactly correct, but the on-chain logic allows for a "no sponsor" state.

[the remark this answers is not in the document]
joemfb: If the sponsor has not been changed and is still just the prefix, is that captured on-chain? Or is it implicit in that field being unset?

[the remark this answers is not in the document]
Fang-: The field gets set on ship creation, so all ships start out with the prefix sponsor. If the sponsor changes, the value here changes. If the sponsor kicks, this value becomes `~`.  
There is no implicit state in a `+hull` that has been updated using the `apply-hull-diff` stdlib logic.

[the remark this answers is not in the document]
joemfb: Right, but is that true of the contract?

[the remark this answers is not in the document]
Fang-: That is also true of the contract. No implicit values there. The only "no sponsor set" case that isn't strictly "it got kicked" is for unactivated/unspawned/non-existent ships.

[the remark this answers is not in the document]
joemfb: Ok, thanks. I'll put the initialization back in. And we need to decide what to do with the sein scry in the case of lost sponsorship. I think you had mentioned falling back to the last set sponsor, which will require state structure updates in %jael. We could produce a double unit, but that will crash in userspace and force vanes to handle it.

Is their anything we would want to do in %ames other than continue trying to use the old sponsor for routing?

//cc @cgyarvin

sys/vane/gall.hoon
```diff
@@ -1241,11 +1243,14 @@
         $serv  `%e
         $tend  `%a
         $them  `%e
+        $turf  `%j
+        $vein  `%j
```
joemfb: @philipcmonk, were these two just for testing? I expect the other additions here to be used from userspace, but these two seem less likely. (And sorry I missed this in #870).

philipcmonk: Yeah, you're right, I'll take those out

sys/vane/jael.hoon
```diff
@@ -644,35 +690,150 @@
...
+      =*  our  our.tac
+      ::  sort-of single-homed
+      ::
+      =.  our.own.sub  our
+      ::  fake keys are deterministically derived from the ship
+      ::
+      =/  cub  (pit:nu:crub:crypto 512 our)
+      ::  save our parent signature (only for moons)
+      ::
+      ::    XX move logic to zuse
```
Fang-: If other parts of arvo want to use this, yes, this should go into zuse.

sys/vane/jael.hoon
```diff
@@ -704,33 +865,45 @@
...
+    ::  XX should be a subscription
+    ::  XX reconcile with .dns.eth
+    ::  request domains
+    ::    [%turf ~]
+    ::
+        %turf
+      ::  ships with real keys must have domains,
+      ::  those with fake keys must not
+      ::
+      ?<  =(fak.own.sub ?=(^ tuf.own.sub))
```
Fang-: Wait huh, so this will crash if a fakenet has DNS configured? I get we don't actually want to *use* it, but it shouldn't crash just because there's data there, should it?

sys/vane/jael.hoon
```diff
@@ -1064,11 +1315,36 @@
...
-    (pubs:feel [[who -(live &)] ~ ~])
+    =;  new=public
+      (pubs:feel (my [who new] ~))
+    ::
+    =/  old=(unit public)
+      (~(get by kyz.puk) who)
+    ?:  ?|  ?=(?(%earl %pawn) (clan:title who))
+            ::  XX save %dawn sponsor in .own.sub, check there
+            ::  XX or move sein:of to sein:su?
+            ::  XX full saxo chain?
```
Fang-: Why would a full sponsorship chain need to be seen here, or anywhere really?

sys/vane/jael.hoon
```diff
@@ -1127,34 +1404,56 @@
         [kyz ..file]
       ::
       ::  sanity check, should never fail if we operate correctly
+      ::  XX is this true in the presence of reorgs?
```
Fang-: Pretty sure that with @philipcmonk's recent changes we now account for reorgs? I'd expect the `?>` below to now go a rewind in case a reorg is detected though.

sys/vane/jael.hoon
Fang-: We change it in like every branch here. (^:

sys/vane/jael.hoon
```diff
@@ -1179,6 +1478,41 @@
...
+      |=  wer=event-id
+      ^+  ..file
+      =?    sap
+          %+  lth  2
+          %+  sub.add
+            (div block.wer interval.sap)
+          (div last-block.sap interval.sap)
+        ~&  :*  %snap  count=count.sap  max-count=max-count.sap 
+                last-block=last-block.sap  interval=interval.sap
+                lent=(lent ~(tap to snaps.sap))
```
Fang-: Can we just list these purely vertically, for readability's sake? Thanks!

sys/vane/jael.hoon
```diff
@@ -1563,20 +1971,68 @@
...
+  ::
+  ++  take-catch-up-step
+    |=  [rep=response:rpc:jstd from-block=@ud next-block=@ud]
+    ^+  +>
+    ?<  ?=(%batch -.rep)
+    ?<  ?=(%fail -.rep)
+    ?:  ?=(%error -.rep)
+      ~&  [%catch-up-step-error--retrying message.rep]
+      (catch-up from-block)
+    ::  XX file
```
Fang-: It's not immediately clear what this means, what should be done here.

sys/vane/jael.hoon
```diff
@@ -1595,11 +2051,24 @@
...
-      +>
+      ?.  removed.u.mined.log
+        ::  ~&  [%ignoring-duplicate-event tx=transaction-hash.u.mined.log]
+        +>
+      ~&  :*  'removed event!  Perhaps chain has reorganized?'
+              tx-hash=transaction-hash.u.mined.log
+              block-number=block-number.u.mined.log
+              block-hash=block-hash.u.mined.log
+          ==
+      +>  ::TODO  undo the effects of this event
```
Fang-: As above, I believe we made the changes necessary to support appropriate behavior here. Can we start actually using that in the logic?

sys/vane/ames.hoon
```diff
@@ -523,8 +142,9 @@
           ::
               rov
             |-  ^-  (unit lane)
-            :: XX check will
-            ?:  ?|  !=(our (sein:title her))
+            ::  XX jael scry
+            ::
+            ?:  ?|  !=(our (^sein:title her))
```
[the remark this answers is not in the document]
joemfb: It can't be trivially converted, because there's no `+sley` function in scope. I deferred because it's not a trivial refactoring. But this code only exists to support the %rove/%tend routing subscriptions used by :dns, and that is all being removed.

sys/vane/ames.hoon
```diff
@@ -1179,8 +754,58 @@
             ?^  ram  raz.bah
             %+  ~(put by raz.bah)  cha
             rum(dod &, bum ?~(cop bum.rum (~(put by bum.rum) did.rum u.cop)))
-          =^  roc  diz  (zuul:diz now [%back cop dam ~s0])
-          (busk(diz (wast:diz ryn)) xong:diz roc)
+          =/  seg  (sein our now her)
+          =^  roc  diz  (zuul:diz now seg [%back cop dam ~s0])
+          (busk(diz (wast:diz ryn)) xong roc)
+        ::  XX move this logic into %zuse, namespaced under %jael?
```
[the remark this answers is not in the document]
joemfb: Just a note that comet/moon signature creation/validation should have a single implementation in an obvious location. Currently spread across %ames, %jael, %zuse, and various generators.

sys/vane/ames.hoon
```diff
@@ -1695,52 +1421,38 @@
...
+          (~(gnaw am [now fox ski]) %dead p.kyz q.kyz)
         ::
             %hole
-          (~(gnaw am [now fox]) %dead p.kyz q.kyz)
+          (~(gnaw am [now fox ski]) %dead p.kyz q.kyz)
         ::
+            %init
+          :_  fox  [[%sake p.kyz] [%brew ~] ~]
+        ::
+          ::  XX confirm unused, remove
```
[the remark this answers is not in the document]
joemfb: Or vere, for that matter. But the entropy provided in %junk is used to create symmetric keys, and no other code sets it.

sys/vane/jael.hoon
```diff
@@ -49,13 +48,23 @@
...
 ++  state-relative                                      ::  urbit metadata
   $:  $=  bal                                           ::  balance sheet (vest)
         $:  yen/(set duct)                              ::  trackers
         ==                                              ::
       $=  own                                           ::  vault (vein)
         $:  yen/(set duct)                              ::  trackers
+            :: XX use this                              ::
+            our=ship                                    ::
+            sig=(unit oath)                             ::  for a moon
+            :: XX reconcile with .dns.eth               ::
```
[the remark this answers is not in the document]
joemfb: I think so, and I don't remember where I picked it up. The XXs here are about a) proper single-homing and b) overall %jael state design. Nothing urgent.

sys/vane/jael.hoon
joemfb: The current fake-key design derives data as needed, and would be messed up by actually receiving ethereum events. We could instead ignore or clear this data if set (here, or in %ames, or in vere), but it really just shouldn't be set imo.

sys/vane/ames.hoon
```diff
@@ -1665,9 +1354,46 @@
     ^-  [(list move) _+>]
     ?-  +<.sih
         %crud  [[[hen [%slip %d %flog +.sih]] ~] +>]
+    ::
         %mack  ?~  +>.sih  $(sih [%g %nice ~])          ::  XX using old code
                $(sih [%g %mean `[%mack +>+.sih]])
+    ::
+        %turf
+      =.  tuf.fox  turf.sih
+      ::  XX give %turf if already recvd %bar?
```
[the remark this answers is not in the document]
joemfb: Terrible comment, made worse by a typo ;) There's an abstract race condition on first boot here. (There's actually lots of these across vane initialization). But it's only abstract, because of a workaround/hack in vere. I'll remove the comment.

sys/vane/ames.hoon
```diff
@@ -1776,35 +1519,14 @@
...
-      =+  luw=(rick mar our law.saf.u.gys)
-      ?.  &(?=(^ muc) ?=(^ luw))  ~
-      ``[%noun !>(`buck`[u.muc u.luw])]
+    ::  XX move to %jael
+    ::
     ?:  ?=([%code ~] tyl)
-      ``[%noun !>((end 6 1 (shaf %pass (shax sec:ex:q:sen:u.gys))))]
-    ?:  ?=([%will ~] tyl)
-      (bind (rick mar our law.saf.u.gys) |=(a=wyll `[%noun !>(a)]))
+      ``[%noun !>((end 6 1 (shaf %pass (shax sec:ex:cub:sen:u.gys))))]
```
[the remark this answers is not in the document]
joemfb: Yeah, it should be. I also suspect it should use `+pac:ex:crub:crypto`.

sys/vane/jael.hoon
joemfb: This is about meeting ships when we don't yet have their keys. This is always the case for moons and comets, and otherwise the case for our sponsor. If the "galaxy table" we supply in the boot event included the sponsor's keys and/or full sponsorship chain, this special casing (and related logic in %ames) would not be necessary.

sys/vane/ames.hoon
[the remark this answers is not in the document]
joemfb: Done.

sys/vane/jael.hoon
joemfb: Refactored

joemfb:
@Fang-, thanks for your multiple, thorough reviews. I realize how difficult that is with a PR of this size, changing as frequently as this has. I'll use your comments as a guide for a style/stability pass after our architecture review.

sys/vane/jael.hoon
philipcmonk: This is obsolete; will remove

sys/vane/jael.hoon
philipcmonk: Will do

sys/vane/jael.hoon
philipcmonk: I believe this check is valid again, since restoring a snapshot reverts `latest-block`, and we intercept `removed` cards earlier.  At any rate, if it isn't true, I would like to investigate that case.

I'll reinstate it for now.

sys/vane/ames.hoon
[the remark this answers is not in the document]
Fang-: That's... even weirder. Is the conclusion that it doesn't get set? This means keys are determined exclusively [by timestamps](https://github.com/urbit/arvo/blob/999255bc39b8afed5c1b70b4accf2e52cfba6074/sys/vane/ames.hoon#L89), which isn't exactly strong entropy.

Can we conclude the real `XX` here is to actually use `%junk`?

Fang-:
I wouldn't call them thorough per se. Asking about comments is low hanging fruit as far as reviews go. As said, I think I understand the broad strokes of what's happening here, but I haven't been able to map that onto the code very precisely, barely at all. Again, PR size, yada yada. Doing multiple smaller PRs doesn't feel *much* better for components that are as heavily in flux as this was during development, but at least it'd be *some* improvement. Another one for on the post-mortem/process feedback list.

That said, poking at some low-hanging buttons like todo's is still good exercise for both of us. Thanks for your answers, realizations and fixes! This has been a huge effort, congrats to you and @philipcmonk on the merge. (:

sys/vane/ames.hoon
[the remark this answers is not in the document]
joemfb: Yes. Or more generally, review and correct the generation mechanism for symmetric keys. %xmas deferred to (old) %jael for this, perhaps we should too? I will compile a list of questions and next steps from the comments in the code and here.

sys/vane/ames.hoon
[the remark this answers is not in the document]
joemfb: Indeed. Correcting (or at least clarifying) this is one of the highest priorities after integrating cc-release, imo. Before that, I'm not sure it's worth pursuing. At this time, I don't know really know more than what I've already said here ...

PR #861, opened 2018-10-22
Fix bug in previous PR

cgyarvin:


joemfb:
There's no need for this, I updated the previous one to point to `release-candidate`.

PR #862, opened 2018-10-24
add a convenient error message

philipcmonk:
@joemfb Not sure what's the best flow for making these sorts of changes.  Should I just push them straight to ames-eth-keys?


sys/zuse.hoon
```diff
@@ -364,7 +364,9 @@
     ++  contracts
       |%
       ++  ships
-        0x84b3.7fbc.6188.da8a.e866.1eae.322a.f4d9.2db4.5ecc
+        ::  0xab87.24a7.a953.ef14.e940.b358.6b21.a889.b62f.3d56  ::  ropsten
+        0x84b3.7fbc.6188.da8a.e866.1eae.322a.f4d9.2db4.5ecc  ::  joe
+        ::  0x7134.3566.74e4.0c93.8736.8699.1af8.86dd.2ae8.e642  ::  philip
```
Fang-: Changes to this should probably only be committed for "official deploy of some kind" addresses.

joemfb:
Yeah, that's probably fine. I'll do my best to avoid rebases moving forward ;)

sys/zuse.hoon
joemfb: Agreed, but this PR is against a still-in-progress branch. I'm fine with it for now, let's finalizing and clean this up (as much as possible) when #858 is ready to be merged.

PR #863, opened 2018-10-24
remove clam from error sigpam

vvisigoth:


ohAitch:
?
Note that it was just using the clam to print "what kind of card is this,
ignore body"; this could also be e.g. [&1 &2]:+.sih

PR #864, opened 2018-10-25
add chunking and start snapshotting

philipcmonk:
I hardcoded the block number of the initial deployment, and now it chunks every 5760 blocks (~1 day) from that point.  This should reduce the frequency of timeout problems quite a lot.

I've also got the beginnings of a snapshot system there (it takes the snapshots, but there's no way to restore them yet).

### DIFF
--- a/sys/vane/jael.hoon
+++ b/sys/vane/jael.hoon
@@ -48,6 +48,7 @@
       urb/state-absolute                                ::  all absolute state
       sub/state-relative                                ::  all relative state
       etn=state-eth-node                                ::  eth connection state
+      sap=state-snapshots                               ::  state snapshots
   ==                                                    ::
 ++  state-relative                                      ::  urbit metadata
   $:  $=  bal                                           ::  balance sheet (vest)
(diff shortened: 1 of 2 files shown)


joemfb:
It looks like the hardcoded initial block is missing in zuse. (At least that's how I read the CI failure)

philipcmonk:
ah, yes.  Just added that.

sys/vane/jael.hoon
```diff
@@ -84,7 +85,20 @@
   $:  source=(each ship node-src)                       ::  learning from
       heard=(set event-id)                              ::  processed events
       latest-block=@ud                                  ::  last heard block
+      foreign-block=@ud                                 ::  node's latest block
   ==                                                    ::
+++  state-snapshots                                     ::  rewind points
+  $:  interval=_100                                     ::  block interval
+      max-count=_10                                     ::  max snaps
```
Fang-: Maybe set these manually in `++init`? idk if value-by-default is considered good style in nu-hoon, but I rarely ever see this used. (Though, then, almost always the bunts are the desired defaults.)

sys/vane/jael.hoon
```diff
@@ -1604,7 +1712,10 @@
     %-  put-request
     :+  /filter/new  `'new filter'
     :*  %eth-new-filter
-        `[%number ?:((lte latest-block 40) 0 (sub latest-block 40))]
+        `number+latest-block
+        ::  XX We want to load from a snapshot at least 40 blocks behind, then
+        ::  replay to the present
+        ::  `[%number ?:((lte latest-block 40) 0 (sub.add latest-block 40))]
```
Fang-: 40 still feels really arbitrary, and longer reorgs are not entirely out of the question. Of course, nothing a re-`%init` can't solve...

PR #866, opened 2018-10-26
Remove old zuse

rmariani:
Removes all references to `old-zuse.hoon` and replaces the arms that it calls from that file with the appropriate arms in `zuse.hoon.`

big thanks to @joemfb for showing me how to use the test app.


ohAitch:
this too should be a PR against rc probably :P

rmariani:
should be going to rc

Fang-:
For future reference: you can hit the "edit" button at the top of the PR to edit not just the title, but also the target branch. (:

PR #870, opened 2018-10-30
%jael: Allow manual restore from snapshot

philipcmonk:
The `%wind` event now lets you rewind to a particular block.

Specifically, we rewind to the most recent snapshot previous to that block number (or blank state if it's before all our snapshots), then run `%look` with our current node to start catching up again.

Additionally, `.ropsten/snap .^(noun %j /=snap=)` saves the most recent snapshot to disk.  There's no way to restore from it yet, though.

I've also included a variety of robustness fixes.


philipcmonk:
This is what I've been using to test:  https://gist.github.com/philipcmonk/125e7f5cab49b66109be8ec5e0877941

sys/vane/jael.hoon
```diff
@@ -1005,6 +1011,54 @@
...
+          last-block.sap  block-number.snap
+          snaps.sap       (~(put to snaps.sap) snap)
+        ==
+      ?:  |(=(~ old-qeu) (lth block block-number:(need ~(top to old-qeu))))
+        [snap.snap +>.^$]
+      $
+    ~&  [%wind block latest-block.etn.snap (~(get by hul.eth.sub.snap) ~ben) ~(wyt by hul.eth.sub.snap)]
+    ::  Keep the following in sync with ++extract-snap:file:su
+    %=  +>.$
+      urb   urb.snap
```
Fang-: Shouldn't this just be restoring `eve.urb`, rather than all of `urb`? I'd guess that this would just be rolling back eth-related state.

So, this is probably more of a general snapshots question that I should've surfaced during the previous PR, but: are snapshots dealing with more than just eth- and eth-derived state? Why? I don't think we want to mess with misc (currently unused) jael state, do we?

philipcmonk: Yes, this should just be rolling back eth-derived state.

I have no idea what `pry.urb` is, so you're probably right.  I'll limit this to just `eve` within `urb`.  Do you see other state that shouldn't be stored in a snapshot?

sys/vane/jael.hoon
```diff
@@ -1005,6 +1011,54 @@
...
+      sub   %=  sub.snap
+              yen.bal  yen.bal.sub
+              yen.own  yen.own.sub
+              yen.puk  yen.puk.sub
+              yen.eth  yen.eth.sub
+            ==
+    ::
+      etn   %=  etn.snap
+              source  source.etn
+            ==
```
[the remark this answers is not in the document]
philipcmonk: Nope; will fix

sys/vane/jael.hoon
Fang-: Probably `bal.sub` and `own.sub` too.

philipcmonk: Aren't at least `[jaw lyf tuf]:own.sub` downstream of ethereum?

joemfb: `own` has just become a somewhat messy grab-bag of state.

`tuf.own` *should* be downstream of ethereum, but is currently just set on boot and static thereafter. This needs to be the case for urbit/ethereum testnets, since the contract will (usually) not have the correct testnet domain in it.

`[jaw lyf]:own` are set on first boot, and will be subsequently changed if private keys are rotated with `%mint %jewel`. This is a holdover from the old %jael rite/promises/attestations/pki stuff. The intersection of that and the new ethereum stuff could really use a re-design pass, imo.

Fang-: Yes, `kyz.puk.sub` should be part of the snapshots I think. They're public keys, which should be downstream from Ethereum as well.

sys/zuse.hoon
```diff
@@ -8022,7 +8022,7 @@
       %+  turn  (gulf 0 255)
       |=  gal=@
       %+  request-to-json:ethereum
-        `(scot %ud gal)
```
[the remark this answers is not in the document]
philipcmonk: Yes, and perhaps this should be in comments somewhere.  Basically, if parity can parse an id as a number, then it returns it as a JSON number.  Geth just spits it back to you how you gave it.  I've opted to force everything into strings, but I *think* we could also resolve this by making them real numbers and parsing them as such.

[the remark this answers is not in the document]
philipcmonk: Also, I'll try to be better about making commits be about one thing at a time.  As a workflow, I work on several problems at once so that I can keep working during the long build cycle, but that shouldn't hit the git history.

sys/vane/jael.hoon
joemfb: The life and pass are, but the `live` bit in `+public` is not derived from ethereum (but from %meet moves sent by %ames upon neighboring with ship). We could put that in a separate structure if necessary.

sys/zuse.hoon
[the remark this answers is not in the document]
joemfb: I don't have any objection to stringifying these. Could you update the tests in `tests/sys/zuse/dawn.hoon`? You can run them specifically with `+test /sys/zuse/dawn`. (for some reason, the CI runner has stopped detecting test failures, which is kind of counter-productive ...)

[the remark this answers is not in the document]
Fang-: > but I think we could also resolve this by making them real numbers and parsing them as such.

As far as the RPC stdlib stuff goes, we probably don't want to do that, since some apps may want to use this with text IDs still. (otoh, a `@t` is a perfectly valid number. (^: )  
As-is, this is a fair workaround for an awful problem to have.

> Also, I'll try to be better about making commits be about one thing at a time.

Thank you! That's my workflow too, so I know separating changes properly can be a pain sometimes. The effort is much appreciated though! (:

sys/vane/jael.hoon
philipcmonk: Now that `jaw` isn't in the snapshot, is there anything else that should be purged before one gives a snapshot to someone else (e.g. a star to a planet)?

PR #874, opened 2018-11-01
adds robustness to :acme app

joemfb:
This PR implements more robust request retries on timeouts and other recognizable failure conditions.

Due to testnet woes, this has not yet been fully tested. Opening now for review, additional fixes will be pushed here as necessary. Rate-limit handlers should also be implemented here.

### DIFF
--- a/app/acme.hoon
+++ b/app/acme.hoon
@@ -20,15 +20,15 @@
   ++  de  |=(a=octs `@u`(rev 3 p.a q.a))
   --
 ::
-:: |grab: acme api response json reparsers
+::  |grab: acme api response json reparsers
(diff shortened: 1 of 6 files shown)


joemfb:
Definitely needs more testing before merging. I'll notify you if I make any big changes.

joemfb:
This PR includes all my planned :acme improvements except a) proper exponential backoff for retried requests and b) api rate-limit handling. Exponential backoff will either require a significant amount of additional state and duplicated logic, or a scheme for generically encoding attempt-counts in wires. I'll work to hit some rate limits soon, and follow up with improvements there.

PR #875, opened 2018-11-02
Make Ethereum snapshots work with children

philipcmonk:
OT: How's the frequency/size of PR's?  Should I be bunching these changes into larger PR's?

Couple of things:

- Use snapshots to initialize children rather than just sending them the event log and having them rederive the state.
- Properly update subscribers when a snapshot happens, as described in the following email snippet:

> When we restore a snapshot, it should be fine to just spam out a `%pubs` update for each ship to every subscriber, right?  In practice, it looks like the subscribers are ames for each our neighbors (?), so this will be linear in the number of neighbors, which is fine.  Subscribers just need to make sure they the the update as complete -- i.e. it may remove keys that you thought were real.
>
> For vents, I figure we just use the snapshot mechanism itself.  When a snapshot is applied, all vent subscribers should get that snapshot.  Thus, for example, when ~zod restores a snapshot, the flow is:
>
> - ~zod jael spams a pub to ames (1 per neighbor)
> - ~zod passes the snapshot to ~marzod
> - ~marzod spams a pub to ames (1 per neighbor)
> - ~marzod passes the snapshot to its children

### DIFF
--- a/sys/vane/ames.hoon
+++ b/sys/vane/ames.hoon
@@ -1365,13 +1365,18 @@
       ~&  [%hear-pubs tea sih]
       =/  our=ship  (slav %p i.t.tea)
       =/  her=ship  (slav %p i.t.t.t.tea)
-      ?:  =(0 life.sih)
-        ~&  %ames-hear-empty-pubs
-        [~ +>.$]
-      =/  ded=deed
-        [life.sih (~(got by pubs.sih) life.sih) oath=~]
       =/  gus  (need (~(us go ton.fox) our))
       =/  diz  (myx:gus her)
+      ?:  =(0 life.sih)
+          ::  this should clear lew.wod.dur.diz because it means
+          ::  we know longer trust that their public key came to
+          ::  us honestly (becuse of a %jael snapshot restore).
+          ::  in practice, that crashes in ++cluy:las:as:go, so
+          ::  we ignore for now.
+          ~&  [%ames-hear-empty-pub her]
+          [~ +>.$]
+      =/  ded=deed
+        [life.sih (~(got by pubs.sih) life.sih) oath=~]
       =.  lew.wod.dur.diz  `ded
       =.  ton.fox  (~(su go ton.fox) (nux:gus diz))
       [~ +>.$]
--- a/sys/vane/jael.hoon
+++ b/sys/vane/jael.hoon
@@ -2091,7 +2091,21 @@
   ::                                                    ::  ++restore-snap:et
   ++  restore-snap                                      ::  restore snapshot
     |=  snap=snapshot
+    ::  update pub subscribers
+    ::
+    =.  +>.$
+      =/  subs=(list (pair ship (set duct)))
+        ~(tap by yen.puk.sub)
+      |-  ^+  +>.^$
+      ?~  subs  +>.^$
+      =/  pub  (fall (~(get by kyz.snap) p.i.subs) %*(. *public live |))
+      =.  +>.^$  (exec q.i.subs [%give %pubs pub])
+      $(subs t.subs)
+    ::  update vent subscribers
+    ::
+    =.  +>.$  (vent-pass yen.eth snap+snap)
     ::  keep the following in sync with ++extract-snap:file:su
+    ::
     %=  +>.$
         eve.urb       eve.snap
         etn           etn.snap(source source.etn)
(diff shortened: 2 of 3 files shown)


sys/vane/ames.hoon
```diff
@@ -1365,13 +1365,18 @@
...
+      =/  gus  (need (~(us go ton.fox) our))
+      =/  diz  (myx:gus her)
       ?:  =(0 life.sih)
-        ~&  %ames-hear-empty-pubs
-        [~ +>.$]
+          ::  this should clear lew.wod.dur.diz because it means
+          ::  we know longer trust that their public key came to
+          ::  us honestly (becuse of a %jael snapshot restore).
+          ::  in practice, that crashes in ++cluy:las:as:go, so
+          ::  we ignore for now.
```
joemfb: What should we do other than crash? Don't we want to drop packets from ships for which we have no trusted key?

philipcmonk: We do want to drop those packets.  This crashes the actual event that clears `lew`, which is bad because we need to get the later packets to get the public key again.

One thing to be careful about is making sure we don't end up in a situation where we drop the key for our sponsor and consequently don't get the later packets which would bring us back up to date.

philipcmonk: > My mental model of this was that we did the "forget, load in snapshot state" in a single event, even to subscribers?

Firstly, this crash means we can't even ack this packet, so the "load snapshot" packet just keeps coming again and again.

Secondly, we do forget and load snapshot state in a single event, but the snapshot could be missing recent events.  Currently we snapshot once every 100 events, then on child boot we send the most recent snapshot followed by the later events, and in theory those missing logs could be important to the conversation.  More generally, the parent could revert back to any block, even block 0.  In that case, all current subscribers would hear "revert to the blank snapshot, then I'll send you events as I hear about them".

sys/vane/jael.hoon
```diff
@@ -1491,7 +1493,7 @@
           last-block  block.wer
         ==
       =?  sap  (gth count.sap max-count.sap)
-        ~&  :*  %dump  count=count.sap  max-count=max-count.sap 
+        ~&  :*  %dump  count=count.sap  max-count=max-count.sap
```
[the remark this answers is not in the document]
philipcmonk: I value style comments and will change this.

sys/vane/ames.hoon
```diff
@@ -1365,13 +1365,18 @@
       ~&  [%hear-pubs tea sih]
       =/  our=ship  (slav %p i.t.tea)
       =/  her=ship  (slav %p i.t.t.t.tea)
+      =/  gus  (need (~(us go ton.fox) our))
+      =/  diz  (myx:gus her)
       ?:  =(0 life.sih)
-        ~&  %ames-hear-empty-pubs
-        [~ +>.$]
+          ::  this should clear lew.wod.dur.diz because it means
+          ::  we know longer trust that their public key came to
```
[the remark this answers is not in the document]
philipcmonk: Ha!

sys/vane/ames.hoon
Fang-: That's.. completely correct. I was holding onto too many assumptions still, but we can't really make any when we rewind!

joemfb: We could set a timer to disassociate the "processing a snapshot" event from the "received a packet containing a snapshot" event. (Breaking up this and other log processing into separate, batched, smaller events is something I'd like to discuss in more detail). But we'll still definitely need special handling for our sponsor here.

I'm comfortable merging this as is. But we should talk through these scenarios soon. (Not to mention how they interact with discontinuity ...)

philipcmonk: Yeah, I'll go ahead and merge this tomorrow after fixing @Fang-'s cosmetic notes.

Currently, I believe the only known failure mode is if a block that sets the public key (hereafter, k1) for someone you talk to is reorg-ed out of the chain and *not* re-added to the chain, then we could still talk to them under a key that isn't reflected on chain (i.e. k1).  This isn't a very serious vulnerability because such a transaction can only be sent by the actual owner of the ship, so if e.g. ~zod picked up the tx and spammed it to everyone but the ship owner had a change of heart and decided not to add k1 after all, then unless he bumped the life to another key k2, some people might only recognize k1, which presumably he controls anyway.  He could work around this by bumping his life.

Basically, we don't need much in the way of double-spend protection because the rightful owner can always reclaim his belongings by cycling his keys (or submitting a PR to fix this bug! :) )

PR #877, opened 2018-11-02
Keygen: latest spec & parity with keygen-js

Fang-:
I realize now that I haven't put up a PR for this yet. So here it is!

This should bring keygen-hoon up to parity with keygen-js, adhering to the latest spec. Before we merge this, I want to put in tests (copied from keygen-js) that prove that's actually the case. All of the code should already be final though, so review away.

...That said, we should probably merge #825 first, but that depends on getting urbit/urbit#1028 merged, which is still on my tasklist but has somehow evaded my view for a little while. I'll put it in my OKRs so I won't forget again!


jtobin:
Did you manage to resolve the problem mentioned in urbit/keygen-js#35 btw?

joemfb:
> I've never tried to run a script straight from booting a fresh Urbit

Oh you sweet summer child: https://github.com/urbit/arvo/tree/master/.travis

jtobin:
> Oh you sweet summer child: https://github.com/urbit/arvo/tree/master/.travis

Fang-:
> Did you manage to resolve the problem mentioned in urbit/keygen-js#35 btw?

The problem had a very different shape in keygen-hoon, but now: sort of. The seed derivation function takes a flexible bytelength input, but the bytelength is always specified, and internally the only lengths that are passed are 32 bytes (the common case) and 64 bytes (the "derive network seed from management bip39" case).

Maybe the seed derivation function should assert that the input is one of those two lengths?

cgyarvin:
I would really like to see this happen... I don’t think it’s hard at all.  Just dump output to put/ the way you do when you make a pill.

Fang-:
Tests are in. I realize now there isn't exact feature parity yet, most significantly in the interface of full wallet generation (returning a (sometimes) sharded ticket, the `booted=?` flag, working for just a single ship, not generating voting keys for non-galaxies). I'll fix those up tomorrow, they should be largely trivial.

Work done here (especially related to tickets) would look a lot nicer if #824 were in already. Soon™!

Fang-:
Updated the CI vere pointer so that tests will actually complete (probably), but I guess this means merging this is blocked on merging of urbit/urbit#1028.

joemfb:
@Fang-, you probably need to merge release-candidate back into crypto-additions (and maybe here too). That build failure suggests a pretty large mismatch between Arvo and vere (like, say, nock 4K).

PR #878, opened 2018-11-02
Rewind automatically when we detect an Ethereum reorganization

philipcmonk:
This adds rewinding in the case where our node tells us that a log got reorg-ed out of the chain, as suggested by @Fang-.

Arguably, we should also rewind when a filter times out or when we switch nodes.  I've opted against that for now and noted the reasons for that decision in the source.  It's a little unusual to put those arguments in the source, but if we stumble over a reorg and someone's debugging it, I want them to see the potential pitfalls of how it works now.

I believe this puts a bow on my snapshotting work, so any final comments or suggestions are welcome.

### DIFF
--- a/sys/vane/jael.hoon
+++ b/sys/vane/jael.hoon
@@ -1107,6 +1107,14 @@
       /(scot %p our)/vent-result
     ^-  note:able
     [%a %want [our who] /j/(scot %p our)/vent-result %vent-result res]
+  ::
+  ++  extract-snap                                    ::  extract rewind point
+    ^-  snapshot
+    :*  eve.urb
+        kyz.puk.sub
+        +.eth.sub
+        etn(source *(each ship node-src))
+    ==
   ::                                                    ::  ++feed:su
   ++  feed                                              ::  subscribe to view
     |_  ::  hen: subscription source
@@ -1138,13 +1146,7 @@
       ==
     ::
     ++  vent
-      =/  last-snap
-        |-  ^-  snapshot
-        =^  snap=[@ud snap=snapshot]  snaps.sap
-          ~(get to snaps.sap)
-        ?:  =(~ snaps.sap)
-          snap.snap
-        $
+      =/  last-snap  extract-snap
       %.  [[hen ~ ~] snap+last-snap]
       %_  vent-pass
       :: %_  ..feed  ::TODO  see ++abet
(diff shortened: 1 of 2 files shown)


philipcmonk:
Whoops, one last bug fix

sys/vane/jael.hoon
```diff
@@ -1122,6 +1107,14 @@
       /(scot %p our)/vent-result
     ^-  note:able
     [%a %want [our who] /j/(scot %p our)/vent-result %vent-result res]
+  ::
+  ++  extract-snap                                    ::  extract rewind point
+    ^-  snapshot
+    :*  eve.urb
+        kyz.puk.sub
+        +.eth.sub
+        etn(source *(each ship node-src))
```
Fang-: This defaults to ~zod. We should arguably change the `each` around so it defaults to an "empty" `node-src` instead, but I'm not sure if this actually matters.

sys/vane/jael.hoon
```diff
@@ -2090,9 +2094,34 @@
...
     (put-change cuz %hull u.dif)
+  ::                                                    ::  ++restore-block:et
+  ++  restore-block                                     ::  rewind before block
+    |=  block=@ud
+    ^+  +>
+    =/  old-qeu  snaps.sap
+    =:  snaps.sap       ~
+        count.sap       0
+        last-block.sap  0
+      ==
```
Fang-: Your closing `==` are consistently indented (here and in other uses of `=:`). I'm used to placing/seeing them on the same indentation level as the `=:`.
I notice that for `:*`, you don't indent the `==`. Is the indent a way of saying "this is not a leaf hoon"?

philipcmonk: I've seen it both ways in the past, don't know what's accepted practice nowadays.  To me, `==` or `--` followed by another line on the same indentation level feels wrong.  As you say, the variable-arity portion of `:*`, `:~`, `?-`, `?+`, `|%`, etc is at the end of the hoon while it's not for `=:`.

sys/vane/jael.hoon
philipcmonk: We always replace `source.etn` in `++restore-snap`, so it doesn't matter in practice.  Arguably we should make our own little `++state-eth-node-snapshot` structure rather than keeping around dummy state like this.

sys/vane/jael.hoon
philipcmonk: Looks like hoon.hoon and arvo use indent-style, zuse.hoon uses dedent-style.

Fang-: Interesting. Indenting closing glyphs for non-leaf hoons seems sanest to me, so do carry on.

PR #880, opened 2018-11-06
refactors %eyre %born task, prunes pending requests

joemfb:
I've been seeing the %eyre `%pump-blocked` error from urbit/urbit#784 on the testnet. I believe it's caused by open requests persisting across restarts.

### DIFF
--- a/sys/vane/eyre.hoon
+++ b/sys/vane/eyre.hoon
@@ -678,9 +678,45 @@
     =.  our  ?~(hov our u.hov)  ::  XX
     =.  p.top  our              ::  XX necessary?
     ?-    -.kyz
+        ::  new unix process - learn of first boot or a restart.
+        ::
         $born
-      ::  XX capture IPs too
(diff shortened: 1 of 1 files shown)


ohAitch:
Wasn't resolved by cttp rewrite? ;)

(Doing this in born instead of explicitly in cttp is probably fine yeah)

joemfb:
To be fair, `cttp.c` does send a 504 if it's still running when the socket is closed ... ;)

ohAitch:
Ah okay, so #784 should have been closed, but this fixes something adjacent?

joemfb:
urbit/urbit#784 is about the absence of a 504 response when the underlying TCP socket for a client request closes. But it's also about `%pump-blocked` in %eyre. The first was resolved, the second wasn't. I'm not completely sure if the vane or the I/O driver should have final responsibility for cleaning up, but it seems good for the vane to do some general housekeeping when a new process is started.

/cc @cgyarvin

PR #881, opened 2018-11-06
Remove 200ms delay in +run-builds in ford.

eglaysher:
In +run-builds in ford, the ordering of `~%  ...  |-`  expands to  `~%  ...  =<  $  |.`, which is wrong and causes a 200ms delay for reasons I suspect are related to battery registration. We then went through and made sure that the ~% is registered directly on the cores in a few places in Eyre, too.

Locally, it now takes about a second to roundtrip a message from landscape to urbit and back.

review: @ixv 
cc: @belisarius222


cgyarvin:
Great — but that’s still insanely slow, brothers...

cgyarvin:
Wow

belisarius222:
Yikes! I apologize for my role in that ... uh ... misguided approach to jet-hinting. Hope I didn't do that anywhere else.

joemfb:
@eglaysher, should this PR be against the release-candidate?

Based on that CI build log, I think some of parent axes in these hints are now wrong.

cgyarvin:
Paul, a bad axis in a hint should definitely not be able to cause a
bail-foul... but if such a bug existed it wouldn't shock me :-/

eglaysher:
Merely commenting out the `~%  ...  |-` block at the top of the top of /sys/vane/ford+run-builds causes a `bail: foul`. (Putting the ~% back fixes it.) 

Slipping this patch into a new zod makes the tests run the first time. Then it causes a `bail: foul` on each subsequent run of the test suite.

Building a new zod from the testnet-rc branch, running the ford tests, making a small change, and then running the ford tests again does not cause a bail.

joemfb:
I'm not claiming that to be the cause -- just pointing the immense number of "parent not found" errors in the build log. It exceeds travis-ci's log length limit, so I can't even see the bail.

I believe the `%eyre` changes in this PR are unnecessary. There shouldn't be anything wrong with a jet registration hint that precedes pinning something to the subject and then producing a core. The only requirement is that the parent axis lines up.

cgyarvin:
Obviously we're going to need to debug this with or without Paul's help :-/

joemfb:
I've updated #879 to print a backtrace if we dump core in CI. Merging that, reparenting this on release-candidate, and fixing the parent axes would provide some nice context. Of course, that context is available in other ways ...

belisarius222:
no one has summoned @frodwith?

joemfb:
With urbit/urbit#1059 merged, is this PR ready to be revived?

eglaysher:
Let me minimize the patch first.

joemfb:
To fix CI, you'll probably need to bump `.travis/pin-vere-commit.txt` so it point to latest release-candidate.

ixv:
You should merge this into release-candidate arvo too. testnet-rc is dead.

joemfb:
Thanks @ixv, I just took the liberty of re-targeting this.

eglaysher:
That doesn't work, the branches are too diverged. I'm going to just rebuild it.

joemfb:
FWIW, you can always force-push to your side of a PR.

PR #886, opened 2018-11-09
send-txs improvements

philipcmonk:
Still have several more improvements to make, but I wanted to get release-candidate merged in, as well as my first couple of improvements.  Look at the last two commits to see what I've actually changed.

The multiple-node spamming isn't 100% robust yet, because it still relies on a single node to provide status updates, but at least the sends should be redundant.  I'll think about how to automatically fail over from one node or something.

The second commit makes a little pipeline so that every time we check for tx confirmations, we send out as many new txs as were confirmed, keeping us as close to 500 outstanding txs as possible.


philipcmonk:
I merged release-candidate with ceremony, but it looks like that hasn't worked through github's cache, so check this: https://github.com/urbit/arvo/compare/ceremony...philip/ceremony

I also added one more commit, which adds an option to generate txs from :send-txs

app/send-txs.hoon
```diff
@@ -100,13 +178,14 @@
...
-  %+  turn  ~(tap in see)
+  =/  s  ~(tap in see)
+  =/  extra  (slag 50 s)
+  =/  check  (scag 50 s)
+  :_  this(see (silt extra))
+  %^  batch-requests  /see  |
+  %+  turn  check
   |=  txh=@ux
-  :-  `(crip '0' 'x' ((x-co:co 64) txh))
+  :-  `(crip 'see-' '0' 'x' ((x-co:co 64) txh))
```
Fang-: This works, but it feels wrong. I'd expect `(weld "see-0x" ((x-co:co 64) txh))` here.

app/send-txs.hoon
```diff
@@ -123,7 +202,10 @@
     |=  r=response:rpc:jstd
     ^-  (unit @ux)
     ?<  ?=(%batch -.r)
-    =+  txh=(tape-to-ux:ceremony (trip id.r))
+    ?<  ?=(%fail -.r)
```
Fang-: Now that we have multiple nodes, might we not want to count node connection fails as just "not confirmed"/`wait`?

lib/ceremony.hoon
```diff
@@ -207,7 +215,9 @@
   =.  network
     ?+  net  0x1
       %ropsten  0x3
+      %fake     `@ux``@`1.337
```
Fang-: Local ganache fakenode uses `0x1` always, as far as I'm aware. Is that configurable?

lib/ceremony.hoon
```diff
@@ -243,7 +253,9 @@
             =<  net
             %+  fall
               (~(get by deed-map) par)
-            (~(got by tlon-map) par)
+            %+  fall
+              (~(get by tlon-map) par)
+            *rights
```
Fang-: iirc if it's not in `deed-map` it *needs* to be in `tlon-map`. I'm guessing you hit the Tlon stars with this?

app/send-txs.hoon
```diff
@@ -23,59 +24,134 @@
...
+  |=  [wir=wire fan=? req=(list [(unit @t) request:ethe])]
+  ^-  (list move)
+  %^    fan-requests
+      wir
+    ?:  fan
+      :~  =>  (need (de-purl:html 'http://localhost:8545'))
+          geth+.(p.p |)
+        ::
+          ::  =>  (need (de-purl:html 'http://localhost:8555'))
+          ::  parity+.(p.p |)
```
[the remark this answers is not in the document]
philipcmonk: It's easier to test other stuff by only requiring one node running locally.  In real life none of the nodes will be at localhost anyway, so I'm not sure it matters much.

Fang-:
Also CI seems to be failing for the same reason keygen/ceremony branches are failing: no jets yet. I continue to be on the case!

app/send-txs.hoon
```diff
@@ -23,59 +24,134 @@
...
-    =+  (need (de-purl:html 'http://localhost:8545'))
-    -(p.p |)
+  (json-request:ethereum url jon)
+::
+++  batch-requests
+  |=  [wir=wire fan=? req=(list [(unit @t) request:ethe])]
+  ^-  (list move)
+  %^    fan-requests
+      wir
+    ?:  fan
```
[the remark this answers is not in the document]
philipcmonk: As it's currently written, we only want to send getTransactionReceipt requests to one node, but we want to send sendTransaction requests to all nodes.  I think I'm going to start making all the requests go to all the nodes, which will remove the need for this option.

app/send-txs.hoon
```diff
@@ -23,59 +24,134 @@
...
-  ~&  'sending 50 txs...'
-  :_  ~
-  %+  batch-requests  /send
-  %+  turn  (scag 50 txs)
+  ~&  ['sending txs...' new-count]
+  %^  batch-requests  /send  &
+  %+  turn  (scag new-count txs)
   |=  tx=@ux
-  :-  `'id'
+  :-  `(crip 'id-' (scot %ux (end 2 10 tx)) ~)
```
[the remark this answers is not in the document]
philipcmonk: I was mostly adding this to help debugging because eyre was crossing wires.  I think maybe I was trying to do `3 10`?  I'll change it.

app/send-txs.hoon
philipcmonk: Sure

app/send-txs.hoon
philipcmonk: yeah, you're right.  I'll test killing one node and make sure it continues.

lib/ceremony.hoon
philipcmonk: `fall` is not a macro, so this failed if the ship was not in `tlon-map` even if it *was* in `deed-map` (because the `got:by` got called anyway).

app/send-txs.hoon
```diff
@@ -23,59 +24,134 @@
...
     |=  r=response:rpc:jstd
     ^-  (unit @ux)
     ?:  ?=(%error -.r)
       ?:  =('Known transaction' (end 3 17 message.r))
-        ~&  %sent-a-known-transaction--skipping
+        ~&  %sent-a-known-transaction--skipping-cap
+        ~
+      ?:  =('known transaction' (end 3 17 message.r))
+        ~&  [%sent-a-known-transaction--skipping wir r]
+        ~
```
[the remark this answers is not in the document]
philipcmonk: I think the errors get sent by different nodes or at different stages, so I'll keep all the condition, but merge them into one branch with `?|`.

philipcmonk:
Thanks for the review and taking a look at CI.

(for the record, there was no delay -- I submitted the PR at 5pm, and you reviewed it at 10:10am on the next business day)

lib/ceremony.hoon
Fang-: oh christ, I feel so dumb for getting stuck on this some time ago lol

philipcmonk:
Just pushed a bit of a restructure of the state machine, and added some comments describing it.  This should be significantly less fragile.  I tested this against failure of one or all nodes, and it seems to work.

philipcmonk:
Two phase transaction generation works now.  The `eth-txs` mark is just comma-separated transaction objects.  We could totally load them into Lotus 1-2-3 and edit them there.

I also redid the `++poke-noun` interface to be more explicit.  I started using generators but decided they didn't buy enough over a simple tagged union.

PR #887, opened 2018-11-11
Invalid Hoon literals in sail syntax

mikolajpp:
Having unleashed working parts of the new parser onto entire arvo tree, the 
tokenizer uncovered a following problem:
In sail syntax it is permissible to have unquoted string literals. 
This is very bad for future code introspection purposes, and I would like to put this up for discussion, 
i.e. IMHO we in the future compiler sail syntax should only permit proper literals without 
non-standard quirks

### DIFF
--- a/sys/vane/eyre.hoon
+++ b/sys/vane/eyre.hoon
@@ -551,7 +551,7 @@
       ;head:title:"Redirecting..."
       ;body
         ;p: Redirecting to ;{a/"{url}" "{url}"}
-        ;script: setTimeout(function()\{document.location = {(en-json (tape:enjs url))}}, 3000)
+        ;script:'setTimeout(function()\{document.location = {(en-json (tape:enjs url))}}, 3000)'
       ==
     ==
   ::
--- a/sys/vane/eyre.hoon
+++ b/sys/vane/eyre.hoon
@@ -551,7 +551,7 @@
       ;head:title:"Redirecting..."
       ;body
         ;p: Redirecting to ;{a/"{url}" "{url}"}
-        ;script:'setTimeout(function()\{document.location = {(en-json (tape:enjs url))}}, 3000)'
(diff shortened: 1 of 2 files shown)


mikolajpp:
It seems sail does not handle cords and requires tapes; If this is by design then fine, 
although we could of course support all Hoon literals.

mikolajpp:
Anyway feel free to disregard this merge request and discuss only the issue itself. 
Such things can wait when the new parser will be ready, when it will also be possible to 
auto-migrate entire Arvo tree onto Hoon %141, so perhaps it is better to handle this in one huge dump

ohAitch:
Making sail require a line closing delimiter seems like a significant UI
downgrade 3:

At any rate, I will note that several of the lines contain {}
interpolations, so single quotes were breaking their semantics anyway

ohAitch:
In particular, iirc sail was designed as a slightly regularized version of
http://jade-lang.com/, taking out some of the implicit HTML defaults(and
put ; everywhere to distinguish from regular hoon). I'm not sure what you
mean by "introspection": a sail string literal is delimited by an opening
 '; ' and a closing newline, or alternately after a sail tag  ;[a-z]([^
]|QUOTE)*  with an opening ': ' and a closing newline. It has an
interpolation syntax vaguely similar to "", but with some possible prefixes
to {} to iterpolate in xml elements/lists/etc instead of pure text.

ohAitch:
I suppose the socially savvy may point out that jade has been throughly
supplanted by jsx, on which basis we should just make sail contents
implicitly-text and require a closing ;/div or w.e. instead of ==

Which I would be sad to hear, but probably still be considerably less
damaging to the ability to copy blocks of text into sail bodies than
requiring quotes, and correspondingly, quote escaping everywhere.

mikolajpp:
Makes sense. I just thought it was some quirk based on the fact that 
in some places there were quotes. So for now lets just leave it as-is :-)

PR #888, opened 2018-11-13
Remove 200ms delay in +run-builds in ford.

eglaysher:
In +run-builds in ford, the ordering of ~% ... |- expands to ~% ... =< $ |., which is wrong and causes a 200ms delay for reasons I suspect are related to battery registration. This supersedes the previous #881 which had multiple other changes and was based off a dead branch; this is the minimal case.

review: @ixv, @joemfb 
cc: @belisarius222 
Closes #881


joemfb:
This absolutely flies!

PR #704, opened 2018-04-06
[test] #703: List addressing bug, &33:~ bails

ohAitch:
can't repro the hanging one actually.

@cgyarvin do we have any good test harness for libnoun that doesn't involve booting a pill?

Issue #703, opened 2018-04-06
List addressing bugs: &33:~ bails out


mikolajpp:
&33:~ bails out with out of loom error.

Calling 
`&33:~[[%leaf p="syntax error"] [%leaf p="\{1 11}"]]` hangs urbit until CTRL+C wherein it prints

```
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[431 9].[443 11]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[431 13].[431 31]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[431 17].[431 31]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[475 7].[485 15]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[478 13].[485 9]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[478 17].[478 40]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[478 21].[478 40]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[1.290 7].[1.294 9]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[1.290 11].[1.290 54]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[1.290 15].[1.290 54]>
  /~sogdus-sibmep-ponmep-litsem/home/~2018.4.2..13.59.16..4850/sys/vane/ford
  <[1.290 24].[1.290 52]>
-find.+17,179,869,182
```

ohAitch:
Geez, nice. What's the !,*twig of that? The != ?

ohAitch:
Oh right &33 as in +(32), somewhere the nock interpreter isn't checking for
indirect atoms presumably.

mikolajpp:
[...]

PR #889, opened 2018-11-14
tests list-addressing bail

joemfb:
forwardports #704


tests/bug/gh-703.hoon
```diff
@@ -0,0 +1,14 @@
...
+::
+/+  *test
+|%
+::
+++  test-hang
+  %-  expect-fail
+    |.  &33:~[[%leaf p="syntax error"] [%leaf p="\{1 11}"]]
+::
+++  test-vere-bail
+  ~&  [`path`% running-unsafe+"bails out with out of loom error"]
```
eglaysher:
```
`path`%
```

this is rather ugly; i assume this is to put the path into the stack trace for when vere itself bails, but if vere bails, the sigpam is unlikely to trigger, no?

ohAitch:
Do we still expect hangs/bails here, if the relevant c issue got fixed?

tests/bug/gh-703.hoon
joemfb: That is ugly -- I meant to remove it. Already removed from the "hang" test case.

ohAitch:
Yep, I would just rename them to "expect-no-hang","expect-no-bail" or sth :P

cgyarvin:
It should just fail

PR #890, opened 2018-11-15
Added support for uppercase letters in sail attributes

baudtack:


ohAitch:
Curious what use-case came up for this? Random capital letters in bare
words are decidedly un-hoon-like, and sail is meant primarily for HTML
which is case-insensitive in attribute names.

belisarius222:
I'm declining this PR since we've decided that Sail will be used exclusively for HTML5 generation, and HTML5 attributes are case-insensitive.

belisarius222:
I am now reversing my earlier decision and accepting this PR, since we need mixed-case attribute names in Sail to produce SVG.

PR #894, opened 2018-11-25
Increase the speed of typing

eglaysher:
A large part of the delay in typing was spent in the kernel where it was scanning the +wa cache because the giant type nouns were reconstructed each time; the VM underneath couldn't do a simple pointer check. Now we use the caching version of +slot in +slur.

In addition, we also were using the compiler in a silly way in +ap-find in Gall, where we were generating arm names and seeing if they existed. Each time we tried to fire that arm. This was 10% of runtime. We cache this value now.

### DIFF
--- a/sys/arvo.hoon
+++ b/sys/arvo.hoon
@@ -158,7 +158,7 @@
     ++  slur                                            ::  call gate on
       |=  {gat/vase hil/mill}
       ^-  (unit (pair vase worm))
-      =+  sam=(slot 6 gat)
+      =^  sam  p.sew  (~(slot wa p.sew) 6 gat)
       =+  ^=  hig
         ?-  -.hil
           %&  (~(nest wa p.sew) p.sam p.p.hil)
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -1,4 +1,3 @@
-!:
 ::  clay (4c), revision control
 ::
 ::  This is split in three top-level sections:  structure definitions, main
--- a/sys/vane/dill.hoon
+++ b/sys/vane/dill.hoon
@@ -1,4 +1,3 @@
-!:
 ::
 ::  dill (4d), terminal handling
 ::
--- a/sys/vane/eyre.hoon
+++ b/sys/vane/eyre.hoon
@@ -1,7 +1,6 @@
 ::  ::  %eyre, http servant
 !?  164
 ::::
-!:
 |=  pit/vase
 =,  eyre
 =,  wired
--- a/sys/vane/gall.hoon
+++ b/sys/vane/gall.hoon
@@ -1,4 +1,3 @@
-!:
 ::  ::  %gall, agent execution  
 !?  163
 ::::
--- a/sys/vane/gall.hoon
+++ b/sys/vane/gall.hoon
@@ -74,6 +74,7 @@
 ++  seat                                                ::  agent state
   $:  misvale/misvale-data                              ::  bad reqs
       vel/worm                                          ::  cache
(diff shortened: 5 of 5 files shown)


eglaysher:
(When reviewing, could you patch this in locally to see if I'm just hallucinating the speed increase?)

joemfb:
I tested this locally and typing feels incrementally faster. (FWIW, it had already gotten significantly faster with the %ford hint fix.)

Timing the test suite with `:time _* +test /` didn't show a consistent difference, but that's probably not a great example. I added some timing printfs to `:test [%cores /]` and saw improvements, but the results were inconsistent from one run to the next. I'm not sure how to get actual comparable numbers for event timings.

eglaysher:
I'm going to wait until Monday to commit this, just to make sure everyone who wants to comment sees it first.

You might also want to combine testing this with urbit/urbit#1073, since the allocator also was a big source of latency inside arvo between card dispatch.

joemfb:
That's definitely the fastest combination. But I'm still looking for an appropriate and consistent timing test to compare across these branches.

ohAitch:
How granular are the printf timing annotations under -DGHETTO, and could
this be combined with |verb to compare the bootup sequence on a per-card
basis?

joemfb:
I added `#define GHETTO` to `all.h` and booted a fake-zod in three different configurations: 1) release-candidate Arvo and vere, 2) release-candidate vere with this branch, and 3) urbit/urbit#1073 with this branch. Output (and a `diff3` of the files is attached.

- [time1.txt](https://github.com/urbit/arvo/files/2613702/time1.txt)
- [time2.txt](https://github.com/urbit/arvo/files/2613703/time2.txt)
- [time3.txt](https://github.com/urbit/arvo/files/2613704/time3.txt)
- [time-compare.txt](https://github.com/urbit/arvo/files/2613701/time-compare.txt)

TLDR, number 3 is the fastest, but not for everything. This still isn't an idea test (it'd be better to have a greater variety of events and more granular timing), but it's interesting all the same.

ohAitch:
Can you +solid a lac=| pill? This seems to be on UNIX event granularity and
not |verb granularity.

(Also several runs under each might be useful to get an idea of variance on
any given vere, but idk if those are easy to summarize concisely)

joemfb:
Sure. "before" is release-candidate vere and arvo, "after" is urbit/urbit#1073 and this branch, and "compare" is a diff:

- [time-verb-before.txt](https://github.com/urbit/arvo/files/2613921/time-verb-before.txt)
- [time-verb-after.txt](https://github.com/urbit/arvo/files/2613920/time-verb-after.txt)
- [time-verb-compare.txt](https://github.com/urbit/arvo/files/2613922/time-verb-compare.txt)

ohAitch:
A `--word-diff` might be easier to parse here: [time-verb-compare-word.txt](https://github.com/urbit/arvo/files/2613970/time-verb-compare-word.txt)
This does look like improvements across the board!

Possibly-final fiddling request: it looks like `DGHETTO` [consolidates timings <10ms](https://github.com/urbit/urbit/blob/7672028/noun/trace.c#L55)? Which I can see being useful in debugging(esp if printing takes ~10ms), but makes the relevant prints harder to compare directly, due to minor fluctuations changing which of the lines end up bearing the full timestamps.

ohAitch:
I suppose an adjacent possibility is measuring from boot time, or the beginning of the event, or sth: this can be done in post-processing obviously, but might be easier to eyeball changes in. (somebody with a stats major [help us budget](https://twitter.com/dril/status/384408932061417472?lang=en) our compute)

jtobin:
> somebody with a stats major help us budget our compute

PR #895, opened 2018-11-26
updates %ames to use %behn, %behn to use effects

joemfb:
The `+doze` arms of %ames and %behn used to be called after every event (more or less), and the I/O drivers for each kept their own timer. Now, %ames explicitly sets %behn timers, and %behn explicitly gives `%doze` effects to set/update/cancel the next timer.

This effectively removes `+keep`, as all the vane `+doze` arms now produce `~`.  I'd be happy to actually remove `+doze`/`+keep` from the vanes and arvo, if that is so desired.

This approach may be slightly slower than I/O polling, as `+doze` was a stateless read (effectively a simpler `+scry`) and we know pay the internal Arvo move overhead for timers (although #894 should help). I expect it will scale better (as ships process more moves and communicate with more ships), as the timers are only touched if they actually need to change, instead of being checked on every event.


joemfb:
pairs with urbit/urbit#1074

joemfb:
It's worth noting that %ames could set %behn timers more efficiently. This just traverses our map of neighbors as `+doze` did after every move that results in a call to one of the three arms (`+wack`, `+whap`, or `+bick`) that updates the `rtn` (next timeout, based on `rtt`) for one of our neighbors. Instead, the smallest RTN should probably be threaded through the %ames state machines that call those arms, which would make the traversal unnecessary.

PR #896, opened 2018-11-26
Additional caching in the arvo kernel

eglaysher:
The +slym on each vane activation and each invocation of +slur should cache the type calculation.

(Over the shoulder coded with Ted.)


belisarius222:
@philipcmonk We'd like your review on this when you get a chance, mostly to double-check that we didn't introduce any space leaks by using +wa more.

eglaysher:
@philipcmonk friendly ping

PR #898, opened 2018-11-26
removes +keep from Arvo

joemfb:
updating hardcoded axes in `+solid` and `|reset`


joemfb:
pairs with urbit/urbit#1075

gen/solid.hoon
```diff
@@ -60,7 +60,9 @@
 =+  txt=.^(@ %cx (weld pax `path`[%hoon ~]))
 =+  sam=[now `ovum`[[%gold ~] [%veer p.i.vay pax txt]]]
 ~&  [%solid-veer i.vay]
-=+  gat=.*(all .*(all [0 42]))
+::  +poke at +10
```
belisarius222: I think all these invocations should be Nock 9s. Feel free to tell me this is out of scope, but it seems like it might be a good time to fix it. If we ever want to jet parts of the Arvo kernel, we'll need to use Nock 9s instead of this form.
```
=+  gat=.*(all [9 10 0 1])
```

joemfb:
`+keep` is obsolete since timers now use effects (as of urbit/arvo#895)

gen/solid.hoon
belisarius222: Hmm, maybe I was wrong about jetting. This still feels like a Nock 9 to me, though.

joemfb: I agree that these should be 9s. I have more updates to `+solid` planned, as I intend to get release-candidate arvo dual-booting on release-candidate and cc-release vere.

PR #900, opened 2018-11-27
dual boot with cc-release

joemfb:
The PR restores the formal Arvo interface gate (equivalent to `+poke`, and expected by cc-release), and updates `+solid` and `|reset` to with/around that interface. It also improves error handling in arvo and some vanes (fixing issues that come up when booting cc-release). And it adds `%zuse` to the `+ivory` pill, and all of (current) userspace to the `+brass` pill.

I have a branch of vere that merges `cc-release` and `release-candidate`, and can boot fake galaxies off of it. I think that's as far as we can get in a dual-boot scenario, without significantly complicating vane initialization or the boot sequence. I have some more cleanup and work to do on that merged branch, than I'll push it. Since the changes in this PR are necessary and compatible, I'm opening it early to avoid potential regressions or future conflicts.

### DIFF
--- a/gen/ivory.hoon
+++ b/gen/ivory.hoon
@@ -39,21 +39,21 @@
 =+  whole-formula=q:(~(mint ut %noun) %noun whole-hoon)
 ~&  %ivory-compiled
 ::
-::  tang: standard library (%zuse) installation event
+::  zuse-ovo: standard library installation event
 ::
-=/  tang
+=/  zuse-ovo=ovum
   =/  pax  (weld sys /zuse)
   =/  txt  .^(@ %cx (weld pax /hoon))
-  `ovum`[/vane/zuse [%veer %$ pax txt]]
+  [/vane/zuse [%veer %$ pax txt]]
 ::
-::  installed: kernel with tang installed
+::  installed: Arvo gate (formal instance) with %zuse installed
 ::
-~&  %tang-installing
+~&  %zuse-installing
 =/  installed
-  .*(0 [%7 whole-formula [%9 2 %10 [6 %1 now tang] %0 1]])
-~&  %tang-installed
+  .*(0 [%7 whole-formula [%9 2 %10 [6 %1 now zuse-ovo] %0 1]])
+~&  %zuse-installed
 ::
-::  produce both whole-formula (for jet registration)
-::  and a kernel with the standard library already installed
+::  produce a formula that evaluates :whole-formula (for jet registration),
+::  then ignores that result and produces :installed
 ::
 [%7 whole-formula %1 installed]
--- a/gen/ivory.hoon
+++ b/gen/ivory.hoon
@@ -39,4 +39,21 @@
 =+  whole-formula=q:(~(mint ut %noun) %noun whole-hoon)
 ~&  %ivory-compiled
 ::
-whole-formula
+::  tang: standard library (%zuse) installation event
+::
+=/  tang
+  =/  pax  (weld sys /zuse)
+  =/  txt  .^(@ %cx (weld pax /hoon))
+  `ovum`[/vane/zuse [%veer %$ pax txt]]
+::
+::  installed: kernel with tang installed
+::
+~&  %tang-installing
+=/  installed
+  .*(0 [%7 whole-formula [%9 2 %10 [6 %1 now tang] %0 1]])
+~&  %tang-installed
+::
+::  produce both whole-formula (for jet registration)
+::  and a kernel with the standard library already installed
+::
+[%7 whole-formula %1 installed]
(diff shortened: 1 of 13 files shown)


gen/ivory.hoon
```diff
@@ -39,4 +39,21 @@
...
+::  tang: standard library (%zuse) installation event
+::
+=/  tang
+  =/  pax  (weld sys /zuse)
+  =/  txt  .^(@ %cx (weld pax /hoon))
+  `ovum`[/vane/zuse [%veer %$ pax txt]]
+::
+::  installed: kernel with tang installed
+::
+~&  %tang-installing
```
belisarius222: Can we please rename this to %zuse?

gen/solid.hoon
```diff
@@ -10,59 +10,85 @@
 ::::
   !:
 :-  %say
-|=  $:  {now/@da eny/@uvJ bec/beak}
-        {arg/$@(~ {top/path ~}) dub/_|}
+|=  $:  [now=@da eny=@uvJ bec=beak]
+        arg=$@(~ [top=path ~])
+        dub/_|
```
belisarius222: `dub=_|` if we're modernizing syntax

cgyarvin:
Wow.  It amazes me that a dual-boot kernel can even work this well.  The sooner we go fully over, the better.

gen/solid.hoon
```diff
@@ -10,59 +10,85 @@
...
-=+  rax=.^(@t %cx (weld arp `path`[%hoon ~]))
-=+  ^=  ken
-    =-  ?:(?=(%& -.res) p.res (mean (flop p.res)))
-    ^=  res  %-  mule  |.
-    ~&  %solid-loaded
-    =+  gen=(rain pax txt)
+=/  compiler-src  .^(@t %cx (weld compiler-path /hoon))
+=/  arvo-src      .^(@t %cx (weld arvo-path /hoon))
+=/  arvo-formula
+  =;  res
```
belisarius222: I know it was already this way, but why do we bother calling +mule here? All we do with the error is print and die, which I would assume is the same as what would happen if we ran it without an extra virtualization layer.

gen/solid.hoon
```diff
@@ -10,59 +10,85 @@
...
+  |=  [ovo=ovum ken=*]
+  [~ .*(ken [%9 2 %10 [6 %1 now ovo] %0 1])]
+::  produce both the arvo-formula (for jet registration)
+::  and a kernel with the vanes already installed
 ::
-=+  gat=.*(all .*(all [0 10]))
-=+  nex=+:.*([-.gat [sam +>.gat]] -.gat)
-$(vay t.vay, all nex)
-
+[arvo-formula .*(installed [%0 7])]
```
belisarius222: I assume there's a good reason, but why do we grab the context from the 'installed' kernel?

belisarius222: Is it to dig the kernel core out of the +poke gate wrapper?

joemfb: `installed` is the formal Arvo interface, ie the anonymous poke wrapper gate, but `release-candidate` expects to be the structural Arvo interface.

joemfb: There should probably be a comment here.

sys/arvo.hoon
```diff
@@ -430,10 +434,11 @@
...
         $give
-      ?>  ?=(^ q.gum)
-      ?.  ?=(^ i.q.gum)
-        ~&  [%jack-bad-duct q.gum]
-        ~&  [%jack-bad-card +>-.p.r.gum]
+      ?.  ?&  ?=(^ q.gum)
+              ?=(^ i.q.gum)
+          ==
+        ~|  [%jack-bad-duct q.gum]
+        ~|  [%jack-bad-card p.gum (@tas +>-.p.r.gum)]
```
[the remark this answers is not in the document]
joemfb: That's fair, I'll add the check

gen/solid.hoon
joemfb: Agreed, there's no reason to mule here.

gen/solid.hoon
joemfb: thanks

gen/ivory.hoon
joemfb: sure

joemfb:
@belisarius222, I locally enabled and ran `+disabled-test-reef-slow` in the %ford test suite and saw jet registration errors. The test passed though, so I'm assuming both the test and %ford's reef logic are now wrong. I'll look closer.

joemfb:
It turns out that, since `%reef` is compiled against `+is`, the addition of the formal Arvo gate doesn't make a difference to %ford. I took the liberty of updating `%reef` to compiled against `..ride`, and enabled the full test. Seems like that should be on in CI, since it's the closest thing to a `|reset` in the testsuite.

belisarius222:
These changes all LGTM.

PR #901, opened 2018-11-28
restores the reproducibility of the +ivory pill

joemfb:
Since `+ivory` exists to be compiled into the urbit binary, it's very important that the pill be perfectly reproducible. My changes in #900 were functionally correct, but not reproducible -- the real date was passed in the %zuse installation event (along with the full path to %zuse, which also includes the real date). This PR corrects both issues. I also took the importunity to modernize the style and improve the comments a little more.

### DIFF
--- a/gen/ivory.hoon
+++ b/gen/ivory.hoon
@@ -1,56 +1,61 @@
 ::
-::::  /hoon/metal/gen
+::::  /hoon/ivory/gen
   ::
 /?    310
 ::
 ::::
   !:
 :-  %say
-|=  $:  {now/@da * bec/beak}
-        *
-    ==
+|=  [[now=@da * bec=beak] *]
 :-  %noun
-::
 ::  sys: root path to boot system, `/~me/[desk]/now/sys`
 ::
-=+  sys=`path`/(scot %p p.bec)/[q.bec]/(scot %da now)/sys
-::
+=/  sys=path
+  /(scot %p p.bec)/[q.bec]/(scot %da now)/sys
 ::  compiler-source: hoon source file producing compiler, `sys/hoon`
 ::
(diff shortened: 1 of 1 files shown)


gen/ivory.hoon
```diff
@@ -1,56 +1,61 @@
...
+  q:(~(mint ut %noun) %noun whole-hoon)
 ~&  %ivory-compiled
-::
 ::  zuse-ovo: standard library installation event
 ::
-=/  zuse-ovo=ovum
-  =/  pax  (weld sys /zuse)
-  =/  txt  .^(@ %cx (weld pax /hoon))
-  [/vane/zuse [%veer %$ pax txt]]
+::    Arvo parses the %veer payload with +rain, so
```
belisarius222: It would be better to pass the spur than the empty path. Otherwise if an error occurs, we won't know which file errored.

gen/ivory.hoon
```diff
@@ -1,56 +1,61 @@
...
-::
+=/  compiler-source
+  .^(@t %cx (welp sys /hoon/hoon))
 ::  compiler-hoon: compiler as hoon expression
 ::
+::    We parse with +ream for reproducibility.
+::
 ~&  %ivory-parsing
-=+  compiler-hoon=(ream compiler-source)
+=/  compiler-hoon  (ream compiler-source)
```
belisarius222: I think this should be `(rain /hoon/hoon compiler-source)`.

gen/ivory.hoon
joemfb: The spur, flopped? `/sys/zuse/hoon`?

gen/ivory.hoon
joemfb: Arvo too?

Issue #457, opened 2017-11-22
^& only useful for introducing type holes


ohAitch:
The "covariance" feature is not used, has no obvious purpose, and can create a type-value mismatch

```
> !>((^+(^&(|=(@ %~)) |=($~ +<)) |))
[p=#t/$~ q=1]
```

joshuareagan:
It's worth pointing out that the same holds of `^?` (lead gate conversion):

```
> !>((^+(^?(|=(@ %~)) |=(%~ +<)) |))
[#t/$~ q=1]
```

joshuareagan:
This is now fixed.

joshuareagan:
Depends on what you mean by "become used"  -- for actual code that people care about?  No.  For toy examples to teach Hoon?  yes.  

Imagine you want to think of a core as being an object like in object-oriented programming.  Zinc allows you to create a core that can serve as a prototype for a "class" of such objects, and cast appropriately.  I've got a tutorial draft on this with some examples.  Sadly the examples are slightly mistaken, but they can be fixed up.  (The mistakes exploit another variance bug that I just submitted a PR for.)

joshuareagan:
I meant to link the tutorial draft:

https://github.com/joshuareagan/doc-drafts/blob/master/polymorphism.md

cgyarvin:
Basically this feature, %zinc, was left as an exercise for someone fearless.

There is certainly a general algebra more powerful than the four-metal
model.  In theory, we could apply a mask that was read-only, write-only or
opaque on top of any type.  This is also a simpler data structure in some
ways.  My concern is just that it is more powerful than it needs to be.

It's clear that covariance is theoretically useful in an OO style of
coding.  Let's actually see it used in practice before we think about
stepping past the four-metal model.  The lack of other four-letter metals
is telling you something!

PR #902, opened 2018-11-29
 variance fix: calling lead, zinc gates

joshuareagan:
You shouldn't be able to call zinc or lead gates (or resolve with changes to an arm in such a core).  This change enforces that.


joemfb:
In order to pass CI, these changes will need to be staged more granularly. You need to be able to +solid and run a pill on a vere without the new enforcement rules in the jet, and then boot from that same pill with a vere that has the new enforcement rules.

The variance enforcement is changing for %lead, but not for %iron, correct? If so, you want to first change the appropriate cores to use %iron instead of %lead, and then change the enforcement rules.

joshuareagan:
How do you know that's the particular change that needs to be made?

joshuareagan:
I mean, re: lead vs iron cores?

joemfb:
On second thought, this might be able to succeed, but only through cheating with jet mismatches. I've restarted the build -- we'll see.

joshuareagan:
fwiw I'm not trying to be skeptical, I'm just trying to understand the process a little better

joemfb:
I'm simply stating this: you're tightening type-system constraints. When you make such a backwards-incompatible change, you first need to make sure everything will work with those tightened constraints, then actually enforce them. In this case, you were able to +solid and a pill from 0eaca39 only because it was run with a vere that did not have the new constraints in the compiler jet, in other words, by exploiting a jet mismatch. The build for f583936 failed because the 0eaca39 pill had not yet been uploaded.

joshuareagan:
https://github.com/urbit/urbit/pull/1078

joemfb:
This is not properly staged, but it appears that it will work anyway. As long as both of these latest builds succeed, I'm not attached to restaging the changes. I will ask that you update the URL in `.travis/pin-parent-pier-url.txt` to include the full hash of f583936, if and when it completes successfully.

joemfb:
That's interesting. I guess it makes sense if compiling the kernel doesn't actually involve any calls into any of the cores that previously (and incorrectly) were %lead. I assumed at least some of the vanes would have such calls, but a quick search doesn't turn anything up.

joshuareagan:
That makes sense, but yeah I don't think those cores get their samples written-to in that process.

In your defense though it seems like CI needs a kick in order to accept my changes...  maybe it's something else?

ohAitch:
I believe you said this wasn't broken on master? (or well, the lead part,
I'm not sure there *are* any zinc cores in the system)

joemfb:
I think you have the wrong commit in the pier url -- it should be `f583936e8e092a133644d184387e8bb7d30c9299`.

joshuareagan:
Sorry about that @joemfb, should be fixed now.

@ohAitch that's a different variance bug (we've discovered a few recently!).  On master you can call zinc and lead gates, which is how you could abuse zinc gates in your bug report last year: https://github.com/urbit/arvo/issues/457

joemfb:
Thanks!

One last staging thing: can you undo my mistake that I mentioned in urbit/urbit#1078 by editing `tests/sys/vane/ford.hoon` and renaming `+test-reef-slow` to `+disabled-test-reef-slow`? After that builds successfully here, you can update the Arvo hash in urbit/urbit#1078, and then it will build in a normal time frame. Sorry!

joshuareagan:
Ok, change made to `ford.hoon`.  I'll update the Arvo commit hash as soon as we get a build here.

benjamin-tlon:
I'm still :eyes: this.

PR #905, opened 2018-12-03
Prime the worm cache in arvo with scry types

eglaysher:
So. This trace is illuminating about why Ames is slow.

why-is-ames-slow

The majority of the time in Ames is spent either parsing things (`+nuck:so`) or inside the compiler before/after the scry to Jael.

This fixes the scry part, so that scry's don't invoke the compiler. The majority of time in handling a scry is spent in the compiler since it doesn't use the compiler cache since scrying can't have side-effects and modify the cache. So prime the worm cache at vane install time so the cache is prepopulated before we call a scry.

I'm not sure if this is the best way to structure this patch, so please read it critically, but this does shave >10ms per Ames call which scrys into Jael. That said, I think I've tested every case of `|reset`/`+solid`/vane load, and have confirmed with printfs that the compiler usage in `+scry` is now a cache hit.

### DIFF
--- a/sys/arvo.hoon
+++ b/sys/arvo.hoon
@@ -65,6 +65,8 @@
 ++  ovum  {p/wire q/curd}                               ::  typeless ovum
 ++  pane  (list {p/@tas q/vase})                        ::  kernel modules
 ++  pone  (list {p/@tas q/vise})                        ::  kernel modules old
++$  scry-sample
+  [fur=(unit (set monk)) ren=@tas why=shop syd=desk lot=coin tyl=path]
 ++  ship  @p                                            ::  network identity
 ++  sink  (trel bone ship path)                         ::  subscription
 ++  sley  $-  {* (unit (set monk)) term beam}           ::  namespace function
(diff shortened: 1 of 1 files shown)


ohAitch:
Ooh!

With the patch, the trace in question now majority nuck:so? What sorts of
coins are being parsed?

eglaysher:
It's parsing wires in a lot of places. It shows up in several Gall apps. I assume wires are what's going on in Ames, but that's on priors. What's really interesting is that `+nuck:so` is slow even when it's just parsing an individual number, like in the case of terminal effects.

Here's a trace file you can open with chrome://tracing if you want to play along at home: [w8-pristine.json.gz](https://github.com/urbit/arvo/files/2638065/w8-pristine.json.gz). This trace contains this PR and you can see that scrys are no longer the majority of Ames' runtime.

sys/arvo.hoon
```diff
@@ -282,19 +319,20 @@
...
+            (flop s.bed)
+        ==
       ^-  (unit (unit (cask)))
-      =+  pro=(slym (slap rig [%limb %scry]) old)
+      =+  fun=-:(~(slap wa worm.vane) rig [%limb %scry])
+      =+  pro=-:(~(slym wa worm.vane) fun old)
       ?~  q.pro  ~
       ?~  +.q.pro  [~ ~]
-      =+  dat=(slot 7 pro)
+      =+  dat=-:(~(slot wa worm.vane) 7 pro)
```
belisarius222: The type in the `dat` vase doesn't get used at all. We should replace this and the next line with:
```
=/  dat  +>.q.pro
[~ ~ (mark -.dat) +.dat]
```

sys/arvo.hoon
```diff
@@ -136,8 +138,43 @@
...
+      ::
+      =^  rig  worm.vane
+        (~(slym wa *worm) vase.vane *[@da @ slyd])
+      ::  cache the access of the %scry arm
+      ::
+      =^  fun  worm.vane  (~(slap wa worm.vane) rig [%limb %scry])
+      ::  cache the call to +mint that the +slym in +scry will do
+      ::
+      =^  pro  worm.vane  (~(mint wa worm.vane) p.fun [%limb %$])
+      ::  cache the call to +mint that the +slot does for the return call
```
belisarius222: We don't need to prime this cache if we replace the +slot call in +scry with untyped axis lookup (see other review comment).

cgyarvin:
Wires should be vases.  We can’t change this right now.  But all of nuck:so should be jetted.

joemfb:
I continue to feel that vases for wires are overkill. But I think turning `+path` and `+wire` into `(list dime)` would be a better UX and would remove a lot of unnecessary parsing.

eglaysher:
Done. PTAL.

PR #906, opened 2018-12-03
Don't run nest checks on every +call.

eglaysher:
This removes an "XX temporary" from 2014 in most of the vane `+call` functions. Ford does not do anything like this, and other vanes shouldn't be calling the compiler to validate their input.

(This doesn't touch Clay because there's additional debugging in that one which I've seen trigger in the wild.)

### DIFF
--- a/sys/vane/ames.hoon
+++ b/sys/vane/ames.hoon
@@ -1227,20 +1227,15 @@
     |%                                                  ::  vane interface
     ++  call                                            ::  handle request
       |=  $:  hen=duct
-              hic=(hypo (hobo task:able))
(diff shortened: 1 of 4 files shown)


ohAitch:
This predates card types being centralized in zuse yup, kill it with fire.

(In 2014 the headers were copy-pasted to the calling vane, which did in
fact introduce type errors on a couple occasions; I'm guessing at this
point arvo is nest-checking the (hypo (hobo task)) properly anyway?)

eglaysher:
PTAL, I removed the Clay one, as well.

PR #909, opened 2018-12-04
implements common pill interface, adds userspace to +solid

joemfb:
This PR updates all our pills to use a consistent interface: a 3-tuple of boot-formulas, kernel/module events, and userspace events. (These events are technically `(list ovum)`, ie, `(list (pair wire card))`.)

This interface is basically the already existing interface to of `+brass`, except that all three elements are now lists. The boot formulas are to be evaluated with the lifecycle formula `[%2 [%0 3] %0 2]`, producing the formal Arvo interface (the outer gate). The subsequent event-lists are to be enqueued as normal events.

This is a pretty natural structure for `+brass`, which exists as a recipe for creating the True Boot Sequence. I've extended it to `+solid` and `+ivory` to lower the conceptual overhead of what exactly a pill is, and to increase developer flexibility moving forward. You can think of these various pills as occuping positions on a dial that controls how long it takes to boot a ship -- `+solid` is more-or-less already booted, `+brass` is completely proper and double-compiled, and `+ivory` is lightweight.

`+solid` has also been extended to include a full copy of userspace -- non-galaxy ships now boot into a working :dojo as quickly as galaxies, and initial syncs now take place in the background, no longer blocking userspace. (This further convergences us with cc-release, which boots from the `+brass` pill into a full Arvo.) 

There's an issue worth noting here; stars/planets/moons cannot sync a desk with their sponsor/parent unless they share a common-ancester commit (a merge-base, for %clay's 3-way diff). Booting into a full Arvo means there is no merge-base. The solution implemented here is to perform `%init` merges up to and including the first merge caused by a sync (an %init merge throws away local state and fully accepts the remote state). This works as implemented, for all classes of ships. But it's definitely awkward, and potentially error prone (if the network is slow and a user makes local edits before the first sync merge, for instance). In the long run, it would be preferrable for %clay to have some kind of internal content-addressing, or some other mechanism for converging with a remote desk without a merge-base.


joemfb:
urbit/urbit#1085 is the dual of this PR.

eglaysher:
I worry a bit about blowing away user data, even if it only happens once a minute after the user boots. This may not be a problem in practice in the short run, but I can totally see this being a real issue once we start using the filesystem for real things.

Otherwise, I don't see anything wrong with this patch; I do wonder how it feels. If I type `.p/pill +solid` in the dojo, what's the additional overhead of generating the filesystem event? If I'm booting from a pill, is there no additional overhead since we're replacing the synthetic event created with `-A` with the one in the pill?

joemfb:
I agree. I also don't like that it's a hardcoded number of revisions (different for the home desk) that marks the threshold between %init and normal merges. I don't know how to implement this differently from outside of %clay, and it seems like it'd be nontrivial to do it properly within %clay.

There aren't any substantial timing changes to first-boot or `+solid`. Adding userspace to the pill is shockingly fast (almost 400 files!).

Here's some timing examples. I didn't rigorously test each scenario multiple times, so I'm not sure what the normal variance is -- I've seen a couple seconds in either direction of these numbers. These were all `debugoptimized` builds.

First, booting a new fake-zod. Note that these exit after the first clay %into, but before the second.

`release-candidate`:

```
$ time debug/urbit -F zod -A ./arvo -s -c -x rc-zod
...
real    0m34.802s
user    0m33.217s
sys     0m0.761s
```

This PR, with a local Arvo sync:

```
$ time debug/urbit -F zod -A ./arvo -s -c -x zzzod
...
real    0m37.553s
user    0m35.617s
sys     0m0.858s
```

This PR, syncing from the pill:

```
time debug/urbit -F zod -u https://bootstrap.urbit.org/git-51ca94d192.pill -c -x zzzod
...
real    0m38.452s
user    0m36.520s
sys     0m0.842s
```

Soliding a pill on release-candidate:

```
> :time _* +solid
>=
[%took ~m1.s28..a737]
```

Soliding a pill on this PR:

```
> :time _* +solid
>=
[%took ~m1.s34..9077]
```

joemfb:
I would guess that the boot overhead is in downloading an 8 MB pill. I expected a slight speedup retrieving the filesystem sync from the pill instead of constructing it from disk, but maybe that's also offset by the time it takes to read the pill into an atom (twice, unfortunately, due to the vagaries of the startup process).

joemfb:
My comment was vague, but so's my idea ... I mean something like "%clay should use content-addressing of files and/or commits to find something like a merge-base". I haven't thought through what our options are in the case that there is no true merge-base (ships booted from different pills, no one commit in either's history is the same).

There's also the somewhat related need to reduce network traffic, see #551.

gen/solid.hoon
```diff
@@ -56,38 +57,31 @@
...
   [~ .*(ken [%9 2 %10 [6 %1 now ovo] %0 1])]
 ::
-::  produce a pair of :arvo-formula (for jet registration) and
-::  the Arvo core (structural interface), which is the context
-::  of the formal Arvo interface gate :installed
+::  our boot-ova is a list containing one massive formula:
 ::
-[arvo-formula .*(installed [%0 7])]
+::    We evaluate :arvo-formula (for jet registration),
+::    then ignore the result and produces :installed
```
belisarius222: produces -> produce

philipcmonk:
I like @belisarius222's idea of blocking any filesystem changes until the parent sync has finished.

I forget if it was ever properly implemented, but in theory `%clay` should ask for the parent's yaki (or maybe the whole ankh), check which lobes it doesn't already have, and ask for only those.  In this case, syncing the parent's data should be very fast if its files are pretty similar to what's in the pill.

If that is implemented, another option is to not initialize a child with a full filesystem, but only with `lat.ran` (ie sync the blob store but not the commits), so the sync from the parent is genuinely the first commit, but it should be quite fast.

ohAitch:
Long-term, presumably we want the not-O-all-files thing of having directory
entries also in the blob store, as git does it; in which case you would be
able to just compare the root directory hashes.

lib/hood/kiln.hoon
```diff
@@ -388,33 +381,39 @@
...
-        [%kiln %sync syd (scot %p her) sud ?:(reset /reset /)]
-        our  syd  her  sud  ud+let
-        ?:  =(0 ud:.^(cass:clay %cw /(scot %p our)/[syd]/(scot %da now)))
-          %init
-        %mate
-    ==
+    =/  =wire  /kiln/sync/[syd]/(scot %p her)/[sud]
+    =/  =cass  .^(cass:clay %cw /(scot %p our)/[syd]/(scot %da now))
+    ::
+    ::    If we will be syncing in remote changes, we need all our sync merges
```
belisarius222: This is pretty gross, and I don't completely get it. It seems like it might be Clay's problem more than a boot problem, though.

lib/hood/kiln.hoon
```diff
@@ -388,33 +381,39 @@
     ==  ==
   ::
   ++  start-sync
-    |=  reset/?
```
belisarius222: Why do we no longer need to know whether we're resetting?

lib/pill.hoon
```diff
@@ -0,0 +1,122 @@
...
+::
+|%
+::  +module-ova: vane load operations.
+::
+::    sys: full path to /sys directory
+::
+++  module-ova
+  |=  sys=path
+  ^-  (list ovum)
+  |^  :~  ::  sys/zuse: standard library
```
belisarius222: This looks like it should just be a +turn.

lib/pill.hoon
```diff
@@ -0,0 +1,122 @@
...
+  ::    /lib    %ford libraries
+  ::    /mar    %ford marks
+  ::    /ren    %ford renderers
+  ::    /sec    %eyre security drivers
+  ::    /sur    %ford structures
+  ::    /sys    system files
+  ::    /tests  unit tests
+  ::    /web    %eyre web content
+  ::
+  %.  [/app /gen /lib /mar /ren /sec /sur /sys /tests /web ~]
```
belisarius222: Is there a reason we enumerate these instead of just doing a `.^(arch %cy path)`?

It might also be worth considering using the `%ct` Clay scry to get a full recursive file listing instead of manually recursing here.

lib/hood/kiln.hoon
joemfb: The only time we were resetting was on the first sync of a star, planet, or moon. I believe we were unconditionally resetting because those ships were booting into kernel without a filesystem and then syncing from their parent. Ie, as soon as they had a full filesystem they did a reset. That's no longer necessary since we're booting into a coherent filesystem that matches the kernel (unless you're freelancing with `-A`).

belisarius222: Sounds good.

lib/hood/kiln.hoon
joemfb: This is indeed gross, and the heart of the problem. Clay doesn't know about syncs, they're a userspace feature implemented here. Maybe clay should have a merge mode that starts normally (as a `%mate` or whatever), but escalates all the way up to %init. It'd still have to be triggered from somewhere, either here in kiln or in the init logic of %dill.

Similarly, I don't know how we would block user edits until a sync succeeds; %clay doesn't know about syncs. It wouldn't be enough even to block user edits until a merge has succeeded once on every desk for all non-galaxy ships -- %home and %kids are created through merges, and receive the first two syncs from the pill or arvo directory as merges from %base. It seems like we're forced to hardcode a threshold like this, unless/until we have a smarter merge type that can synthesize a merge-base.

lib/pill.hoon
joemfb: Sure, that seems appropriate. `|^` is usually kind of weird.

lib/pill.hoon
joemfb: This is just copying around the original implementation for cc-release pills -- first in `+metal`, later in `+brass`. Maybe it was for documentary reasons the first time? I'm not opposed to refactoring and improving it, but it does work, and reasonably quickly.

belisarius222: Ok. We can clean it up in the glorious future after everything else works perfectly.

lib/hood/kiln.hoon
belisarius222: Ok, that makes a little more sense. It sounds like we'll be more or less stuck with this until we move sync from userspace into Clay itself, which seems more and more like the thing to do. I'm willing to accept the necessity of this hack for now.

What is it about receiving the userspace files from the pill instead of from the local Arvo that forces us to do this weird hardcoded logic? Current release-candidate does something during boot to get Clay up to speed that doesn't require this.

belisarius222: I've read this a few more times and have a better sense of it now. It's really brittle, but looks like it should work fine unless the user does make a commit before the sync happens. Ensuring the order of Clay transactions is a problem we'll have to solve more generally at some point anyway.

ohAitch:
@philipcmonk has some war stories back from when sync was directly in clay!
But it might be time for another go; and even if we keep the update loop in
userspace, merges themselves still need to work correctly at that level,
and right now definitely don't.

PR #911, opened 2018-12-05
Trim ethereum snapshot in Jael

belisarius222:
**NOTE: requires more testing before merge; waiting on booting a ship on the testnet for manual testing and snapshot generation**

Removes some fields from the ethereum snapshot:
- Ethereum event logs
- the state of Urbit's connection to the Ethereum node

Also removes the `live` bit from the `+public` datatype, since it wasn't used and probably belongs as Ames state.

This changes the snapshot format, but we don't need to change Vere or perform any staging. We should be able to export a snapshot from a ship running this code, then boot off that.

### DIFF
--- a/sys/vane/jael.hoon
+++ b/sys/vane/jael.hoon
@@ -701,7 +701,7 @@
         =/  cub  (nol:nu:crub:crypto key.seed.tac)
         %+  ~(put by kyz.puk.sub)
           our
-        [& lyf.seed.tac (my [lyf.seed.tac pub:ex:cub] ~)]
+        [lyf.seed.tac (my [lyf.seed.tac pub:ex:cub] ~)]
       ::  our initial private key, as a +tree of +rite
       ::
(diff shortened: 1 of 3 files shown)


sys/vane/jael.hoon
```diff
@@ -2137,11 +2133,11 @@
...
     %=    +>.$
-        eve.urb       eve.snap
-        etn           etn.snap(source source.etn)
-        kyz.puk.sub   kyz.snap
-        +.eth.sub     eth.snap
-        sap           sap(last-block 0)
+        heard.etn         heard.snap
+        latest-block.etn  latest-block.snap
+        kyz.puk.sub       kyz.snap
+        +.eth.sub         eth.snap
```
philipcmonk: Since we manually specify `[dns hul]:eth.sub` in `++extract-snap`, arguably we should refer to them specifically here as well.

belisarius222: Good call. I'll fix that.

PR #913, opened 2018-12-05
Add ~| to arvo to try to debug intermittent crash.

eglaysher:
There is some weird failure in the +wa cache code that happens when hoon.hoon gets changed. I can't repro this on demand, so I want to add a bunch of ~| to debug.

### DIFF
--- a/sys/arvo.hoon
+++ b/sys/arvo.hoon
@@ -175,7 +175,9 @@
   ::
   ++  wink                                              ::  deploy
     |=  {now/@da eny/@ ski/slyd}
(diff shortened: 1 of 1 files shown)


sys/arvo.hoon
```diff
@@ -496,6 +506,7 @@
   ++  kick                                              ::  new main loop
     |=  {lac/? mor/(list muse)}
     =|  ova/(list ovum)
+    ~|  [%failed-kick lac]
```
belisarius222: `lac` is the laconic bit, which is what gets flipped when you run `|verb`. I don't think it's worth printing. I think we should move this `~|` into `+jack', since that's the only part of this arm that could fail.

eglaysher:
PTAL

PR #914, opened 2018-12-06
fix another type variance bug

joshuareagan:
Another variance bug -- wings with arm names can be used to evade variance restrictions.

This wing resolution to a lead `sample` fails, which is the desired outcome:

```
> =b ^?  |=(sample=@ 15)

> sample.b
-find.sample.b
```

But if we use the `$` arm in the wing it goes through (unsafe):

```
> sample.$.b
0

> ? sample.$.b
  @
0
```

This PR squishes the bug.


joshuareagan:
This PR corresponds to: https://github.com/urbit/urbit/pull/1089

sys/hoon.hoon
```diff
@@ -9765,10 +9765,13 @@
...
-      %|  %-  fire
+      %|  %-  fork
           %+  turn  ~(tap in q.lap)
-          |=  {a/type b/foot}
-          [a [%dry %$ 1]]
+          |=  [a=type b=foot]
+          :-  %hold
+          ?>  ?=([%core *] a)
+          =+  dox=[%core q.q.a q.a]
+          [dox [%$ 1]]
```
joemfb: I think it could be more clear that is throwing away the foot. The sample of your gate could be `[a=type foot]`, or even `[a=type *]`. No strong feeling on this, it just took me a minute to realize that's what's happening.

joemfb: Thanks, I think this is clearer.

sys/hoon.hoon
```diff
@@ -9765,10 +9765,13 @@
     ^-  type
     ?-  -.lap
       %&  p.lap
-      %|  %-  fire
+      %|  %-  fork
           %+  turn  ~(tap in q.lap)
-          |=  {a/type b/foot}
-          [a [%dry %$ 1]]
+          |=  [a=type *]
+          :-  %hold
```
joemfb: Is there any reason this has to be a `%hold`. Is there any distinction between expanding `[%hold [%core ...] %$ 1]` and just producing the `%core` itself?

joshuareagan: No, I don't think so.  I only left it in out of conservatism (e.g., matching the pattern in +fire), but after testing it a bit there doesn't seem to be any reason to lose it.

sys/hoon.hoon
```diff
@@ -9765,10 +9765,11 @@
...
     ?-  -.lap
       %&  p.lap
-      %|  %-  fire
+      %|  %-  fork
           %+  turn  ~(tap in q.lap)
-          |=  {a/type b/foot}
-          [a [%dry %$ 1]]
+          |=  [a=type *]
+          ?>  ?=([%core *] a)
+          [%core q.q.a q.a]
```
joemfb: I know I should know this, but what's the difference between the type in the %core and the type in the `+coil`? How do `p.a` and `q.q.a` differ here?

joshuareagan: This is a good question and I was thinking about this too.  q.q is the original payload type and p is the current payload type.  Here I was again mimicking +fire too closely (re: firing a dry arm).  But I think I can just use the original type, unmodified.  Give me a second to test this...

joshuareagan: When I have +felt return `a` as the unmodified type I get a nest-fail on rebuild (for clay).  Something depends upon using the q.q type for wing resolution...

This issue will be revisited in the future (as is often said, the compiler needs some rewriting), so I suggest setting aside this simplification for the moment.

joemfb: Thanks -- I was just curious.

PR #915, opened 2018-12-06
single-home arvo

joemfb:
This PR updates Arvo to have a single, persistent identity for its entire lifetime. At long last. To limit the ambiguity of a kernel that exists before it learns who it is, I've added an outer larval core, corresponding to the stage of the boot sequence in which we break symmetry. In other words, we boot into a baby kernel that learns who it is and imbibes of the randomness of the world, and then upgrades itself to a full-grown kernel -- the true Arvo that knows itself always.

To accommodate a variety of boot sequences (see #909), the larval kernel also handles %zuse and vane installations, only upgrading itself when it has identity, entropy, and an installation of the standard library.

I've also restored the new (old) reboot implementation `+velo`. It currently exists side-by-side with the old reboot `+vega`, and is not wired into `|reset`. I expect we will want to do so, but would like to see it reviewed first.

Finally, I've trivially single-homed the vanes, by adding identity to their sample (which is now `[ship time entropy namespace]`). I've deferred the work of single-homing them in their implementations, which will range from trivial (`%behn`) to quite invasive (`%ames`, `%jael`).

I'm eager to hear your reactions to this approach.


joemfb:
This is the dual of urbit/urbit#1090

sys/arvo.hoon
```diff
@@ -683,6 +689,83 @@
...
+  ^-  (unit (pair (list ovum) *))
+  ::  virtualize; dump error if we fail
+  ::
+  =-  ?:(?=(%| -.res) ((slog p.res) ~) `p.res)
+  ^=  res  %-  mule  |.
+  ::  produce a new kernel and an effect list
+  ::
+  ^-  (pair (list ovum) *)
+  ::  compile the hoon.hoon source with the current compiler
+  ::
```
eglaysher: Philosophically, should this process happen in the kernel?

If this was to happen in user space, we could do all sorts of validation before userspace handed arvo its new self.

joemfb: It's an interesting question -- to what degree is it appropriate for Arvo proper to defer to its modules? [neo arvo](https://github.com/cgyarvin/arvo/blob/remaint-merge/neo/arvo.hoon) (which will need a closer look and broader discussion at some point in the near future) goes in the opposite direction, giving Arvo a miniature filesystem and additional independence from the vanes. I'm personally torn, I like the idea of a more independent Arvo, but is it worth the maintenance and conceptual burden? Does Arvo ever need to reset itself without having vanes and a filesystem?

For now, I think it's clear that `%velo` is superior to `%vega` (of course, I implemented neither of these ...)

/cc @cgyarvin

joemfb:
I made a mistake a separate formal interface gate for the larval stage and the full kernel. The inner interface gate was unreachable and superfluous. I've corrected that, and moved the gate back to the top of the file.

sys/arvo.hoon
belisarius222: Apologies for the long-winded comment.

The idea of Arvo having a filesystem but not preloaded state in other vanes seems a bit hacky to me. A filesystem is just an initial Clay fixture. I feel like an initial Jael snapshot and an initial Ames map from ship to IP address should go in the same place.

If we go the route of Arvo handling its own updates, we might want to give it an argument of validation checks that it can perform -- things like unit tests. Unfortunately, making that fully general is quite difficult, since we might want it to be able to run arbitrary Ford builds and do other things that require asynchronicity or remote data, and that would be pretty ugly to build into Arvo at that level.,

For that reason, I think I'd be happier taking Arvo in a more minimal direction long-term, where maybe it just receives a vase of a new vane. That way we can separate the concern of constructing that vane from the concern of installing the vane, and we can build arbitrarily complex development logic for the developer workflow without corrupting the cleanliness of Arvo itself.

There are also potential performance implications based on where the construction of new vanes happens. If it happens in userspace, it can use Ford's caching to generate kernels faster, which could save a lot of time when doing kernel and vane development. Then again, long-term I suspect we'll want a persistent compiler cache in Vere, which should get a hit when rebuilding the same vane, no matter whether that happens through Ford or otherwise.

joemfb: Sorry, I think I was unclear. That new Arvo implementation doesn't have a %clay snapshot or something of that sort. It has its own "deep storage" -- what appears to be a small filesystem implementation to back kernel reboots and vane installations.

I agree that we will want the flexibility you've described. But I'm not sure that means Arvo needs to just accept already-built vanes as vases. That sounds kind of crippled to me. Definitely more to discuss here.

sys/arvo.hoon
```diff
@@ -502,101 +515,232 @@
     $(ova (weld p.nyx ova), mor (weld q.nyx t.mor))
   --
 --
-=+  pit=`vase`!>(.)                                     ::
-=+  bud=pit                                             ::  becomes tang
-::  =+  vil=(viol p.bud)                                ::  cached reflexives
-=+  vil=(viol p.bud)                                    ::  cached reflexives
-=|  $:  lac/?                                           ::  laconic bit
-        eny/@                                           ::  entropy
+=<  ::  Arvo larval stage
```
belisarius222: My gut feeling here is that this larval core is too flexible and accommodating, and that we'll want to lock it down more later. I think we'll get a better sense for that after using this version for a while, though, so I think it's good as it is.

sys/arvo.hoon
```diff
@@ -502,101 +515,232 @@
...
+            eny=(unit @)
+            bod=(unit vase)
+        ==
+    ::  larval Arvo structural interface
+    ::
+    |%
+    ++  come  ^come                                     ::  22
+    ++  load  ^load                                     ::  46
+    ::
+    ::  XX my preference would be for +peek to no-op here
```
belisarius222: Why would we want +peek to no-op? Seems useful

joemfb: The larval stage doesn't let any events through to the vanes. Are there useful scrys into vanes with default state?

sys/arvo.hoon
```diff
@@ -502,101 +515,232 @@
...
+                  ::
+                  ~&  [%larval-ignore p.ovo -.q.ovo]
+                  +>.$
+                ::  install %zuse or vane
+                ::
+                    %veer
+                  ^+  +>.$
+                  ::  use the maximum comet if we don't know who we are yet
+                  ::
+                  =/  our  ?^(who u.who (dec (bex 128)))
```
belisarius222: I feel like this should emit a debug message if we're assuming a temporary identity. Otherwise bugs related to this could be hard to figure out.

belisarius222: Do we really need this feature? As part of locking this down, I wouldn't be surprised if we want to remove the ability for a larval Arvo to install vanes. That would also let us bake the `our` identity into a vane at install time, which feels right to me.

joemfb: The debug message is good idea. We need the feature for a `+solid` pill.

sys/arvo.hoon
```diff
@@ -502,101 +515,232 @@
...
+        our=ship                                        ::  identity
+        bud=vase                                        ::  %zuse
         vanes=(list [label=@tas =vane])                 ::  modules
     ==                                                  ::
-=<  |%
-    ++  come  |=  {@ (list ovum) pone}                  ::  22
+=<  ::  Arvo structural interface
+    ::
+    |%
+    ++  come  |=  {@ @ @ (list ovum) vise pone}         ::  22
```
belisarius222: Can we put some faces on these atoms? Same for +load

belisarius222: Oh, maybe it's fine since it just wraps `^come`.

sys/arvo.hoon
```diff
@@ -651,8 +800,100 @@
...
+          ::  hun: hoon.hoon source
+          ::  arv: arvo.hoon source
+          ::
+          now=@da
+          ova=(list ovum)
+          hun=@t
+          van=@t
+      ==
+  ^-  (unit (pair (list ovum) *))
+  ::  virtualize; dump error if we fail
```
belisarius222: This is a useful pattern, and I think it deserves its own arm in hoon.hoon. Out of scope for this PR.

sys/arvo.hoon
belisarius222: Then, I say go for it. I want us to err more on the side of removing unused code in general.

sys/arvo.hoon
joemfb: Cosign

sys/arvo.hoon
joemfb: Right. It's worth noting that `+come` and `+load` are the only typed methods of the arvo core -- they're only intended to be called on a new Arvo by the old Arvo during an upgrade. I've re-used that mechanism for the larval-stage upgrade.

sys/arvo.hoon
joemfb: That may be, but it's also pretty locked down already. It needs to support installing %zuse for `+ivory`, plus the vanes for `+solid`. Besides that, it only accepts identity and entropy. It has to support `+come` and `+load` for future kernel upgrades, and `+wish` for using `+ivory` on cc-release. I just removed the old %vega reset and `+peek` is now a no-op. I'll keep thinking about further ways to lock it down (it might be nice to use a stricter core variance), but nothing comes to mind at the moment.

joemfb:
I believe all is resolved, PTAL.

PR #916, opened 2018-12-06
Don't do two passes in the unjetted +sort

eglaysher:
While jetting +sort (urbit/urbit#1088), I made it do only a single
pass to separate the list around a pivot instead of using two. Do
this in the hoon code, too.


cgyarvin:
Both these sorts are stable sorts, right?  I always wanted to use timsort in a jet...

eglaysher:
This is equivalent to the previous version; this just does one pass instead of two; a skid instead of two skims.

PR #917, opened 2018-12-06
Use a trap to lazily evaluate the arguments to %dude in Eyre.

eglaysher:
Also adds a bunch of hints to make runtime more legible.


joemfb:
There's been lots of incompatible changes staged recently, and it looks like the CI config isn't quite right on release-candidate. You can update `.travis/pin-parent-pier.url` to include the hash of the current HEAD on release-candidate -- that should fix the PR build here.

belisarius222:
LGTM

wcable:
Note this is the inverse of #408; presumably the relevant "defunct requests" no longer occur, or no longer clog up the ford cache, or such?

(a midpoint might be `=<(|.([%leaf "eyre: execute {(spud tea)}"]) [spud=spud tea=tea])`; otoh I think the bloat was old hoon.hoon versions, and not just captured ++execute-turbo context)

PR #918, opened 2018-12-07
Move from just hashing time to using actual entropy.

eglaysher:
When we activate a vane, we should pass in real entropy instead of just hashing `now`. When we modify entropy, we should hash the concatenation of the old entropy and the new entropy instead of just xoring them.


belisarius222:
@jtobin What do you think of hashing the concatenation here, vs. XOR? I recall you had some opinions about this sort of thing.

eglaysher:
Merging per @belisarius222 , but I'm still curious about @jtobin 's thoughts.

PR #919, opened 2018-12-07
Fix turn types

belisarius222:
+turn now no longer has a TMI problem, nor does it lose its product type, even if passed a wet gate to transform the list.

The product type had to be calculated by type-inferring the product of running the input gate on an element of the list. Otherwise, a wet input gate will cause +turn to lose specificity in its output type, since the type of each element in the product list would be the result of inferring the input gate across its default sample.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -821,10 +821,13 @@
 ++  swag                                                ::  slice
   |*  {{a/@ b/@} c/(list)}
   (scag +<-> (slag +<-< c))
+::  +turn: transform each value of list :a using the function :b
 ::
-++  turn                                                ::  transform
+++  turn
   ~/  %turn
-  |*  {a/(list) b/gate}
+  |*  [a=(list) b=gate]
+  =>  .(a (homo a))
+  ^-  (list _(b ?~(a !! i.a)))
   |-
   ?~  a  ~
   [i=(b i.a) t=$(a t.a)]
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -827,7 +827,7 @@
   ~/  %turn
   |*  [a=(list) b=gate]
   =>  .(a (homo a))
-  ^-  (list _(b ?~(a !! i.a)))
+  ^-  (list _?>(?=(^ a) (b i.a)))
   |-
   ?~  a  ~
   [i=(b i.a) t=$(a t.a)]


sys/hoon.hoon
```diff
@@ -821,10 +821,13 @@
...
   (scag +<-> (slag +<-< c))
+::  +turn: transform each value of list :a using the function :b
 ::
-++  turn                                                ::  transform
+++  turn
   ~/  %turn
-  |*  {a/(list) b/gate}
+  |*  [a=(list) b=gate]
+  =>  .(a (homo a))
+  ^-  (list _(b ?~(a !! i.a)))
```
joemfb: I think I prefer `_?>(?=(^ a) (b i.a))`, and I think that's the convention for these casts. Now that I write it out, I'm surprised to see yours is shorter.

ohAitch:
Huh, in this case I definitely prefer the ?~ version: in some cases you are
actually conceptually asserting a is a cell, but in this case calling turn
on a statically empty(and thus untyped) list is valid, and will produce a
(list _(b !!)), i.e. (list _!!), i.e. $~

joemfb:
If `a` is a list, there's no semantic difference between `?~(a !! ...)` and `?>(?=(^ a) ...)`. In both cases, the crash/assertion is there only for type inference -- in neither case is it actually run.

joemfb:
... when those crashes/assertions are being used in a typecast, as is in question here. There's lots of these all over the `+in` and `+by` cores.

sys/hoon.hoon
eglaysher: I too have a minor preference for `_?>(?= ...)` for consistency, but this is only a minor preference.

ohAitch:
They are certainly both equivalent given list is ?(~ ^), the question is
about what connotations the code presents to the reader :)

joemfb:
But it won't infer `-:!>(*(list _!!))` for the empty case, it'll infer a list of items of the type of `b` when called with `a`.

joemfb:
The null list is not the void type, or of the void type. But we seem to be in agreement about the actual behavior, and simply disagreeing about the perception of some hypothetical observer...

ohAitch:
…which is the only measure by which a semantically equivalent code
substitution could plausibly be better or worse ^-^

joshuareagan:
I think we are all real observers here, not merely hypothetical!

joemfb:
I believe you recently appealed to consistency and conventions

ohAitch:
Against, but, fair. I think consistency is still firmly in the realm of
aiding the perception of observers, hypothetical or otherwise, since it has
been established that the behavior is equivalent.

belisarius222:
I'm pretty sure the perception of most observers will be "wtf is this" either way, so I updated it to match the convention.

PR #923, opened 2018-12-07
+slum to call raw nock gate; some associated cleanup

belisarius222:
Added a `+slum` gate to encapsulate calling a gate using Nock. Also removed cases where we were using `.*(gat -.gat)` where we should have been doing `.*(gat [%9 2 %0 1])` so we don't lose jetting.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -6170,7 +6170,7 @@
   ^-  toon
   ?.  &(?=(^ gat) ?=(^ +.gat))
     [%2 ~]
-  (mock [[-.gat [sam +>.gat]] -.gat] gul)
+  (mock [gat(+< sam) %9 2 %0 1] gul)
 ::
 ++  mule                                                ::  typed virtual
   ~/  %mule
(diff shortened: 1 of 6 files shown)


sys/hoon.hoon
```diff
@@ -6186,12 +6186,19 @@
...
     $1  [%| (turn p.ton |=(a/* (smyt (path a))))]
     $2  [%| p.ton]
   ==
+::  +slum: slam a gate on a sample using raw nock, untyped
+::
+++  slum
+  ~/  %slum
+  |=  [gat=* sam=*]
+  ^-  *
+  .*(gat(+< sam) [%9 2 %0 1])
```
joemfb: You don't like the explicit `%10`? It compiles to the same thing, of course.

belisarius222: I do like the explicit %10 better. Let me try it that way.

belisarius222:
@joemfb I switched the pill generators to use nock 9s too.

joemfb:
Nice! Took me a minute to realize those can't be `+slum` calls ...

belisarius222:
Hmm, gen/solid failed in CI, but worked on my machine. Not sure why, but it did die in +slum.

belisarius222:
Oh, CI does a double-solid. That must be the difference. Not sure what would be different about that, though.

belisarius222:
Hmm, double-solid also works on my machine. CI also had a different version of ames.hoon, so maybe that's related ... https://travis-ci.org/urbit/arvo/builds/465185913#L1568

joemfb:
I don't know specifically why this has broken, but it looks like you'll need to update `.travis/pin-parent-pier.url`. You can update that commit hash to point to any successful commit in this PR (or in this whole repository -- the pier is uploaded after every successful test.

joemfb:
My bad, your branch is on your own fork, but piers are only uploaded for branches in this repo. I guess you could try setting the pier URL to the HEAD of release-candidate.

belisarius222:
Ah, ok, thanks

joemfb:
It's failing because of the `+slum` call in `+solid`, since that is being run in an old pier with an old kernel. You'll need to be on a branch of this repo to stage all of these changes, but you could drop the change to `+solid` and save it for another PR -- the rest should be able to go through.

PR #925, opened 2018-12-09
use +slum in gen/solid

belisarius222:
Minor follow-up to https://github.com/urbit/arvo/pull/923 that uses the new +slum gate in the gen/solid generator.


joemfb:
You'll need to update the pinned pier url to point to latest release-candidate (which now has has the relevant kernel changes).

PR #928, opened 2018-12-10
Bumped ames protocol version

loganallenc:


joemfb:
We don't need to fix the CI on master, we're not going to be staging any more changes there.

PR #929, opened 2018-12-11
Delete all the things

eglaysher:


joemfb:
I tend to use `:time` for one-off "profiling". It's kind of hard to figure out how long something like +solid takes otherwise. Not opposed to some other approach, but I'm not sure we have a lightweight one right now.

In the other column, we could probably delete the markdown marks and structure.

web/talk.hoon
```diff
@@ -1,13 +0,0 @@
...
-  ::
-/?    310
-;module(nav_title "Talk", nav_no-dpad "", nav_no-sibs "", nav_subnav "talk-station")
-  ;script@"/~~/~/at/lib/js/urb.js";
-  ;script@"https://cdn.rawgit.com/seatgeek/react-infinite/0.8.0/dist/react-infinite.js";
-  ;script@"https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.11.2/moment-with-locales.js";
-  ;script@"https://cdnjs.cloudflare.com/ajax/libs/moment-timezone/0.5.1/moment-timezone.js";
-  ;script@"/talk/main.js";
-  ;link/"/talk/main.css"(rel "stylesheet");
-  ;talk;
```
Fang-: Goodbye, abomination. I'll miss you. ):

Fang-:
I'm guessing the thinking here is that we're not actively patching these things to keep working as we change all the fundamentals they build upon, so they're broken forever, so we might as well remove them?

On the one hand, I hope this code isn't drowned and lost in git history. On the other hand, all of these things would need to be redone from scratch anyway to work with Modern Urbit, on both technical and non-technical fronts.

[_thanks the apps for their service_]

c-johnson:
FYI, we probably can remove everything in /web outside of the landscape code. Or at least, Landscape doesn't use any files outside of /web/landscape, I'm not sure if other interfaces exist that use them.

eglaysher:
The reason we want to delete everything we can is to minimize the amount of data we transmit on first sync; all of this dead code gets sent over ames. (Also shrinks the size of pills that contain the whole filesystem, though that's not really as much of a concern.)

Also, much of this code doesn't compile anymore and was disabled in the compilation test app.

This patch is now ready for review.

eglaysher:
Regarding tree, I'm not even sure if it works with new Ford; the old dependency system from Ford is no longer here, and landscape is the new frontpage to your Urbit.

c-johnson:
The JS bundle is now down to 2.5 megabytes, which is still pretty bad. We have a plan to kill about ~75% of that, but in the meantime, it's worth considering a CDN. Is anyone strongly opposed to considering one on "it's not really decentralized" grounds?

cgyarvin:
Well, we seem to be using Urbit as a CDN right now, which it's not really
designed to do :-/

c-johnson:
Update on JS minification; I did a stupid with the initial implementation. We're now down to 800kb.

PR #931, opened 2018-12-11
removes strange %ames self-gift when forwarding to self

joemfb:
This pattern required a special-case in the event/effect processing. That pattern was synchronous, so handling it in the runtime offered no advantages -- instead we just recur here.

### DIFF
--- a/sys/vane/ames.hoon
+++ b/sys/vane/ames.hoon
@@ -1298,7 +1298,10 @@
       :~  [s.bon %give %woot q.p.bon r.bon]
       ==
     ::
-        %mead  :_(fox [[hen [%give %hear p.bon q.bon]] ~])
+        %mead
+      =^  moz  +>.$  (knob hen [%hear p.bon q.bon])
+      [moz fox]
+    ::
         %milk
       ::  ~&  [%milk p.bon q.bon]
       ?>  ?=([@ @ *] q.q.bon)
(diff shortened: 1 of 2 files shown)


belisarius222:
Where's the Vere code that handles this? I want to make sure I understand what the runtime was doing.

belisarius222:
Looks like maybe here?
https://github.com/urbit/urbit/blob/release-candidate/noun/vortex.c#L157

belisarius222:
I assume you'll have a followup PR to urbit/urbit that removes the associated runtime machinery.

belisarius222:
I'm a little concerned that omitting the extra Arvo event will break this error-handling:
https://github.com/urbit/urbit/blob/release-candidate/vere/raft.c#L1490

I don't fully grok it, though.

joemfb:
Thanks, @ohAitch, much obliged.

@belisarius222, the code you linked is inspecting the head of the event, not the effect. The code that handles this special event/effect to-self is `u3v_nick` in vortex.c, and this block of `_raft_punk()`: https://github.com/urbit/urbit/blob/release-candidate/vere/raft.c#L1603-L1619. With this change, that's effectively dead code. Nothing here depends on it, or any changes to it.

joemfb:
In other words, the code you linked runs after a crashed event, and checks if it was an ames packet. These changes are unrelated.

PR #932, opened 2018-12-11
Remove some tiscom

eglaysher:
This removes some redundant libraries, replaces some `=,` with explicit references, locks some libraries and structures with `^?` and removes some `^`.

### DIFF
--- a/app/ask.hoon
+++ b/app/ask.hoon
@@ -94,7 +94,7 @@
     (transmit set+~ pro+prompt ~)   :: XX handle multiple links?
   ::
       $det                              :: reject all input
-    =^  inv  som  (~(transceive ^sole som) +.act)
+    =^  inv  som  (~(transceive sole som) +.act)
     =.  sos  (~(put by sos) ost.bow som)
     ?~  wom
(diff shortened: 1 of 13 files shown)


Fang-:
> replaces some =, with explicit references

Maybe there's a conversation here I haven't been following, but... why? That's a lot of explicit references in hall!

joemfb:
urbit/urbit#1100 is one, but there's several others discussing the pitfalls of `=,`.

Fang-:
Working around issues with `=,` by just not using it seems... ugly at best. I appreciate there's other priorities, so fixing `=,`/how it's dealt with is probably not on the top of the list, but I do expect the explicit `:namespace`ing here to be reverted eventually (read: in 2019 (^: ).

eglaysher:
In addition to the linked bug, this is also an attempt to stop scanning the standard library over and over again during compilation to speed up first boot (which is also why this is part of a patch that locks libraries with `^?`).  We found that a large chunk of our first app compilation was in +fond.

ohAitch:
I'm still a big fan of the   =,  ^?(format)  convention, as an alternative
to pushing that decision to the definition site.

PR #935, opened 2018-12-12
Behn dumb

belisarius222:
Old Behn used a Brodal skew queue -- or maybe two? [Brodal said of his queue](https://en.wikipedia.org/wiki/Brodal_queue#cite_note-Brodal-1) that it was "quite complicated" and not "applicable in practice." New Behn uses a list, sorted by date.

It's hard to imagine Behn being dumber than this. While the asymptotic performance can't approach that of a skew queue, I think the real-world performance will be fine, and maybe even better in the common case. Setting a timer is now O(n), where n is the number of timers that will elapse before it. Using Landscape from a web client, I see four timers total at any given time, and the eight-second heartbeat timer is always the first one, which means no traversal -- O(1) and dead-simple. I did some benchmarking that suggests that even with a thousand active timers, the worst case to set a new timer is still firmly sub-millisecond.

This does not seem to suffer the bug @ixv noticed plagues current Behn, which sometimes neglects elapsed timers without notifying their clients.

This implementation notifies clients of all elapsed timers in a single event. We might want to switch that to one timer per event at some point for cleaner error-handling.

All this being said, I'm not 100% sure we should merge this -- it doesn't have tests, and I don't really know what the spec of old Behn was, so there's likely some deviation. If we don't merge it, at least @ixv should be able to use it temporarily for his development to see if it really was old Behn that was causing the heartbeats to get lost in Landscape.


belisarius222:
I agree with @joemfb's sentiments. It should also be noted that we'd need to breach this in as written, since its +load doesn't do a migration from the old Behn state.

belisarius222:
It should also be noted that this scry handler is not completely referentially transparent. Any elapsed timers might or might not show up in the result, depending on whether Behn has already run during this event.

This problem is similar to Clay scrys at now resolving differently based on whether a commit has been applied during this Arvo event.

For Behn, we could get back to referential transparency by omitting elapsed timers from the scry output, at the expense of losing a fair amount of visibility into its state.

belisarius222:
@ixv says he's using this in his landscape-api branch for his own development and confirmed that it fixed his bug.

joemfb:
I don't think this scry is any worse than lots of other current-date scrys. I wonder if we can make that safe at the top-level by only allowing scrys into completed events. I realize that would make some things quite a bit more difficult. That's completely out of scope here, of course. But I don't think this scry is an obstacle to merging.

PR #936, opened 2018-12-12
Don't call +sell in normal cases in Ford.

eglaysher:
+sell invokes the type checker and thus adds 10ms to each event
where it is used. +sell is invoked in "tapes like {<this>}". Ford
error messages use that format on errors. Many common Ford builds
are expected to fail, but swallow the error messages. This modifies
them so that they generate their error messages in a cheaper way.

### DIFF
--- a/sys/vane/ford.hoon
+++ b/sys/vane/ford.hoon
@@ -2485,11 +2485,8 @@
         =^  maybe-schematic-results  out
           %-  perform-schematics  :*
             ;:  weld
-              "ford: %bake "
-              (trip renderer)
-              " on "
-              (spud (rail-to-path path-to-render))
-              " contained failures:"
+              "ford: %bake "  (trip renderer)  " on "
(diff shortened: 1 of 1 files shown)


ohAitch:
Gosh these are incredibly more difficult to read than the single-line
versions

Would it be possible/desirable to restructure them to something more like
e.g.

    =+  [renderer=(trip renderer) path=(spud (rail-to-path path-to-render))]
    "ford: %bake {renderer} on {path} contained failures:"

ohAitch:
Four lines on average, actually: a more specific version of my complaint is
"having welds at all".

The desirable property I was hoping could be preserved, is that for every
single output line there is a source line that compactly describes its
structure. Further lines are necessary to work around the "we have nice
things" dynamic dispatch banhammer, and beyond that blank  ::  lines are
probably a good idea in the list contexts; it's not the number of source
lines, it's the usage of explicit concatenation, and the extra overhead /
misleading whitespace that introduces.

PR #939, opened 2018-12-13
Fix Behn bug that broke Clay

belisarius222:
Clay expected responses to its timers to come back in the same order it sent them. So needy!

Now Behn actually does this.


belisarius222:
The glorious refactor caused a bail:meme. Do not merge.

c-johnson:
Works beautifully. Even on non-zod galaxies! 

Thank you man! This is great.

PR #945, opened 2018-12-14
Dill, stop hitting yourself

belisarius222:
Dill used to send Arvo moves to itself to unwrap `%flog` tasks. Now it just unwraps and recurses internally.


sys/vane/dill.hoon
```diff
@@ -507,9 +507,9 @@
       [~ ..^$]
     ::  this lets lib/helm send %heft a la |mass
     ::
-    =/  not=note-dill
-      ?:(?=([%crud %hax-heft ~] p.task) [%heft ~] p.task)
-    [[u.hey.all %slip %d not]~ ..^$]
+    =?  p.task  ?=([%crud %hax-heft ~] p.task)  [%heft ~]
+    ::
+    $(wrapped-task p.task)
```
joemfb: Does this `%=` need to also change `hen` to be `u.hey.all`?

belisarius222: Yes, it does. I am somewhat embarrassed I missed that.

belisarius222:
Fixed and rebased. PTAL @joemfb

belisarius222:
Arvo moves will always have a certain amount of performance overhead -- at least some extra function calls -- although hopefully eventually nearly negligible. There's conceptual overhead from creating new events, though. It clutters up |verb, for example, and if we do duct-based tracing, it'll pollute that with essentially spurious moves.

Also, there's a fair amount of wire parsing for each move. If we switch paths to `(list dime)`, that cost should go almost to zero, but for now, the extra moves probably do matter for performance at least a little.

PR #946, opened 2018-12-14
Compatibility with Azimuth

Fang-:
With this PR, we can boot against the latest version of the contracts as deployed on Ropsten.

This also applies the rebranding to the stdlib arms and associated faces.

I want to do a re-organizing of the `++ethe` and `++ethereum` cores, too. Here's the structure I want (and tried) to use, but currently doesn't actually work:

```hoon
++  azimuth
  =>  ::  everything under ++azimuth:ethe (azimuth-related structures)
  ::  everything under ++azimuth:ethereum (logic for those structures)
::
++  ethereum
  |%
  ++  key
    ::  address-from- logic
    ::  signing logic
  ::
  ++  rlp
    ::  rlp logic as it currently exists
  ::
  ++  abi
    =>  ::  ethereum abi structures from ++ethe
    ::  everything under "encoding"
    ::  everything under "decoding"
  ::
  ++  rpc
    =>  ::  ethereum rpc structures from ++ethe
    ::  everything under "making calls to nodes"
    ::  everything under "parsing responses from nodes"
  ::
  ::  misc utils that aren't *strictly* ethereum-related, but have nowhere else to go
  --  
```

Problem here is that zuse is split up into two cores: models and engines. This is why `++ethe` existed as separate from `++ethereum`, a reason I completely forgot.  
We strictly need the models to be in the first core, because they're used in the models for jael. We strictly need to engines (ie old `++ethereum`) to be in the second core, because it uses arms from that engine core.

There's probably reasons why we care about separating models and engines so strongly, rather than just having them be in separate sections within the same core. That would mean we do want to keep the `++ethe` and `++ethereum` split. Please confirm, and any advice about making that split (accounting for the `++ethereum` `++azimuth` split too) slightly more elegant would be appreciated. I'm probably just being dumb/stubborn about it.

(There's also a couple todo's here I still want to try touching up, but I'll get to that after the re-org.)


joemfb:
First, I want to note that all the arms in the `|dawn` core exist to be called by the runtime (see `vere/dawn.c`). Unfortunately, there's also a couple places where we grab these types by name. So these changes must be mirrored in that file before this could go in. It might be a good idea to make the `|dawn` interface more generic, perhaps just `+give` and `+take` gates that operate on a list of RPC requests or responses, respectively,  but that is a larger task. 

I like your proposed refactor, and think we can get close to something like it. A smattering of thoughts on the subject:

I believe the core separation is for performance (constant-folding, and I'm not sure what else). I don't know if separate chapters of the same core provide the same benefit, but that could be an option if so.

Is there are a technical distinction we could leverage for namespace-ing? `+solidity` vs `+ethereum`, or something along those lines? Also, it seems like `+jael` only depends on a few of these types, so I bet we could lift those specifically into the models core.

And we can always put two cores in the same namespace:

```
++  foo
  =>  |%  ...  --
  |%   ...  --
```

Fang-:
>First, I want to note that all the arms in the |dawn core exist to be called by the runtime

I am aware. Was pleasantly surprised to find that no C changes were needed at all here! I also haven't touched, and don't plan to touch, the organization or naming of `++dawn`. It seems to serve its purpose well enough for now.

>Is there a technical distinction we could leverage for namespace-ing?

Aside from what I'm already proposing? Not really. What you imagine as `++solidity` lives under `++abi`, since that's the only place outside of the blockchain you really ever interface with solidity concepts (that is, its types and functions).

>And we can always put two cores in the same namespace

So I'm already doing that for the type-logic distinction within the proposed cores. How sane would it be to suggest:

```hoon
|%  ::  zuse structures core
++  azimuth-types
  |%  etc  --
--
::
|%  ::  zuse engines core
++  azimuth
  =>  azimuth-types
  |%  etc  --
--
```

and similar for all the other types cores. Then just use `++azimuth` always, outside of zuse's structures core?

sys/zuse.hoon
```diff
@@ -8377,13 +8370,13 @@
       ^-  [ship [life pass]]
       :-  gal
       ~|  czar-gal=gal
-      [life pass]:(need net:(~(got by hul.eth.snap) gal))
-    ::  +hull:snap:dawn: extract ship's contract state
+      [life pass]:(need net:(~(got by pos.eth.snap) gal))
+    ::  +point:snap:dawn: extract ship's contract state
     ::
-    ++  hull
+    ++  point
```
joemfb: https://github.com/urbit/urbit/blob/release-candidate/vere/dawn.c#L309

joemfb: Weird, github lost a bunch of comments here. I'll just add them individually.

sys/zuse.hoon
```diff
@@ -8285,31 +8277,32 @@
...
       =/  pub=(unit pass)
-        (pass-from-eth:constitution:ethereum enc aut sut)
+        (pass-from-eth:azimuth:ethereum enc aut sut)
       ?~  pub  kyz
       (~(put by kyz) who [rev u.pub])
-    ::  +hull:take:dawn: parse ship's contract state
+    ::  +point:take:dawn: parse ship's contract state
     ::
-    ++  hull
+    ++  point
```
joemfb: https://github.com/urbit/urbit/blob/release-candidate/vere/dawn.c#L344

sys/zuse.hoon
```diff
@@ -8202,9 +8194,9 @@
         =-  [from=~ to=tract gas=~ price=~ value=~ data=-]
         (encode-call:ethereum 'getKeys(uint32)' [%uint gal]~)
       [%number boq]
-    ::  +hull:give:dawn: Eth RPC for ship's contract state
+    ::  +point:give:dawn: Eth RPC for ship's contract state
     ::
-    ++  hull
+    ++  point
```
joemfb: https://github.com/urbit/urbit/blob/release-candidate/vere/dawn.c#L341

sys/zuse.hoon
```diff
@@ -243,36 +243,35 @@
     ==
   ::
   ::
-  ++  constitution
+  ++  azimuth
     |%
-    ++  hull
+    ++  point
```
joemfb: Bunted here: https://github.com/urbit/urbit/blob/release-candidate/vere/dawn.c#L316. It doesn't look like the structure changes are themselves significant.

joemfb:
I think your proposed core structure(s) would work. I would try to limit `+azimuth-types` to the bare minimum necessary for the `%jael` moves.

Fang-:
It would, of course, have been nice if I actually tested this thing after doing the renaming. (^:

See urbit/urbit#1111 for the C-level changes, just renaming.

Still wrestling a little bit with finding the least shitty structure, commit Soon™.

Fang-:
Not 100% satisfied with the reorg of the cores, but it's good enough for now. Happy to incorporate feedback if there is any.

@joemfb I touched up some logic in `++point-from-eth`, ensuring two things:
1) Don't crash in the not-uncommon case of key revision 0 (usually as in "no keys configured").
2) Activated ships always have networking state that can be modified.

The latter was already there, and an implementation error on my part. The former comes from `need`ing the product of `++pass-from-eth`, which returns `~` if the crypto suite is not `1` (or if either pubkey is not 32 bytes, but they're always parsed and stored as if they were, so that's practically never hit).

There's other TODOs remaining in the Ethereum and Azimuth code, but none of them are strictly necessary. I'll put them on the release-candidate punchlist, maybe I'll poke at a couple over the holidays.

For convenience, to test against latest Ropsten deploy, use this key for ~zod:

```
-G '0wln.i~qct.VWDJf.DCmIw.GQnvj.QUE4n.ZcbkZ.yclBJ.rSaER.9-wxe.EZf3G.1pMDq.Am6pu.kjaK3.1Zw6d.RoFeF.h2pNq.ZtTEg.4w0sp'
```

sys/zuse.hoon
```diff
@@ -2056,8 +1778,8 @@
     ::  through %ames and use %behn timers.
     ::
     ++  logs                                            ::  on-chain changes
-      %+  map  event-id:ethe                            ::  per event log
-      diff-constitution:constitution:ethe               ::  the change
+      %+  map  event-id:ethereum-types              ::  per event log
```
joemfb: Plz fix margin comment.

tests/sys/zuse/ethereum/encoding.hoon
```diff
@@ -42,3 +42,4 @@
             "48656c6c6f2c20776f726c642100000000000000000000000000000000000000"
         ==
 --
+[l+~ l+[l+~ ~] l+[l+~ l+[l+~ ~] ~]]
```
joemfb: Spurious change to trigger a sync?

joemfb:
Tested with the latest Ropsten contracts, everything looks good (barring the style requests above).

tests/sys/zuse/ethereum/encoding.hoon
Fang-: Wh- what is this? Looks like some test noun for RLP encoding, but why in this file? And why did it show up now? How did I not see this in the staged diff?

mystery-smallest

joemfb:
@Fang-, please temporarily hold of merging this. I'm fixing pill creation in CI, and would like to get that in before this.

PR #948, opened 2018-12-18
Fix arvo ova order

eglaysher:
Behn is sensitive to the order in which ova are emitted, which was not
correct. Previously, ova emitted first within a single vane activation
were processed as effects last.


eglaysher:
This is ready for review.

belisarius222:
LGTM now. My previous comment was incorrect: `+said` does not reverse move order.

PR #960, opened 2019-01-02
Remove duplicated code between arvo and hoon

eglaysher:
Also removes some dead code from arvo which lets us clean up some
^ in dojo.


ohAitch:
Yeah, scaffolding doesn't belong on(/the way to) master obviously, but might be good to put a revert in a "future-marks" branch or similar.

(It'll probably be useful as part of staging a live network update, but like, with the specific types the new features turn out to actually need)

PR #963, opened 2019-01-03
Handle the move which makes the ova first

eglaysher:
Ref #962

### DIFF
--- a/sys/vane/behn.hoon
+++ b/sys/vane/behn.hoon
@@ -136,13 +136,13 @@
     ==
   ::  +set-wake: set or unset a unix timer to wake us when next timer expires
   ::
-  ::    We prepend the unix %doze event so that it is handled first. We must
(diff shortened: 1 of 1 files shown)


philipcmonk:
I feel like this should just reorder the moves in the various `?-` cases rather than removing the flop at the end.  Virtually every core that maintains a `(list move)` flops it at the end so that incrementally adding is easy and happens in the order you'd expect.

ohAitch:
Yeah, if the issue is (the moves of) notify-clients ought to happen after
set-wake, it would be a lot clearer if the invocations were reordered
accordingly in ++call.

ohAitch:
Hm, or barring that due to data flow constraints, a renaming to e.g.
(set-wake-before moves)

belisarius222:
A comment explaining why we prepend the Unix move would also be helpful.

PR #965, opened 2019-01-07
switch +mug hash from FNV to Murmur3

joemfb:
This PR is the dual of urbit/urbit#1123.

The diff here hides the ball when it comes to the complexity of this change: `+mug` determines the horizontal and vertical ordering of the hash-balanced treaps `+set` and `+map`, and ordering of `+map` determines the order of nock formulas in a core battery. To switch these out, I staged parallel copies of the hash-order, container, and compiler jets. This didn't quite fully work, so I also had to switch back and forth with the correct boot-sequence implemented on the `cc-release` branch. The process was quite messy, and isn't easily conveyed in a single commit history, so I've just included the minimum changes here and manually uploaded the resulting pill.


tests/sys/vane/ford.hoon
```diff
@@ -1297,15 +1297,15 @@
         ==
       ::
       ^=  moves
-        :~  :*  duct=~[/live]  %give  %made  ~1234.5.6  %complete
+        :~  :*  duct=~[/once]  %give  %made  ~1234.5.7  %complete
```
belisarius222: This test's move order fragility is lame. Definitely out of scope, though

ohAitch:
Shouldn't it technically be ++mum?

joemfb:
Technically?

ohAitch:
Or ++muk, looks like? My point is, switching over the system to a new hash
function, vs redefining what the existing hash function name points to.

joemfb:
I guess, maybe. It's such a deeply breaking change that it hardly seems to matter. Does the name `+mug` have some kind of broader significance?

ohAitch:
Hm, I suppose if it did that would warrant keeping the name. My
understanding of it was just "temporary fnv hash, to be replaced with a
more secure production algorithm at a later date"

ohAitch:
(This is vaguely similar to the distinction between %b the interface name
and %behn the implementation, which I suppose has similarly not been
properly cycled, bc testnet?)

joemfb:
Are you saying that %behn should be renamed (say, %bane) since it's been reimplemented, but should preserve the letter %b since the interface is the same? I haven't been privy to vane interface contract discussions, but this would be surprising to me. And how would we handle that fact that I changed the %behn interface from removing +keep/+doze and the whole i/o polling subsystem?

ohAitch:
Removing the i/o polling subsystem wants an arvo kelvin(hoon kelvin? Idk if
arvo formally has its own kelvins), and is I think orthogonal; by %b I am
referring to ++note-arvo/++sign-arvo, aka whether other vanes are still
sending-receiving the same cards.

I believe *-arvo does not use the full name %behn(%ford etc) to be able to
talk about multiple competing implementations, which you will inevitably
have in a mature Urbit system if you try to rewrite a vane from scratch; we
are not currently in a mature urbit system, so redefining the name seems
reasonable. ++mug is more iffy bc it is pretty solidly bound in people's
minds as a particular hash, and if hypothetically there existed .hoon files
outside the arvo repo, using ++mug for directly hashing, and those hoon
files had tests, they would hereby be broken; since this is approximately
not the case, it's totally your prerogative as the implementer :)

joemfb:
There was problems with the new +mug jet in urbit/urbit#1123, so these changes have been fully restaged. The end result should be exactly as it was (with the addition of some basic tests), but this PR now includes the compiler ping-pong needed to get to this point.

One issue worth noting explicitly: the initial order of console apps in the drum is circuitously dependent on a hash-order. In `lib/hood/drum`, that order explicitly depends on the order in which console subscriptions are established. On boot, console subscriptions are requested before the apps themselves have been started, so the subscription move gets placed in a per-app queue in %gall. The map from app-name to queue appears to be traversed in hash-order. I spent some time trying to hack around this in drum, but I wasn't able to come up with a solution there. I think this will have to be fixed in %gall.

joemfb:
@ohAitch, there were Kelvins sacrificed in the as-of-yet unreleased language changes on release-candidate. I've been operating under the assumption that any additional changes made prior to release can use house money wrt to Kelvins.

Note that when I single-homed Arvo and the vanes, and changed almost every vane interface. Between that and I/O polling, I have changed every vane interface. I'm certainly looking forward to having a more complete Arvo versioning and interface-contract story.

joemfb:
This has been restaged once more, this time replacing `+hor` with `+gor` for horizontal set balancing. I've also renamed `+vor` to `+mor` to excise any association between the hash-order names and horizontal vs vertical.

The previous staging is on the `murmug-only` branch. As you can see, the only difference is in the presence and use of the hash-order functions (and the move order of a %ford test): https://github.com/urbit/arvo/compare/murmug-only..murmug

belisarius222:
The order changes LGTM. Reapproved

PR #966, opened 2019-01-08
Fix eyre multi-request crash

belisarius222:
Formerly, if you asked Eyre to make two requests to the same domain in a row, Eyre would make three requests and then crash. This fixes that, and it should also have better %born restart handling.

The bug turned out to be that Eyre was both enqueuing and immediately sending a request, which caused an extra request to be sent later, which in turn caused Eyre to receive a response to an unknown request and die.


joemfb:
Does this supersede #959?

belisarius222:
Yes, it does. I forgot to mention that, but I'll close that one.

PR #969, opened 2019-01-09
stop appending new lines to wains with trailing new lines

ixv:


cgyarvin:
They shouldn’t be there, true...

PR #973, opened 2019-01-10
Reduce ++hard usage in jael for speed

philipcmonk:
How do you get typed responses from eyre without running into issues around ++hard?  I've got a poor solution working, but I'm interested in the general solution.

Up untill now, Jael makes HTTP requests to an Ethereum node, and Eyre parses the response through the %json-rpc-response mark, which parses the httr to json and a couple more things.  Eyre gives a %sigh back to Jael, which includes the json response as a vase.  However, Jael needs to use that result as typed data, so it ++hard's it to the appropriate type (mostly a bunch of json).

This basically works on testnet, but it fails on mainnet because we have much more data, and ++hard is superlinear in the amount of json, so it crashes.  We can get around this by splitting the requests into really small batches (so the superlinearity doesn't kill us), but that's not a real solution.

What I've done for now is stored the JSON part as simply text in the mark translation, then parsed it in Jael, because parsing JSON is much faster than +hard'ing json.  This splits the parsing weirdly across the two locations, though.  I suppose one solution would be to forget the mark and just parse the whole httr in Jael.  Another solution would be to parse out the json into a more specific type in the mark, but I'm not sure if we can actually give it a sufficiently specific type that the ++hard is fast.  Plus, that leaves in the ++hard, which is a code smell.

Are there other solutions to get typed to a vane from eyre?

### DIFF
--- a/mar/json/rpc/response.hoon
+++ b/mar/json/rpc/response.hoon
@@ -2,14 +2,15 @@
 /-  json-rpc
 =,  json-rpc
 ::
-|_  res=response
+!:
+|_  res=raw-response
 ::
 ++  grab                                                ::  convert from
   |%
-  ++  noun  response                                    ::  from noun
+  ++  noun  raw-response                                    ::  from noun
   ++  httr                                              ::  from httr
     |=  hit/httr:eyre
-    ^-  response
+    ^-  raw-response
     ~|  hit
     ?:  ?=($2 (div p.hit 100))
       =,  html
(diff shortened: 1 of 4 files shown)


joemfb:
My preference in this situation has been to skip the mark and just get `+octs` from %eyre. I'm surprised that parsing JSON is that much faster that `(hard json)` though.

mar/json/rpc/response.hoon
```diff
@@ -18,6 +19,16 @@
...
     |=  a=json
+    ^-  raw-response
+    =;  resp
+      ?-    -.resp
+          %error   resp
+          %fail    resp
+          %batch   resp
+          %result
+        ~|  resp
+        [%result id.resp (crip (en-json:html res.resp))]
```
joemfb: The need to re-serialize here make me lean even harder towards just requesting `%httr` from %eyre.

philipcmonk:
Yeah, it's the difference between "a minute or two" and "doesn't finish as far as we can tell".

I'll write a version that skips the mark entirely and see if it's cleaner.

philipcmonk:
Done.  I do feel like this is cleaner.

cgyarvin:
This is what %meta is for, I think...

PR #978, opened 2019-01-11
scry for +life in %ames (fewer %open packets)

joemfb:
This fixes two issues that were leading to more than necessary %open-outgoing and dropped-incoming packets. In one case, we weren't scrying for a +deed where we could be, in the other, we were scrying, but always guessing a +life of 1. This PR also renames the %jael scry gates for clarity.

### DIFF
--- a/sys/vane/ames.hoon
+++ b/sys/vane/ames.hoon
@@ -523,6 +523,19 @@
       =/  pur=spur
         /(scot %ud lyf)/(scot %p who)
       [[151 %noun] %j our %deed da+now pur]
+    ::  +life-scry: for a +life
+    ::
+    ++  life-scry
+      ~/  %life-scry
+      |=  who=ship
+      ^-  (unit life)
+      =;  lyf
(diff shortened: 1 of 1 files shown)


sys/vane/ames.hoon
```diff
@@ -741,13 +754,13 @@
...
             rum(dod &, bum ?~(cop bum.rum (~(put by bum.rum) did.rum u.cop)))
-          =/  seg  (sein her)
+          =/  seg  (sein-scry her)
           =^  roc  diz  (zuul:diz now seg [%back cop dam ~s0])
           (busk(diz (wast:diz ryn)) xong roc)
         ::  XX move this logic into %zuse, namespaced under %jael?
         ::
         ++  deng                                        ::    deng:ho:um:am
-          |=  law=(unit ^deed)                          ::  accept inline deed
+          |=  law=(unit deed)                          ::  accept inline deed
```
philipcmonk: Uneven comment spacing

PR #983, opened 2019-01-13
improves exponential backoff and :hall notifications in :dns

joemfb:


app/dns.hoon
```diff
@@ -10,7 +10,7 @@
       [%hall-action %phrase audience:hall (list speech:hall)]
   ==
 +$  card
-  $%  [%hiss wire [~ ~] %httr %hiss hiss:eyre]
+  $%  [%hiss wire ~ %httr %hiss hiss:eyre]
```
joemfb: This *should* fix the `auth-lost` printfs.

PR #984, opened 2019-01-13
on the nature of comparative quantification

joemfb:
maximum is the *opposite* of minimum


jtobin:
🤔

joemfb:
I tell the truth where no one else will.

PR #941, opened 2018-12-13
single-home the vanes

joemfb:
The Costello to #915's Abbott. This removes all the multi-homing indirection in the state, state-machines, and moves of all the vanes. The closes thing to an exception is `%gall`'s `%deal` task, which takes a pair of ships, but uses them alternatively for bidirectional communication.

I've tested networking, syncing/merging, and cross-ship app pokes in a number of scenarios, but there's obviously still risk involved in making this many changes across this many places. Drive-by reviews of the area one knows best would be greatly appreciated.

PR #987, opened 2019-01-13
adds vane-sample type to arvo, fixes type bug in +load/+stay calls

### LINKED ISSUE #985: Weird our in vanes
If you add a `~&  [%vane-load our]` into any vane's `++load', you will get the following result:

```hoon
[%vane-load ~laszod-dozser-hiddur-podzod--dozzod-dozzod-dozzod-dozzod]
```

That value is consistent across ships, and I have no idea where it's coming from. Somehow, this doesn't seem to be breaking anything?

I poked around in `arvo.hoon` too, putting a bunch of printfs to have it show every `who` and `our` it deals with, but couldn't find any place where it is using that value.

joemfb:
closes #985


belisarius222:
LGTM

joemfb:
this bug was introduced in #941.

PR #989, opened 2019-01-14
use foreign aeon for foreign mark definitions

philipcmonk:
When getting mark definitions to compare a foreign commit against its merge base, we need to use `case-to-aeon` of the foreign desk, not our local one.

I've tested that this fixes the issue locally, but I haven't tested it with multiple ships.

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -631,6 +631,7 @@
     ::  translate other cases to dates
     ::
     =/  aey  (case-to-aeon:ze case)
+    ::  ~&  [%case-to-date aey let.dom our her syd case]
     ?~  aey  `@da`0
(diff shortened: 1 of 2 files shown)


philipcmonk:
I think this is what changed the semantics from "use foreign marks at foreign case" to "use foreign marks, and convert the aeon to a date, but using the local aeon-to-date mapping": https://github.com/urbit/arvo/commit/165083b253d8b86feae3f47ca6c94ed450cfce90#diff-f459f5d2b2ef649e3812066f90c3906eR3028

This confirms that the original intent last time clay was rewritten was to use foreign marks.  I don't trust the original intent too much because I don't trust the author, but at least he had it all in his head at the time.

PR #993, opened 2019-01-14
try iterating over the right thing

belisarius222:
@joemfb

### DIFF
--- a/sys/vane/eyre.hoon
+++ b/sys/vane/eyre.hoon
@@ -757,7 +757,19 @@
         $them                                           ::  outbound request
       ?~  p.kyz
         ?~  sud=(~(get by kes) hen)
+          ::  delete an element from q.ask by traversing to prevent leakage
+          ::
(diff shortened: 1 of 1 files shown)


ohAitch:
Ooh, is  ?~ sud=foo  new syntax?

Fang-:
Woah, that's wild. Adds to the subject so sneakily! Is that intended to work? Do we want to dis/allow that?

ohAitch:
To elaborate, pop quiz:

|_ count=@
++  this  .
++  frobnicate
  |= a=@
  ?: (complicated-condition a) ~
  [count=count etc=(stuff a)]
++  foo
  ?~  (frobnicate 5)
    =.  count +(count)
    this(that (dec that))
  =.  count  +(count)
  this(that (add 2 that))
--

what would you naively expect to happen to count.this upon calling ++foo,
vs what will happen if ?~ chooses to spill its internal subject willy-nilly?

PR #996, opened 2019-01-15
removes |ask prompt from :dojo

joemfb:


lib/hood/helm.hoon
```diff
@@ -99,9 +99,8 @@
   [[her %hood] %helm-hi ?~(mes '' (crip u.mes))]
 ::
 ++  poke-send-ask
-  |=  mel/cord  =<  abet
-  %^  emit  %poke  /helm/ask/(scot %p ~socden-malzod)
-  [[~socden-malzod %ask] %ask-mail mel]
+  |=  mel/cord
```
belisarius222: Let's either remove gen/hood/ask/hoon entirely, or have it print a message about why it's deprecated. In either case, we should just delete the hood interface here.

joemfb: I was just thinking in terms of minimizing changes -- there's also `mar/ask-mail` and `app/ask` itself. Do you think an error is preferable to silence if somebody familiar with past versions give it a try?

belisarius222: Yeah, I think a message is better; something like:
```
|ask is no longer available. To obtain a planet, see the Azimuth documentation at https://urbit.org/docs/getting-started/azimuth/
```

I also don't think it would be terrible for there to be an error like "gen/ask/hoon doesn't exist."

belisarius222: I just don't think we should leave in code that explicitly no-ops.

baudtack: I agree with @belisarius222 here. This should either be gone so  there is an error or spit out a message. Either is fine, though... do the docs actually  address HOW to get a point atm? Last thing I read was basically "hopefully you got one from somewhere already"

PR #999, opened 2019-01-15
Jael & zuse fixes

ixv:


philipcmonk:
This doesn't pass the tests, I'll see if I can fix them.

sys/vane/jael.hoon
```diff
@@ -687,6 +687,11 @@
       ::  save our parent signature (only for moons)
       ::
       =.  sig.own.sub  sig.seed.tac
+      ::  if we're given a snapshot, restore it
```
Fang-: Since we hit this as an issue, maybe worth adding another line of comment saying why it's important that this happens before "our initial public key".

philipcmonk: Sure

philipcmonk:
Travis completed successfully on the previous commit, so I'm going to merge this.

PR #1000, opened 2019-01-16
Quality-of-life improvements for %jael

philipcmonk:
`.^([latest-block=@ud source=(each ship node-src:jael) %j /=eth-status=)` now gives you some useful information about your current jael state.

I also added a `snap` mark so we can load these snaps into clay and inspect them.

Finally, I added a `%snap` option to `%jael` and `%eth-manage` so that we can restore a snapshot that we put in clay on-demand.

### DIFF
--- a/mar/snap.hoon
+++ b/mar/snap.hoon
@@ -1,0 +1,16 @@
+::
+::::  /hoon/snap/mar
+  ::
+=,  mimes:html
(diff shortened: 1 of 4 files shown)


ohAitch:
<s>hopefully it parse-errors, actually</s>

PR #1009, opened 2019-01-17
always use our reef for clay

### LINKED ISSUE #1008: checking in syntax error for zuse breaks desk forever
@belisarius222 @joemfb 

Currently, when we read a file, we go through the following process:

- parse/compile hoon, zuse, and arvo on the same desk as the file to create a reef
- run the mark against that reef
- use the result

The problem is that if {hoon,zuse,arvo}.hoon fails to compile at any revision, then all future revisions will fail because we can't create the reef to run the mark (because it uses the previous revision's *.hoon file).  The only reason this isn't a problem for %base and %home is that we explicitly short-circuit those in ford.

My best guess at a fix is to use the {hoon,zuse,arvo}.hoon from the incoming revision, though I'm not sure how that flow of control works in ford.

Repro:

- boot fakezod
- |mount /=kids=
- put syntax error in zuse in kids
- try making any change (including fixing the syntax error)
- observe that your %kids desk is irretrievably broken

philipcmonk:
fixes #1008 

I'm not sure this is the right solution, but if we run into issues with clay being rickety, this is a solid hotfix that makes sense to me, at least while we're regularly breaching (so we can expect peoples' hoon/zuse/arvo/marks to be pretty compatible).

I suspect we'll eventually want to revert this and instead compile the foreign marks with the foreign kernel, but it has to be done at the new revision, and I don't know how to tell ford about the new kernel for that desk before the revision is even complete.

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -838,7 +838,7 @@
         %+  turn  `(list path)`mus
         |=  a/path
         :-  [%$ %path !>(a)]
-        :^  %cast  [her syd]  %mime
+        :^  %cast  [our %home]  %mime
         =+  (need (need (read-x:ze cas a)))
         ?:  ?=(%& -<)
           [%$ p.-]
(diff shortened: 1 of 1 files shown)


philipcmonk:
Of note, this should be a drop-in fix for a broken ship, it shouldn't require recreating the pier.

joemfb:
To be clear, this just changes %clay to always use our %home desk for marks?

philipcmonk:
Yes, at least that's my intent

ohAitch:
*puts on 2014 hat* or we could, you know, not accept clay commits with
broken hoon.hoon's, and then the current version of the mark definition
will always be well-defined.

philipcmonk:
@belisarius222 You know ford best, do you agree we should merge this for now?

PR #1010, opened 2019-01-17
adds rate-limit handling to :acme

joemfb:
This was anticipated to work with the generic `:acme` retry mechanism through `&recoverable-error`, but the trigger for that code path is failed mark conversions.

This PR adds explicit handling for `HTTP 429` responses, and matches on the two rate-limits we're likely to hit, falling back to more generic retry with exponential backoff if it can't match them. Additionally, it updates the certificate renewal timer logic to be biased towards the end of the week (see comment).

### DIFF
--- a/app/acme.hoon
+++ b/app/acme.hoon
@@ -983,8 +983,21 @@
         (emit (notify msg ~))
     ::  set renewal timer, install certificate in %eyre
     ::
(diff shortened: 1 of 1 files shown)


ixv:
Its probably a good idea to test a merge of this from galaxy -> star -> planet on a local network before merging right?

PR #1014, opened 2019-01-17
slum in Arvo scry

belisarius222:
@eglaysher's recent profile shows a significant percentage of Ames processing time inside Arvo's +scry function, almost all of which happens on +slym:wa cache misses. Since we don't actually use the type from the result of the +slym call, there's no reason to call the compiler. Just run the function untyped using +slum and clam the output to `(unit (unit (cask)))`.

### DIFF
--- a/sys/arvo.hoon
+++ b/sys/arvo.hoon
@@ -191,10 +191,7 @@
         (~(slym wa *worm) vase.vane *vane-sample)
       ::  cache the access of the %scry arm
       ::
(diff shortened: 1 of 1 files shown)


belisarius222:
I tested this by running |reset and doing some scrys. They all worked.

ixv:
Do we want this in next-ota?

belisarius222:
I think so

—
~rovnys-ricfer
https://urbit.org

PR #1018, opened 2019-01-18
Update talk ;help URL

Fang-:
Fixes #1016.

(idk why it's under arvo internals, but we can update this as the docs evolve)

PR #1019, opened 2019-01-18
Remove trailing whitespace & put check in CI

Fang-:
If we don't do it now we'll do it never.

May want to merge this into a different branch than master, depending on whatever decision comes out of discussion in #1018.


joemfb:
I'm embarrassed to see some of my code in here.

Fang-:
The best version of this PR makes CI fail when you're trying to check in trailing whitespace, but idk if that's desirable.

joemfb:
I'd approve that.

Fang-:
[That seems to work!](https://travis-ci.org/urbit/arvo/builds/481232890#L1043)

Would moving the test to earlier in CI (ie `before_install`) comply with common expectations of the Travis semantics?

joemfb:
I think that would be fine -- this check seems best placed as early as possible.

Fang-:
Alright, it's now the very first thing that's checked.

PR #1027, opened 2019-01-28
+has:in typecheck hack

belisarius222:
`+has:in` now ensures type safety on the element. It ain't pretty, but it works. Also keep in mind that this function is jetted, so its internal mischief won't cause performance issues.

cc @eglaysher

### DIFF
--- a/app/hall.hoon
+++ b/app/hall.hoon
@@ -498,7 +498,7 @@
       ~/  %hall-action-create
       |=  {nom/name des/cord typ/security}
       ^+  ..ta-action
-      ?.  (~(has in stories) nom)
+      ?.  (~(has by stories) nom)
         %^  impact  nom  %new
         :*  [[[our.bol nom] ~] ~ ~]
             des
@@ -516,7 +516,7 @@
       ::
       ~/  %hall-action-design
       |=  {nom/name cof/config}
-      ?.  (~(has in stories) nom)
+      ?.  (~(has by stories) nom)
         (impact nom %new cof)
       (ta-evil (crip "{(trip nom)}: already exists"))
     ::
@@ -2654,15 +2654,16 @@
     ==
   ::
     ?:  =(wat ~)  &
-    %-  ~(has in wat)
-    ?+  -.det  %hasnot
-      $gram     %grams
-      $new      %config-l
-      $remove   %config-l
+    ?+  -.det   |
+      $gram     (~(has in wat) %grams)
+      $new      (~(has in wat) %config-l)
+      $remove   (~(has in wat) %config-l)
       $config   ?:  =(cir.det [our.bol nom])
-                %config-l  %config-r
+                  (~(has in wat) %config-l)
+                (~(has in wat) %config-r)
       $status   ?:  =(cir.det [our.bol nom])
-                %group-l  %group-r
+                  (~(has in wat) %group-l)
+                (~(has in wat) %group-r)
     ==
   ==
 ::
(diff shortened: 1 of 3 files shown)


sys/hoon.hoon
```diff
@@ -1335,10 +1335,26 @@
...
+    ::
+    ::    However, bunting that expression fails if :a is ~. If we wrap it
+    ::    in a unit, the bunted unit doesn't include the bunted item type.
+    ::
+    ::    This way we can ensure type safety of :b without needing to perform
+    ::    this failing bunt. It's a hack.
+    ::
+    %.  [~ b]
+    |=  b=(unit _?>(?=(^ a) n.a))
+    =>  .(b ?>(?=(^ b) u.b))
```
joemfb: Yikes!

It seems like all the other arms of `+in` or `+by` that use the `_?>(?=(^ a) n.a)` pattern have this same problem. But is it really a problem?  Do we need to be able to default them?

belisarius222: It's definitely a problem that a bunch of our standard library doesn't enforce type safety. Not all those arms have this problem. Something mysterious happens when the `_?>(?=(^ a) n.a)` is inside the sample definition for a wet gate...

joemfb: I was thinking of `+gas:in`, but that sample is a list and can be defaulted without issue. Something like this pattern is probably necessary for the arms of a treap engine that operate directly on individual items. It would certainly be nicer if it could be more encapsulated somehow.

Not sure what you mean about the wet gate sample -- why would it need to be wet if you have a specific structure like that?

belisarius222: I don't know! Seems like samples of wet gates don't get bunted or something.

ohAitch:
Personally I'd put in a  ?: =(~ a) &  to handle the null case without any
type implications

sys/hoon.hoon
belisarius222: Or maybe wetness has nothing to do with it. The assertion in `+turn` probably only works because it works on a list. Please disregard my earlier comments ...

belisarius222:
@ohAitch feel free to make a PR

ohAitch:
I'm accumulating a number of those aren't I

I think in that case I'd go with the more assert-shaped

+-  has
  |*  b=*
  =+  ~:[?~(a ~ ^+(n.a b))]
  etc...

(consider this a note to self)

ohAitch:
(Or okay ?> &:[…] probably conveys meaning better there)

belisarius222:
Fixed the call sites, including a bug in the new printer (cc @benjamin-tlon) and a couple in hall (cc @Fang-). Tests pass now, so feel free to re-review.

app/hall.hoon
```diff
@@ -2654,15 +2654,16 @@
     ==
   ::
     ?:  =(wat ~)  &
-    %-  ~(has in wat)
```
Fang-: Is this just a style change (fair!) or did this actually break?

belisarius222: It broke, because `%hasnot` isn't a valid element of the set. The other two changes to hall used `in` where they should have been using `by`, which was also caught by the typechecker.

@ohAitch I don't like messing with the type of the map inline to add a special sentinel value that gets used in one place just so that we can add a call to `+has` that will always fail.

ohAitch: It "broke", in that the code was entirely dynamically valid, but nonetheless not entirely following all of the type rules.

I agree that it's ugly, it just seemed less ugly; probably the special sentinel should be named `%other` and formally in the type, if possibly never set. (`circle-data` is what the one-letter subscription modes turned into right?)

ohAitch: Actually looking at the line above, it looks `%other` effectively *is* set when `wat` is null(denoting "all"); which suggests it might not be entirely useless in the API proper.

belisarius222:
@joemfb ready for re-review

belisarius222:
@joemfb rebased off "next" branch

PR #1028, opened 2019-01-29
added functionality to call |mass on a timer

ixv:
run
> |automass ~m10


belisarius222:
Looks good other than that.

vvisigoth:
> Let's add a way to turn this off, too.

coward!

PR #1036, opened 2019-01-31
fix self memory printing in dill |mass

belisarius222:
This resets `hef.all` before printing |mass, so that other vanes' memory reports don't get absorbed into Dill's.

### DIFF
--- a/sys/vane/dill.hoon
+++ b/sys/vane/dill.hoon
@@ -378,7 +378,7 @@
         ^+  +>
         ?-    sih
             {?($a $b $c $e $f $g $j) $mass *}
-          (wegt -.sih p.sih)
+          (wegh -.sih p.sih)
         ::
             {$a $nice *}
           ::  ~&  [%take-nice-ames sih]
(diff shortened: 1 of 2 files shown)


sys/vane/dill.hoon
```diff
@@ -459,9 +451,18 @@
                 ?=(^ j.hef.all)
             ==
           +>.$
-        %+  done(hef.all [~ ~ ~ ~ ~ ~ ~])
+        ::
+        =/  ven  hef.all
+        =>  .(hef.all [~ ~ ~ ~ ~ ~ ~])
```
joemfb: I this would be clearer as a =., but that presumably nest fails.

sys/vane/dill.hoon
```diff
@@ -459,9 +451,18 @@
...
+        ::
+        %+  done
           %mass
-        =>  [hef.all d=wegh]
+        =/  d=mass  :+  %dill  %|
```
joemfb: Could you add a comment here mentioning that `d` must be computed after `hef.all` has been cleared? Also, I think the noun would be clearer on a newline.

sys/vane/dill.hoon
```diff
@@ -459,9 +451,18 @@
...
+                    :~  hey+&+hey.all
+                        dug+&+dug.all
+                        dot+&+all
+                    ==
+        =,  ven
         [%vanes %| ~[u.a u.b u.c d u.e u.g u.f u.j]]
```
joemfb: I don't really like the `=,` here. Maybe `~[u.a u.b ...]:[ven .]`, although that might be confusing in it's own right. Alternately, you could `=.  u.d` instead of `=/  d` above, then `~[u.a u.b ...]:ven` would just work.

sys/vane/dill.hoon
belisarius222: done

sys/vane/dill.hoon
belisarius222: I replaced the offending `=,` with a mutation into the result data structure.

PR #1038, opened 2019-01-31
dojo: move parsers to separate core

belisarius222:
Moved and renamed the `|dp` core to an outer `|parse-at` core. Also created a little `|at` core for path helpers, which are still a bit awkward.

Did some minor renaming and cleanup within the new parser core. Functionality should remain unchanged, and local testing seems to confirm that.


joemfb:
CI is breaking because `:dojo` is no miss-parsing `:hood` pokes like `|verb` or `|start`, and trying to poke some other ship over %ames. Not sure exactly what's going wrong with the parser yet.

ohAitch:
…could we please come up with either new dojo syntax prefixes, or
hoon-descriptive prefixes that aren’t +foo and |bar

PR #1039, opened 2019-01-31
Dojo wipe generator

belisarius222:
Add a `+poke-wipe` arm to Dojo so its state can be cleared. This can be fired by running the new `gen/dojo/wipe` generator and piping its results into the dojo, or it can be triggered externally in case we decide to run this on a timer or on low memory.


belisarius222:
I rebased this so it no longer depends on https://github.com/urbit/arvo/pull/1038. Shouldn't cause merge conflicts.

ohAitch:
Definitely -1 on the suggestion of the repl just like helpfully deleting
variables for you in the background, but otherwise looks good!

(Low memory might make any sense, but would still want to be a lot more
granular or at least require a prompt. At which point it could just be "run
=foo to avoid running out of memory" or similar)

PR #1040, opened 2019-02-01
build tank outside of |.

ixv:
Inside +execute-turbo in eyre, we moved the tank outside of the |. so that it's not held as a thunk. This stops ford from holding onto an 80MB error message in one of its caches.

### DIFF
--- a/sys/vane/eyre.hoon
+++ b/sys/vane/eyre.hoon
@@ -1158,9 +1158,10 @@
   ++  execute-turbo
     ~/  %execute-turbo
     |=  [tea=whir live=? request=schematic:ford]
+    =/  tan=tank  [%leaf "eyre: execute {(spud tea)}"]
     %+  pass-note  tea
     :*  %f  %build  live
-        [%dude |.([%leaf "eyre: execute {<tea>}"]) request]
+        [%dude =>(a=tan |.(a)) request]
     ==
   ::
   ++  add-links                                           :: x-urbit:// urls


joemfb:
+1 to wiping %ford on `|reset`. Or probably wiping once builds and rebuilding live builds.

belisarius222:
Oh, right, that trap has the scry function in it, so it effectively includes all the other vane states.

I'm trying to measure `cain` now to see how bad it would be for just that part of the kernel to hang on.

@joemfb I think you're right; Ford should wipe its cache and rebuild all live builds during a `+load`.

ohAitch:
Yeah, also interface-wise, having a result in the cache with a stale kernel
that will then be unpredictably evicted and produce a different noun for
the same request, seems kind of bad. (To some extent this is true
regardless if a clay-pinned external request is simply repeated, but at
least the discontinuity can happen at the |reset and not inconsistently
piece by piece over the following days to months)

Especially since if we have an old kernel in the keys, and have just
dropped all other references to said kernel, no further requests with those
keys can conceivably come in which is the whole point of caching :P

ohAitch:
I mean, ++cain is gong to use the pretty-printer, which wants ++ut, which
in turn is the heaviest part of hoon.hoon and probably the reason you're
installing a new one :P

belisarius222:
This is actually sort of interesting:
```
> (met 3 (jam cain))
899.010
> (met 3 (jam noah))
896.387
> (met 3 (jam noah cain))
896.394
> (met 3 (jam noah cain abel))
896.406
> (met 3 (jam noah cain abel onan))
896.411
> (met 3 (jam noah cain abel onan text))
896.412
> (met 3 (jam noah cain abel onan text ride))
896.419
> (met 3 (jam ride))
898.936
> (met 3 (jam ..ride))
897.987
```

ohAitch:
Looks like they have an approximately shared +>? I think I'm missing
something here, the <1% size differences are going to be variations in
back-reference index encoding lengths(i.e. xeb), all of which get bumped
right various amounts by the gate's actual formula(head)

belisarius222:
```
> (met 3 (jam ..add))   ::  layer 1
4.113
> (met 3 (jam ..in))    ::  layer 2
18.731
> (met 3 (jam ..egcd))  ::  layer 3
46.582
> (met 3 (jam ..po))    ::  layer 4
93.978
> (met 3 (jam ..ut))    ::  layer 5
897.987
```

ohAitch:
Right, while those are actually different cores. <foo> is layer 5 :P

belisarius222:
Yeah, I assume the small variations within a layer have to do with idiosyncrasies of the jam encoding. Over 90% of the weight of hoon.hoon is in layer five. Makes me think we should break that up into layers.

ohAitch:
Something like "compiler", "vases and mock and friends"? How much of it is
literally -:ut weight?

belisarius222:
About a quarter of it:
```
> (met 3 (jam -:ut))
247.735
> (met 3 (jam ..ut))
897.987
```

joemfb:
It looks to me like the `"{}"` expands to a relatively straightforward list reduction at compile time. I don't expect evaluating such an expression to invoke the compiler at runtime.

ohAitch:
The {} is not the expensive part, it's the spud/<tea>

(The latter of which uses the type to guide printing)

joemfb:
`(spud /foo/bar)` is just `~(ram re (smyt /foo/bar))`, and `+smyt` is just path to tank. The whole thing is completely statically typed, and looks like it should be low overhead.

belisarius222:
```
=/  =path  /~zod/home/~2000.1.1/sys/hoon/hoon  "some error string prefix {(spud path)}"
```
takes an average of 20 microseconds on my machine, using a test of 500.000 iterations.

belisarius222:
```
=/  =path  /~zod/home/~2000.1.1/sys/hoon/hoon  "some error string prefix {<path>}"
```
takes a little over 100 microseconds, so about five times slower.

belisarius222:
Running the `"({spud path)}"` version inside a new `+mule` each time slows it down from 20 microseconds to 40, although `+mule` that just returns the precalculated string takes 10 microseconds. Seems like there's an extra 10 microseconds of warmup of some kind in the child road.

PR #1041, opened 2019-02-01
notify vanes of kernel upgrades

joemfb:
This PR updates arvo to send a `[%vega ~]` task to every vane after a kernel upgrade (and after the %zuse and the vanes have been recompiled as well).

Right now, the task handlers in each vane are just stubs. I think the most pressing feature is to make %ford clear it's cache and rebuild all live builds. This PR could go in as is, or it could be held open until the task handlers themselves are also implemented.

Note that this change creates an awkward naming collision in %dill, which already handles a %vega task (to trigger the kernel upgrade). I've handled this for now with an ugly type fork. Eventually, the different moves that together make up the kernel upgrade procedure should probably have different names.

### DIFF
--- a/lib/hood/helm.hoon
+++ b/lib/hood/helm.hoon
@@ -176,9 +176,9 @@
   =+  way=?:(zus (welp top /sys/[nam]) (welp top /sys/vane/[nam]))
   =+  fil=.^(@ %cx (welp way /hoon))
   [%flog /reload [%veer ?:(=('z' tip) %$ tip) way fil]]
-::  +poke-reset:  send %vega to reboot kernel
+::  +poke-reset:  send %lyra to initiate kernel upgrade
 ::
-::    And reinstall %zuse and the vanes.
+::    And reinstall %zuse and the vanes with %veer.
 ::    Trigger with |reset.
 ::
 ++  poke-reset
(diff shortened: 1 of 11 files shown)


ohAitch:
%vegd

eglaysher:
Seconding giving it a different name.

joemfb:
Well, we actually already have this dual-meaning. `%dill` accepts a `[%vega @t @t]` task, and uses the same structure as a gift to initiate an upgrade (the `@t`s are hoon and arvo as source). After upgrade is successful, a `[%vega ~]` effect is produced at the top level. Vere prints `<<<reset>>>>` and clears some caches on this effect. This change doesn't impact any of that overall flow, it just adds some task-punning to %dill's interface.

I agree that the whole is somewhat unpleasant. It seems best to change the initiating task/gift in %dill, and preserve `[%vega ~]` as the notification that goes to all vanes and out to unix.

belisarius222:
Yes, let's target this at hotfix. More likely to help than hurt.

belisarius222:
We should probably merge it into next, too

joemfb:
Rebased/re-targeted. The only real change was in %behn.

belisarius222:
LGTM again

PR #1044, opened 2019-02-01
behn: ignore duplicate %wait requests

belisarius222:
Same fix as https://github.com/urbit/arvo/pull/1043, but made against the refactored behn in the `next` branch. @joemfb I hope this is following hygienic git practices


eglaysher:
A question: is it an error condition to make multiple timers with the same date on the same duct? Does this open the possibility of multiple timer cancelations, too? I suspect that if another vane is making multiple timer requests, that it will have significant state around this which won't be mirrored in %behn.

joemfb:
I think duplicating PRs across maintained branches is not a sustainable approach. It's a lot of extra overhead, easy to miss or just forget in some cases, and probably just substitutes one kind of future merge conflict for another.

@eglaysher, I think it should generally be considered an error condition. From the hotfix perspective, it's probably right to just swallow the error for now. But the fact that this is happening at all points to problems in the vane setting the timer.

belisarius222:
@eglaysher Behn ignores duplicate %wait requests and (now) also duplicate %rest requests, so this should be safe. We could also make it an error condition, but I'm reluctant to build too much machinery enforcing how we do or don't reuse ducts until we determine how we want that to work more generally.

belisarius222:
Closing in favor of a later merge from the 'hotfix' branch.

PR #1047, opened 2019-02-02
Ames Behn smart

belisarius222:
Ames now stores its timers in Behn instead of itself, avoiding some costly traversals through its own state.

I've done some local testing with a couple of galaxies, which seems to work, but nothing extensive. We should definitely test it more before deploying to livenet.

This should have almost exactly the same behavior as the old Ames. The only behavior I've knowingly changed was removing the `~s32` default for the wakeup timer, which used to be calculated in the moribund `+next` arm. I'm not confident it's ok to remove that, but there is no longer a global Ames timer to update, so I'm not sure how to translate that to the new design.

### DIFF
--- a/sys/vane/ames.hoon
+++ b/sys/vane/ames.hoon
@@ -657,7 +657,7 @@
       ~|  %ames-bad-behn-wire^wire
       ?>  ?=([@ ~] wire)
       =/  her=ship  `@p`(slav %p i.wire)
-      ::  send relevant packets, updating internal state1
+      ::  send relevant packets, updating internal state
       ::
       zork:zank:thaw:(ho:um her)
     ::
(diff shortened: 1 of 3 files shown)


joemfb (commented):
The code itself looks good, and I think this is an overall improvement.

But I'm concerned about the removal of the global timer and `~s32` default. It seems like peer discovery might be arbitrarily delayed without galaxies and stars frequently attempting to re-establish routes to their sponsees. I'm disinclined to merge this until we've figured out the impact of removing that frequent route maintenance (although the old/current way does seem to be too expensive, and to be run too frequently).

belisarius222:
@ohAitch Look, we've got to keep this productive. Please stop polluting the repos with useless comments

belisarius222:
Transferring work from this to cc-release. Closing this PR.

PR #1049, opened 2019-02-03
scavenge some dead printing code from dojo

belisarius222:
Removes a bunch of dead code from dojo dealing with printing. Also removed the `%help` functionality. We'll want to build a replacement for that eventually, but most of that code was already commented out anyway.


ohAitch:
I love it when people remove help functionality!

PR #1051, opened 2019-02-03
Provide better json-rpc request affordances

Fang-:
This adds a type for json-rpc requests, and provides a few utility functions for turning it into more directly useful things (`json`, `hiss:eyre`).

Also removes a `/sur/json/rpc` that was still there, even though the json-rpc structures have lived in zuse for a while already.

Yes, the latter `++jrpc` shadows the former. This seems like a more elegant solution than separating types across cores, as we do for some `++ethereum`  and `++azimuth` things.

"This belongs in a lib!" json-rpc support is already in zuse though, and needs to be because jael and dawn depend on it. Having it only partially in the stdlib feels even weirder. Happy to discuss this though.

### DIFF
--- a/mar/json/rpc/response.hoon
+++ b/mar/json/rpc/response.hoon
@@ -1,6 +1,5 @@
 ::
-/-  json-rpc
-=,  json-rpc
+=,  rpc:jstd
 ::
 |_  res=response
 ::
--- a/sur/json/rpc.hoon
+++ b/sur/json/rpc.hoon
@@ -1,8 +1,0 @@
-|%
-++  response  ::TODO  id should be optional
-  $%  [%result id=@t res=json]
-      [%error id=@t code=@t message=@t]  ::TODO  data?
-      [%fail hit=httr:eyre]
-      [%batch bas=(list response)]
-  ==
---
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -95,7 +95,17 @@
   |%
   ++  rpc
     |%
-    ++  response  ::TODO  id should be optional
+    +$  request
+      $:  id=@t
+          method=@t
+          params=request-params
+      ==
+    ::
+    +$  request-params
+      $%  [%list (list json)]
+          [%object (list (pair @t json))]
+      ==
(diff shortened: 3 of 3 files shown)


Fang-:
> Is there an early-boot dependency

That's what [dawn](https://github.com/urbit/arvo/blob/611b9261b831f53c8706a8de77502bd9baadd681/sys/zuse.hoon#L8170) is! During boot we send a couple RPC requests to verify our own keys, and to get keys for the galaxies, so that we may start talking to the network.  
Afterwards, for the I-use-my-own-node cases (galaxies, and any other ships that want to do so), jael is continuously speaking to an Ethereum RPC node to stay up to date.

I agree this feels like it belongs in userspace. The most we can do to that end is split up the structures and logic between zuse and userspace, but as I mentioned that would lead to a very artificial separation of otherwise related components...

joemfb:
I think it's fine for this to be in %zuse now, given the reasons you've stated.

@ohAitch, in the very long term, I suspect it would be best for `|dawn` pre-boot validation to be entirely implemented in the runtime. But there's a lot that goes into Ethereum RPC request generation, de-/serialization, etc., and I have no interest in writing that in C right now.

ohAitch:
Right, I guess my question is why is this an *early* boot dependency(pre-userspace), as opposed to a *late* boot dependency(you have a hoon repl and stuff, but functioning networking is blocked on people accepting your packets obvs, so you're not technically finished booting until the ETH whatever returns some other packets)

If the former is the case, there is a natural separation betwee "json"/"endpoint" part logic in userspace, and the actual commands(`abi:ethereum`?) the vanes can make over a card interface.

joemfb:
It seems like a significantly better experience to validate the users keys before booting, rather than booting and then figuring out that we're in an invalid state. Also, it seems right to boot with galaxy public keys, instead of booting and then trying to get them. And the same applies to the PKI snapshots that are used by default.

At the same time it seems suboptimal to have all these things in the standard library. I think the pre-boot validation itself, and the specifics of how it's implemented are both worth revisiting down the road.

PR #1052, opened 2019-02-03
Minimum viable Bitcoin RPC support

Fang-:
This includes a `bitcoin` `sur` and `lib` for dealing with Bitcoin RPC requests and responses.  
I've included a simple `/app/bitcoin-cli` that uses the lib to forward RPC call nouns to an actual node. Very bare-bones, with hard-coded URL and all... [but it works!](https://twitter.com/mdfang/status/1092171862178701313)

This PR is easily considered incomplete, but I'm opening it for discussion to support further development. Maybe we're fine merging this as-is though? It works, and "support additional Bitcoin RPC calls" might be a good contributor issue.

It'd be cool if we could send transactions in whatever way. I think my best bet it to talk to @loganallenc about what the deal is with the accounts/wallets and transactions interfaces.  
I vaguely remember there being auras for bitcoin addresses, does anyone know what they are?

The request-tag-as-json-id pattern here feels pretty nice. We might want to consider using that for Ethereum RPC integration as well, as a default, but I haven't thought it over much yet.

Includes/depends on #1051.

### DIFF
--- a/app/bitcoin-cli.hoon
+++ b/app/bitcoin-cli.hoon
@@ -1,0 +1,103 @@
+::  minimum viable bitcoin app
+::
+::    first, run |init-auth-basic with bitcoin-rpc as the domain (no .tld)
+::    to configure your rpc username and password.
+::    then, poke this app with either an %app command or %rpc request.
+::
+/+  *bitcoin, basic-auth
+::
+|%
+++  task
+  $%  [%rpc req=request:rpc]
+      [%app cmd=command]
+  ==
+::
+++  command
+  $%  [%select-wallet name=(unit @tas)]
+  ==
+::
+++  state
+  $:  wallet=(unit @tas)
+  ==
+::
+++  move  (pair bone card)
+++  card
+  $%  [%hiss wire (unit user:eyre) mark %hiss hiss:eyre]
+  ==
+::
+--
+::
+|_  [=bowl:gall state]
+++  prep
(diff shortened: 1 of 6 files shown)


Fang-:
> this will require running a bitoin core node locally?

This will require RPC access to any node. It doesn't need to be local. But I got the impression (from hasty searches) that public RPC-enabled nodes aren't at all common in Bitcoin?

> I wish urbit could do bitcoin crypto internally

Generating and signing raw transactions seemed too complicated for MVP work (you need to manually specify inputs and outputs? I haven't looked into this too deeply either), but I do definitely want to support the "wallets on urbit" case. We already have [ECDSA](https://github.com/urbit/arvo/blob/611b9261b831f53c8706a8de77502bd9baadd681/sys/zuse.hoon#L4007), [BIP32](https://github.com/urbit/arvo/blob/master/lib/bip32.hoon) and [BIP39](https://github.com/urbit/arvo/blob/master/lib/bip39.hoon) implementations, so most of the groundwork is out of the way. It's just a matter of applying that to actual transactions.

I think you might still need at least a local light node in that case, unless I'm wrong about the public RPC nodes thing. (There's probably also services offering plain ol' web APIs, but those aren't as interesting.)

dpc:
I've shoot you an email @Fang- . TTYL

Fang-:
Yeah very fair. I will return!

PR #682, opened 2018-03-20
Add |invite support for up to sixty-four ships

keatondunsford:
Our |invite Hood generator for Womb invites currently caps at twenty
ship invites. This commit bumps that hard-coded cap to sixty-four, which
will allow me to send the appropriate number of star invites to our
[Winter 2018 contributor-award recipients](https://fora.urbit.org/posts/~2018.2.13..20.22.22..bddb~).

PR #1053, opened 2019-02-04
Add library for converting atoms to English

### LINKED ISSUE #683: Arvo doesn't have a library for converting atoms to English
As discussed with @ohAitch in #682, it'd be nice to have a `/lib` file that provided this utility function of converting Hoon atom digits to their English text values.

For example, something like:

```
~zod:dojo> /+  to-english
>=
~zod:dojo> (atom:to-english 42)
"forty-two"
```

rovdyl:
Resolves #683, and therefore also works as a possible replacement for #682. It is essentially a port of https://www.npmjs.com/package/number-to-words, as suggested by @ohAitch, but expanded to work with even larger numbers according to what I could find at https://en.wikipedia.org/wiki/Names_of_large_numbers. I currently only have English implemented but tried to leave room for other possible languages. It returns a unit because it only supports up to vigintillion, and past that returns null.

I'm also fairly new to all this so apologies if there are any problems, but I tried to keep it relatively simple and it worked in all of my tests. The main issue I can think of is that it uses `++need`, but it's used in a way that should be guaranteed to be safe and never *actually* crash. Still, though, if anyone has any suggestions for how to do it more safely they are very welcome.


lib/number-to-words.hoon
```diff
@@ -0,0 +1,105 @@
...
+  ?:  (lth num one-undecillion)  (num-break one-decillion "decillion" ", ")
+  ?:  (lth num one-duodecillion)  (num-break one-undecillion "undecillion" ", ")
+  ?:  (lth num one-tredecillion)  (num-break one-duodecillion "duodecillion" ", ")
+  ?:  (lth num one-quattuordecillion)  (num-break one-tredecillion "tredecillion" ", ")
+  ?:  (lth num one-quindecillion)  (num-break one-quattuordecillion "quattuordecillion" ", ")
+  ?:  (lth num one-sexdecillion)  (num-break one-quindecillion "quindecillion" ", ")
+  ?:  (lth num one-septendecillion)  (num-break one-sexdecillion "sexdecillion" ", ")
+  ?:  (lth num one-octodecillion)  (num-break one-septendecillion "septendecillion" ", ")
+  ?:  (lth num one-novemdecillion)  (num-break one-octodecillion "octodecillion" ", ")
+  ?:  (lth num one-vigintillion)  (num-break one-novemdecillion "novemdecillion" ", ")
```
Fang-: Some of these lines near the end are >80 characters. We try to stay <80 throughout the codebase.

lib/number-to-words.hoon
```diff
@@ -0,0 +1,105 @@
...
+  =/  num-break                                         ::  repeated pattern
+    ::  min: minimum to qualify for this break
+    ::  str: english word for this break
+    ::  sep: separator between this and the previous break
+    |=  [min=@u str=tape sep=tape]
+    =/  rem  (mod num min)
+    :-  ~
+    %+  weld
+      (weld (need ^$(num (div num min))) [' ' str])
+    ?:(=(rem 0) ~ (weld sep (need ^$(num rem))))
```
Fang-: This line feels a bit crammed. Tall-forming the `?:` would probably help readability.

rovdyl:
Thanks for the help! Some of these I had actually considered but decided against for personal taste, but for the sake of consistency I'll go ahead and fix them.

> I believe you! But for bonus points, you can try writing automated tests in `/tests/lib/number-to-words.hoon`, then running them with `+test /lib/number-to-words`. (:

I've been caught! I've been meaning to look into how urbit does tests, anyway, so I'll go ahead and try to write up these up.

philipcmonk:
The usages of `++need` look fine to me since they work as long as you maintain the invariant that the new `num` is less than the old `num`, which you do.  The other option would be explicitly put the "greater than vigintilion" case in the beginning, then put a `:-  ~  |-` after that case so that the rest produces a guaranteed tape.

tests/lib/number-to-words.hoon
```diff
@@ -0,0 +1,68 @@
...
+      !>
+        ^-  (unit tape)
+        :-  ~
+        ;:  weld
+          "eighteen quintillion, four hundred and forty-six quadrillion, seven "
+          "hundred and forty-four trillion, seventy-three billion, seven "
+          "hundred and nine million, five hundred and fifty-one thousand, six "
+          "hundred and sixteen"
+        ==
+        !>  (eng:number-to-words 18.446.744.073.709.551.616)
```
Fang-: Indentation. (;

joemfb:
Thanks, @rovdyl! I re-targeted this PR to the `next` branch, as it's not a high-priority, over-the-air release.

rovdyl:
> The Brits certainly don't!

Interesting, good catch.

I'll go ahead and reorganize it and fix that indentation. I would like to stick with the three letter language codes like `eng-us` and `eng-gb`, though, just so as to not get anything confused with the `en` and `de` for encoding/decoding I've seen used elsewhere in arvo, unless you'd really prefer the two letters.

Fang-:
Nah, either is fine. Both `en` and `eng` are fair game here according to the [IETF language tag](https://www.wikiwand.com/en/IETF_language_tag) (which I was thinking of adhering to for this) standard, which I was thinking of using for cases like this.

rovdyl:
Alright, I think I've done enough bikeshedding for now. If everyone else is happy with where things are I'm good with it being merged.

Thanks for holding my hand through my first attempt at actually contributing something to this project! Hope I didn't eat up too much of anyone's time.

Fang-:
> Hope I didn't eat up too much of anyone's time.

Hah, no way! Thanks so much for your contribution, I hope we'll be seeing you around. (:

PR #1056, opened 2019-02-05
!: in Ford, except inner loops

belisarius222:
This turns on debugging in all of Ford except the `+execute-loop`, `+gather`, and `+reduce` loop arms, which need to maintain their TCO.

### DIFF
--- a/sys/vane/ford.hoon
+++ b/sys/vane/ford.hoon
@@ -1771,8 +1771,7 @@
   ::    to wake us up immediately. This has the advantage that Ford stops hard
   ::    blocking the main Urbit event loop, letting other work be done.
   ::
-  ++  execute-loop
-    !.
+  ++  execute-loop  !.
     ~/  %execute-loop
     |=  builds=(set build)
     ^+  ..execute
@@ -1793,7 +1792,6 @@
   ::    +build-receipt algorithms to the ford state.
   ::
   ++  execute
-    !:
     ~/  %execute
     |=  builds=(set build)
     ^+  ..execute
@@ -1810,8 +1808,7 @@
   ::    should run the candidate build this cycle through the +execute loop, we
   ::    place it in :next-builds. +gather runs until it has no more candidates.
   ::
-  ++  gather
-    !.
+  ++  gather  !.
     ~/  %gather
     |=  [builds=(set build) force=?]
     ^+  ..execute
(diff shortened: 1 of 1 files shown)


sys/vane/ford.hoon
```diff
@@ -1792,6 +1793,7 @@
   ::    +build-receipt algorithms to the ford state.
   ::
   ++  execute
+    !:
```
joemfb: Is this one strictly necessary? Looks like it'd already be set by the top-level `!:`.

belisarius222: Well, this only gets called by `+execute-loop`, which is `!.`ed. Not sure which takes precedence, so I figured I'd err on the safe side and explicitly enable debugging here.

belisarius222:
I don't have a strong stylistic preference, so I'll change them to match the standard.

ohAitch:
!: is lexically scoped(well, works at parse-time not runtime) so this
shouldn't be any different than e.g. calling ++turn on a non !. gate

sys/vane/ford.hoon
joemfb: Per a quick test in dojo, it should be safe to remove:

```
> =<  foo  |%  ++  foo  !.  bar  ++  bar  !!  --
```

```
> !:  =<  foo  |%  ++  foo  !.  bar  ++  bar  !!  --
/~zod/home/~2019.2.5..22.34.32..70c1:<[1 5].[1 51]>
/~zod/home/~2019.2.5..22.34.32..70c1:<[1 9].[1 12]>
/~zod/home/~2019.2.5..22.34.32..70c1:<[1 27].[1 34]>
/~zod/home/~2019.2.5..22.34.32..70c1:<[1 45].[1 47]>
```

```
> !:  =<  foo  |%  ++  foo  !.  bar  ++  bar  !:  !!  --
/~zod/home/~2019.2.5..22.34.52..c489:<[1 5].[1 55]>
/~zod/home/~2019.2.5..22.34.52..c489:<[1 9].[1 12]>
/~zod/home/~2019.2.5..22.34.52..c489:<[1 27].[1 34]>
/~zod/home/~2019.2.5..22.34.52..c489:<[1 45].[1 51]>
/~zod/home/~2019.2.5..22.34.52..c489:<[1 49].[1 51]>
```

belisarius222:
Ok, I updated the style and removed the superfluous `!:`s. Thanks @ohAitch

Issue #1033, opened 2019-01-30
Dropped into talk after booting a ship


baudtack:
Booting a fakezod

``urbit -F zod``

results in being dropped into talk now instead of the dojo.

PR #1063, opened 2019-02-09
Make a new ship launch into dojo

pilfer-pandex:
Currently, when you launch a new ship, after initialization, you are dumped into talk. This change makes it so you are dumped into dojo instead.

The thing you are dumped into is a consequence of sort order over the `ray` set defined in `lib/hood/drum.hoon`. Under the previous `mug` algorithm, it happened that `dojo` was first.

This fixes issue #1033.

### DIFF
--- a/.travis/test.js
+++ b/.travis/test.js
@@ -83,16 +83,6 @@ function barMass(urb) {
 }
 
 Promise.resolve(urbit)
-//  XX temporary
-//  send ctrl-x to select dojo
-//
-.then(function(){
-  return urbit.expect(/talk\[\] /)
-  .then(function() {
-    return urbit.pty.write("\x18")
-  })
-  .then(function() { return urbit })
-})
 .then(actions.safeBoot)
 .then(function(){
   return barMass(urbit);
--- a/lib/hood/drum.hoon
+++ b/lib/hood/drum.hoon
@@ -99,12 +99,12 @@
   ^-  part
   :*  %drum
       %2
-      ~                                                 ::  sys
-      (deft-fish our)                                   ::  eel
-      (deft-apes our)                                   ::  ray
-      ~                                                 ::  fur
-      ~                                                 ::  bin
-  ==                                                    ::
+      sys=~
+      eel=(deft-fish our)
+      ray=(deft-apes our)
+      fur=~
+      bin=~
+  ==
 ::
 ::
 ++  en-gill                                           ::  gill to wire
@@ -245,7 +245,9 @@
 ::
 ++  se-adit                                           ::  update servers
   ^+  .
-  %+  roll  ~(tap in ray)
+  ::  ensure dojo connects after talk
+  =*  dojo-on-top  aor
+  %+  roll  (sort ~(tap in ray) dojo-on-top)
   =<  .(con +>)
   |:  $:{wel/well:gall con/_..se-adit}  ^+  con
   =.  +>.$  con


joemfb:
I spent a while trying to make apps show up in the drum in the order they were started, but that would have required more disruptive changes to %gall. This seems like the right approach for now.

The CI builds passing here are spurious successes, though. CI is hardcoded to ctrl-x on boot, and all the test commands just ran as :talk messages. Please remove the following entirely from `.travis/test.js` to fix this:

```
//  XX temporary
//  send ctrl-x to select dojo
//
.then(function(){
  return urbit.expect(/talk\[\] /)
  .then(function() {
    return urbit.pty.write("\x18")
  })
  .then(function() { return urbit })
})
```

Note that this will also need to be changed in the urbit/urbit CI, once these changes make it into `./travis/pinned-arvo-commit.txt` in that repo.

ohAitch:
Hah, anyone want to hack in a "+http://localhost:4321/test-get" dojo
test step, with the test framework opening port 4321 and failing the
test unless it receives the get request? (or if the test response
isn't accurately printed, maybe throwing in some unicode and newlines
and stuff)

On 2/8/19, Joe Bryan <notifications@github.com> wrote:
> I spent a while trying to make apps show up in the drum in the order they
> were started, but that would have required more disruptive changes to %gall.
> This seems like the right approach for now.
>
> The CI builds passing here are spurious successes, though. CI is hardcoded
> to ctrl-x on boot, and all the test commands just ran as :talk messages.
> Please remove the following entirely from `.travis/test.js` to fix this:
>
> ```
> //  XX temporary
> //  send ctrl-x to select dojo
> //
> .then(function(){
>   return urbit.expect(/talk\[\] /)
>   .then(function() {
>     return urbit.pty.write("\x18")
>   })
>   .then(function() { return urbit })
> })
> ```
>
> Note that this will also need to be changed in the urbit/urbit CI, once
> these changes make it into `./travis/pinned-arvo-commit.txt` in that repo.
>
> --
> You are receiving this because you are subscribed to this thread.
> https://github.com/urbit/arvo/pull/1063#issuecomment-462010903


-- 
Here lies a toppled god —
His fall was not a small one.
A narrow and tall one.

pilfer-pandex:
Goodness, thanks for pointing that out Joe. How would I go about adding an expectation that we start in dojo to that test? Naively, I'd think I could change the lines you mentioned to

```
.then(function(){
  return urbit.expect(/dojo> /)
  .then(function() { return urbit })
})
```

but the `urbit.expect(/talk\[\] /)` in the thing I just deleted did *not* cause the test to fail, so I don't see why the thing I proposed above would fail on master.

Anton, not sure I understand your suggestion.

ohAitch:
I don't have a repl at hand, but in broad terms:
- there's a dojo syntax to make an HTTP request
- the Travis test framework is in node,js, where the event loop makes
starting and stopping http servers rather trivial
- most current tests are of the form "input a dojo command, check the shell
output"
- a test that could be added is "start one-off http server on localhost,
input into dojo the command to ping the localhost http server, wait for the
ping to arrive(or a timeout), if it doesn't then you're typing into chat or
the c runtime is having problems making http requests or something"

Probably just typing (add 40 2) and waiting for a 42 to pop up would catch
literally the "we're in talk" case, tbf

joemfb:
The "safeBoot" action was supposed to be this: https://github.com/urbit/runner-js/blob/master/lib/actions.js#L7, but an echo is a bad test given the talk app's behavior. I'm not sure why the `expect(/dojo\>/)` call isn't enough ... The HTTP request thing would work, but so would any hoon expression.

@benjamin-tlon is experimenting with `urb.py` as a test harness, which will probably be a more robust approach.

ohAitch:
Right, that's why you want an (add 40 2) instead of just expectEcho

My best guess for 'dojo> ' is that it briefly appears, but is then
overwritten by the talk prompt; is it tested for *after* the echo?

urb.py (or at any rate its JSON api) is definitely the desired endgame,
especially now that dæmon mode is being revived C-side - last I checked
it's missing important features like "report an error caught at any layer
of the system" i.e. capturing ~& output. (At least to a logfile! Which we
should be doing anyway tbh)

PR #1064, opened 2019-02-09
Don't fire timer in the middle of another event and take one timer at a time

philipcmonk:
While debugging issues in aquarium, I changed these behaviors of behn.  With these changes, I think it's now true that we fire no more than one timer in an event, and only if this is a %wake event straight from unix.  This should make it much easier to reason about behn's behavior.

There was never any good reason to fire a timer while setting another timer, right?  I think I was running into a situation where I'd have one thing set a timer for `now` to try to wait for the next event, and then another would do the same thing in the same event, which would fire the first in the same event.

It's super weird, btw, that we `(flop moves)` in several places rather than once in a `++abet`-style function.  I *think* the flops are right, but I really can't tell.

### DIFF
--- a/sys/vane/behn.hoon
+++ b/sys/vane/behn.hoon
@@ -59,9 +59,6 @@
   ++  wait
     |=  date=@da
     ^+  [moves state]
-    ::  process elapsed timers first to maintain sort order
-    ::
-    =.  event-core    notify-clients
     =.  timers.state  (set-timer [date duct])
     set-wake
   ::  +wake: unix says we should wake up; notify clients and set :next-wake
@@ -92,7 +89,7 @@
   ::
   ++  notify-clients
     =*  timers  timers.state
-    |-  ^+  event-core
+    ^+  event-core
     ::
     ?~  timers
       =.  moves  (flop moves)
@@ -102,10 +99,10 @@
       =.  moves  (flop moves)
       event-core
     ::
-    %_  $
-      timers  t.timers
-      moves   [[duct.i.timers %give %wake ~] moves]
-    ==
+    =.  moves  [[duct.i.timers %give %wake ~] moves]
+    =>  .(timers t.timers)
+    =.  moves  (flop moves)
+    event-core
   ::  +set-wake: set or unset a unix timer to wake us when next timer expires
   ::
   ::    We prepend the unix %doze event so that it is handled first. Arvo must


ohAitch:
<s>you could have a "bonus priority moves" variable, that you flop in
parallel with the normal moves, but then weld onto their head</s>

philipcmonk:
Seems like the way to handle that would have been to just swap the two `=.` lines in `++wait`.  Isn't that semantically exactly what we wanted to do?

It's my opinion that it's virtually always clearer to reframe `(quip moves _this)` as `_this` with a `moves` state entry.  Then you just make sure to just add stuff on to the front in the order you want it executed, and flop it all at once at the end.

For example, see `moves`, `++abet-pe`, `++abet-aqua`, and `++apex-aqua` here and how they make things like `++turn-plow` much cleaner than it would be if it had to juggle these moves explicitly (like it did in an earlier version).

https://github.com/urbit/arvo/blob/1a87a5c9fad7c96a45ee051eed4f962aa895d4af/app/aqua.hoon#L53

ohAitch:
I'm guessing that doesn't work because ++notify-clients conflates "pop next timer if any" and "actually notify its subscribers", which want to happen before and after set-wake respectively?

(also I'm probably missing something, but is there a reason it only grabs _1_ timer?)

belisarius222:
@philipcmonk That's a good idea in theory, but the way it's written right now, we don't know when the `%doze` should be set to until we've popped the first timer off the queue. We could rewrite that to calculate both things during one function if we want to maintain a stack discipline for the moves. I suspect that would make the logic harder to follow, though, since it would need to blend `+set-wake` and `+notify-client` into one more complex function.

Come to think of it, Behn should be even dumber:

- `+notify-clients` should be renamed to `+notify-client`, and it should error out if our timer queue is empty, or if its first element is in the future compared to `now`. It should unconditionally produce a single `%wake` move to a client.
- `+born` should just emit the `%doze` and exit. That way the timer firing will happen in a new event, which is important for transaction isolation. This also means `+notify-client` only needs to be called from `+wake`, so maybe it should be inlined.

`+set-wake` is very similar to `+abet`, so maybe it should be called with a `=<  set-wake` at the beginning of relevant arms to make that clearer. Again, we don't actually need to flop moves anymore because of the new one-timer-notification-per-Arvo-event semantics.

belisarius222:
@ohAitch We want each timer firing to happen in its own Arvo event so that downstream processing can fail separately. It's a way to isolate transactions from each other.

philipcmonk:
Okay, yeah, then "time to set the new timer" seems like another piece of global state that would get updated throughout the call, and then the `++abet` would make the actual `%doze` at the right end of the list.  Similar to `unix-effects` in aquarium.

In this case, though, I agree with "Behn should be even dumber".  Every call to Behn should produce either nothing, one `%doze`, or one `%wake` and one `%doze`.

ohAitch:
Ah, that makes sense yeah

The unix semantics issue mostly suggests a return to `++doze` - at least, in some limited sense, where behn sends an idempotent `[%doze ~]` which signals "at the end of this event, scry behn to figure out when the next timer actually is". I guess we might want a more general pattern that addresses identical issues coming up in userspace for subscription delivery?

ohAitch:
```
::  /sys/vane/behn
|_  [tym=(list [p=@da q=duct]) hen now etc]
++  wait
  |=  wen=@da
  =<  :_(tym [unix %doze ?~(tym ~ `p.i.tym)]~)
  |-  ^+  tym
  ?~  tym  [wen hen]~
  ?:  (lte wen p.i.tym)  [[wen hen] tym]~
  [i.tym $(tym t.tym)]
::
++  rest
  |=  wen=@da
  =<  :_(tym [unix %doze ?~(tym ~ `p.i.tym)]~)
  tym=(skip tym |=(a=[@da duct] =(a [wen hen])))
::
++  wake
  |=  ~
  =^  wak  tym
    |-  ^-  [wak=(list duct) tym=_tym]
    ?~  tym  [~ ~]
    ?:  (gth p.i.tym now) [~ tym]
    [[i wak] tym]:[i=q.i.tym $(tym t.tym)]
  ::
  :-  :-  [unix %doze ?~(tym ~ `p.i.tym)]
      (turn wak |=(a=duct [a %wake ~]))
  tym
--
```
kind of thing?

<details><summary>"scry-doze" variant that addresses the issue at hand</summary>

```
::  /sys/vane/behn
|_  [tym=(list [p=@da q=duct]) hen now etc]
++  scry-doze  ?~(tym ~ `p.i.tym)
++  wait
  |=  wen=@da
  :-  [unix %doze /bx/doze]~
  |-  ^+  tym
  ?~  tym  [wen hen]~
  ?:  (lte wen p.i.tym)  [[wen hen] tym]~
  [i.tym $(tym t.tym)]
::
++  rest
  |=  wen=@da
  :-  [unix %doze /bx/doze]~
  (skip tym |=(a=[@da duct] =(a [wen hen])))
::
++  wake
  |=  ~
  =^  wak  tym
    |-  ^-  [wak=(list duct) tym=_tym]
    ?~  tym  [~ ~]
    ?:  (gth p.i.tym now) [~ tym]
    [[i wak] tym]:[i=q.i.tym $(tym t.tym)]
  ::
  :-  :-  [unix %doze /bx/doze]
      (turn wak |=(a=duct [a %wake ~]))
  tym
--
```

</details>

ohAitch:
…wait okay is there anything the current 240-line implementation is doing that this isn't? Documentation presumably, I guess "space" "grade" coding conventions have really hit the codebase that hard DX

ohAitch:
I stand corrected, the behn is only 1/3 comments by weight! Like I guess you need to elaborate the above to the actual arvo interfaces, that's not like, more than a couple lines per arm though 🤔

(also %wegh and diagnostic scry and stuff, still doesn't super add up)

ohAitch:
As an exercise in readability reduction:
```
::
::::  /sys/vane/behn
  ::
  ::  %timer vane
::
|_  [current-timers=(list [set-to=@da for=duct]) unix=duct event=duct now=time]
++  wait
  |=  until=@da  ^-  [(list move) _current-timers]
  =<  :_  current-timers
      ::  fire next timer
      ::
      [unix %doze set-to:(snag 0 current-timers)]~
  |-  ^+  current-timers
  ::  base case: there are no further timers
  ::
  ?~  current-timers
    ::  insert at end of list
    ::
    [until event]~
  ::  terminal case: all other timers are further in the future
  ::
  ?:  (lte until set-to.i.current-timers)
    ::  insert at current position
    ::
    [[until event] current-timers]~
  ::  recursion case: advance past timer earlier than the one being inserted
  ::
  [i.current-timers $(current-timers t.current-timers)]
::
++  rest
  |=  which=@da  ^-  [(list move) _current-timers]
  =<  :_  current-timers
      ::  fire next timer, or clear if none are left
      ::
      [unix %doze ?~(current-timers ~ `set-to.i.current-timers)]~
  ^=  current-timers=_current-timers
  ::  remove the timer whose removal is being requested
  ::
  %+  skip  current-timers
  |=(timer=[@da duct] =(timer [which event]))
::
++  wake
  |=  ~
  =^  wake-subscribers  current-timers
    ::  separate out subscribers in past from timers that remain in the future
    ::
    |-  ^-  [wake-subscribers=(list duct) current-timers=_current-timers]
    ::  base case: no timers to wake
    ::
    ?~  current-timers
      [wake-subscribers=~ current-timers=~]
    ::  terminal case: hit first future timer
    ::
    ?:  (gth set-to.i.current-timers now)
      [wake-subscribers=~ current-timers=current-timers]
    ::  recursion case: get remaining past timers, and add current to that list
    =/  [wake-subscribers=(list duct) rest-timers=_current-timers]
      $(current-timers t.current-timers)]
    :-  wake-subscribers=:_(wake-subscribers for.i.current-timers)
    rest-timers
  ::  set remaining timer, and wake desired wakes
  ::
  :-  :-  ::  fire next timer, or clear if none are left
          ::
          [unix %doze ?~(current-timers ~ `set-to.i.current-timers)]
      ::  notify subscribers
      ::
      (turn wake-subscribers |=(a=duct [a %wake ~]))
  ::  update timer list
  ::
  current-timers
--
```

(I guess I probably should also have used skim and skip in `++wake`)

ohAitch:
(oh sorry missed the
```
::  $current-timers: (list [@da duct]), list of current timer ducts and when they should fire
::  $unix: duct, duct for sending timers to unix
::  $event: duct, duct for the current event
::  $now: time, time right now
::  $until: time, how long should this timer +wait
::  $which: time, which timer for this duct should be laid to +rest
::  $wake-subscribers: (list duct), which subscribers we are deciding to +wake
```
)

cgyarvin:
Hell needs to watch out for snowballs!

PR #1070, opened 2019-02-14
Clear ford, send %incompletes on %vega

pilfer-pandex:
Currently, when a kernel reset occurs, ford's state will find itself with copies of contents of the old kernel. To address this, on a %vega event, we purge all builds, and send %incomplete %made responses to all ducts who might expect another build result from us in the future (i.e. in-progress once builds and all live builds). Further commits will be added to this pull request to ensure that recipients perform the correct action (re-requesting the builds) on receiving such messages.

@belisarius222, could you please take a look at the `TODO: understand this` in my implementation of `++reset`? This point confused both Elliot and me. Thanks!

### DIFF
--- a/sys/vane/ford.hoon
+++ b/sys/vane/ford.hoon
@@ -77,7 +77,7 @@
 ::
 ::    Ford uses dependency tracking, caching, and results of previous builds
 ::    to eliminate excess work. When rerunning a live build, Ford "promotes"
-::    previous results to the new time if the build's dependencies hvaen't
+::    previous results to the new time if the build's dependencies haven't
 ::    changed since the previous build's formal date. Ford does this check
 ::    for each build in a tree of sub-builds under the "root build," which
 ::    is the build that was requested directly.
@@ -1093,6 +1093,7 @@
 ::      +cancel       stop trying to run a build and delete its tracking info
 ::      +wipe         wipe the build storage to free memory
 ::      +keep         resize caches, deleting entries if necessary
+::      +reset        cancel all live and in-progress builds; notify clients
 ::
 ::    The main internal arm is +execute-loop, which is called from +start-build,
 ::    +rebuild, and +unblock. +execute defines Ford's build loop.
@@ -1448,6 +1449,28 @@
               compiler-cache.state
       compiler-cache-size
     ==
+  ::  +reset: purge all builds and send %incomplete messages
+  ::
+  ::    The contents of Ford's build tables indirectly references the standard
+  ::    library and other core code. When a kernel reset occurs, these copies
+  ::    go out of date, making all the entries invalid. To handle this, we
+  ::    delete everything and send a message to everyone who is expecting a
+  ::    build result. These clients are expected to reiterate their original
+  ::    requests on hearing such %incomplete messages, allowing us to
+  ::    subsequently build our state anew.
+  ::
+  ++  reset  ^+  [moves state]
+    =<  finalize
+    =/  incompletes
+      %+  turn  ~(tap by ducts.state)
+      |=  [=^duct *]
+      `move`[duct %give %made now %incomplete ~]
+    ::
+    =.  moves  (weld incompletes moves)
+    :: For some reason, setting it to *_state instead of *ford-state
+    :: doesn't work. TODO: understand this.
+    =.  state  *ford-state
+    ..execute
   ::  +cancel: cancel a build
   ::
   ::    When called on a live build, removes all tracking related to the live
@@ -6185,11 +6208,11 @@
   ::
       ::  %vega: learn of kernel upgrade
       ::
-      ::    XX clear cache, rebuild live builds
-      ::
       %vega
     ::
-    [~ ford-gate]
+    =^  moves  state.ax  reset:this-event
+    ::
+    [moves ford-gate]
   ::
       ::  %wipe: wipe stored builds, clearing :percent-to-remove of the entries
       ::
--- a/sys/vane/ford.hoon
+++ b/sys/vane/ford.hoon
@@ -1467,8 +1467,6 @@
       `move`[duct %give %made now %incomplete ~]
     ::
     =.  moves  (weld incompletes moves)
-    :: For some reason, setting it to *_state instead of *ford-state
-    :: doesn't work. TODO: understand this.
     =.  state  *ford-state
     ..execute
   ::  +cancel: cancel a build
--- a/tests/sys/vane/ford.hoon
+++ b/tests/sys/vane/ford.hoon
@@ -7012,7 +7012,7 @@
       ::
       moves=~
     ==
-  (weld results (expect-ford-empty ford-gate ~nul))
+  (weld results (expect-ford-very-empty ford-gate ~nul))
 ::
 ++  test-vega-complete-once-build  ^-  tang
   ::
@@ -7044,7 +7044,7 @@
   ;:  weld
     results1
     results2
-    (expect-ford-empty ford-gate ~nul)
+    (expect-ford-very-empty ford-gate ~nul)
   ==
 ::
 ++  test-vega-live-build  ^-  tang
@@ -7077,7 +7077,7 @@
   ;:  weld
     results1
     results2
-    (expect-ford-empty ford-gate ~nul)
+    (expect-ford-very-empty ford-gate ~nul)
   ==
 ::
 ++  test-vega-many  ^-  tang
@@ -7148,7 +7148,7 @@
     results2
     results3
     results4
-    (expect-ford-empty ford-gate ~nul)
+    (expect-ford-very-empty ford-gate ~nul)
   ==
 ::
 ++  test-vega-in-progress-once-build  ^-  tang
@@ -7186,7 +7186,7 @@
   ;:  weld
     results1
     results2
-    (expect-ford-empty ford-gate ~nul)
+    (expect-ford-very-empty ford-gate ~nul)
   ==
 ::  |data: shared data between cases
 ::  +|  data
@@ -7536,12 +7536,13 @@
 ::  +expect-ford-empty: assert that ford's state is one empty ship
 ::
 ::    At the end of every test, we want to assert that we have cleaned up all
-::    state.
+::    state, modulo caching.
 ::
 ++  expect-ford-empty
   |=  [ford-gate=_ford-gate ship=@p]
   ^-  tang
   ::
+  :: clear all caches
   =^  results1  ford-gate
     %-  ford-call  :*
       ford-gate
@@ -7565,6 +7566,35 @@
   ?:  =(default-state state)
     ~
   ::
+  =/  build-state=(list tank)
+    %-  zing
+    %+  turn  ~(tap by builds.state)
+    |=  [build=build:ford build-status=build-status:ford]
+    :~  [%leaf (build-to-tape:ford build)]
+        [%leaf "requesters: {<requesters.build-status>}"]
+        [%leaf "clients: {<~(tap in ~(key by clients.build-status))>}"]
+    ==
+  ::
+  =/  braces  [[' ' ' ' ~] ['{' ~] ['}' ~]]
+  ::
+  :~  [%leaf "failed to cleanup"]
+      [%leaf "builds.state:"]
+      [%rose braces build-state]
+  ==
+:: +expect-ford-very-empty: assert that ford's state is a very empty ship
+::
+::   At the end of some tests, we want to assert that we have cleaned up all
+::   state, *including* the caches.
+++  expect-ford-very-empty
+  |=  [ford-gate=_ford-gate ship=@p]
+  ^-  tang
+  ::
+  =/  ford  *ford-gate
+  =/  state  state.ax.+>+<.ford
+  =/  default-state  *ford-state:ford
+  ::
+  ?:  =(default-state state)
+    ~
   =/  build-state=(list tank)
     %-  zing
     %+  turn  ~(tap by builds.state)
--- a/tests/sys/vane/ford.hoon
+++ b/tests/sys/vane/ford.hoon
@@ -14,6 +14,7 @@
 ::
 =/  test-pit=vase  !>(..zuse)
 =/  ford-gate  (ford-vane test-pit)
+=/  ford-types  *ford-gate  :: use only for getting the defined types, please
 ~!  +6.ford-gate
 ::
 |%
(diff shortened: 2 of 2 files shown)


sys/vane/ford.hoon
```diff
@@ -1448,6 +1449,28 @@
...
+  ::
+  ++  reset  ^+  [moves state]
+    =<  finalize
+    =/  incompletes
+      %+  turn  ~(tap by ducts.state)
+      |=  [=^duct *]
+      `move`[duct %give %made now %incomplete ~]
+    ::
+    =.  moves  (weld incompletes moves)
+    :: For some reason, setting it to *_state instead of *ford-state
```
belisarius222: That's because `_some-value` sets the default to `some-value`. This is how we can do
```
++  mul
  |=  [a=_1 b=_1]
  ...
```

tests/sys/vane/ford.hoon
```diff
@@ -6999,6 +6999,195 @@
     results4
     (expect-ford-empty ford-gate ~nul)
   ==
+::
+++  test-vega-empty  ^-  tang
+  ::
+  =^  results  ford-gate
+    %-  ford-call  :*
+      ford-gate
+      now=~1234.5.8
```
belisarius222: I think this should be changed to `~1234.5.6` for consistency with other tests.

tests/sys/vane/ford.hoon
```diff
@@ -6999,6 +6999,195 @@
...
+      scry=scry-is-forbidden
+      ::
+      call-args=[duct=~[/wherever] type=~ %vega ~]
+      ::
+      moves=~
+    ==
+  (weld results (expect-ford-empty ford-gate ~nul))
```
belisarius222: `+expect-ford-empty` isn't quite good enough as a test case here, because it clears Ford's caches: https://github.com/urbit/arvo/pull/1070/files#diff-bae0fc44604de49eb6a2fb9a968d81a9L7352

Probably the thing to do is make a new similar arm that doesn't send the `%keep` move.

belisarius222 (changes requested):
I requested a couple minor changes to the tests, but this looks good other than that.

sys/vane/ford.hoon
pilfer-pandex: Huh. Good to know.

tests/sys/vane/ford.hoon
```diff
@@ -7536,12 +7536,13 @@
...
 ::
 ::    At the end of every test, we want to assert that we have cleaned up all
-::    state.
+::    state, modulo caching.
 ::
 ++  expect-ford-empty
   |=  [ford-gate=_ford-gate ship=@p]
   ^-  tang
   ::
+  :: clear all caches
```
belisarius222: need a blank comment line below this comment

tests/sys/vane/ford.hoon
```diff
@@ -7580,4 +7581,33 @@
       [%leaf "builds.state:"]
       [%rose braces build-state]
   ==
+:: +expect-ford-very-empty: assert that ford's state is a very empty ship
+::
+::   At the end of some tests, we want to assert that we have cleaned up all
+::   state, *including* the caches.
```
belisarius222: also blank line here

tests/sys/vane/ford.hoon
```diff
@@ -7580,4 +7581,33 @@
...
+++  expect-ford-very-empty
+  |=  [ford-gate=_ford-gate ship=@p]
+  ^-  tang
+  ::
+  =/  ford  *ford-gate
+  =/  state  state.ax.+>+<.ford
+  =/  default-state  *ford-state:ford
+  ::
+  ?:  =(default-state state)
```
belisarius222: Instead of duplicating this code in `+expect-ford-very-empty` and `+expect-ford-empty`, I think the latter should call the former.

belisarius222 (approved):
LGTM, although I imagine we shouldn't merge this until Ford's clients retry on %incomplete. Otherwise I assume a |reset will sink a ship, which would be bad.

ohAitch:
> until Ford's clients retry on %incomplete

Are there any existing cases where ford will nondeterministically*
%incomplete a %once? Making all clients responsible for remembering copies
of all outstanding requests adds a lot of complexity to practical usage of
ford; I'm assuming there are architectural reasons ford can't do the
retrying itself?

* over clay, and the request noun itself

belisarius222:
It's possible for a client to ask Ford for a once build that won't complete, such as a build with a %scry sub-build that blocks and Ford can't translate to an async read move. We don't have any instances of those as far as I'm aware.

Ford builds can have vases in them, including vases of gates that contain the kernel. For those builds, I don't think there's a way Ford can update them to the new kernel and lose all references to the old, so we'd have to send an %incomplete.

But we could special-case builds that don't have vases in them. This includes the live builds Gall maintains on apps, which might be the only live builds in the system these days. Those builds just look like `[%core /path/to/app]`, so as long as we jettison the sub-builds and results, many of which contain the old kernel, we should be able to hold onto the root builds themselves and rebuild them without sending an %incomplete.

It makes Ford's internal logic a little more complicated, but I think it's still relatively straightforward: any root build whose schematic doesn't contain a vase should be held onto; all others need to have %incomplete's sent. We delete all build results, caches, and sub-builds. Then we rebuild all of our builds from scratch.

@pilfer-pandex Sorry, this is a bit more complex than I thought it would be.

belisarius222:
Not all clients need to retry. Dojo, for example, might just report the %incomplete to the user and let them decide if they care to rerun it. Eyre should probably also treat an %incomplete as a failure.

joemfb:
The only ford schematics with vases are `%slit` and `%$` literals, right? Any build not containing those should be straightforwardly rebuild-able?

I think sending `%incomplete` on all live builds during `|reset` punts too much to ford's clients. How should they track whether to rebuild incomplete builds? Will they need to retry all incomplete builds, but tracking retry counts so they can give up after some heuristic threshold?

belisarius222:
Yes, we should be able to rebuild any root build with no vases.

@pilfer-pandex and I have been talking about making the %incomplete contain a tagged union of `%vega`, `%scry-multiple-live-desks`, and `%scry-blocked`. A `%vega` should basically always be rerun, `%scry-blocked` might be totally dead or retryable depending on why the scry blocked (this one is a good candidate for retrying heuristically), and `%scry-multiple-live-desks` is totally dead.

If a Ford client gets a `%vega`, it knows the build could be rerun. I think in a lot of cases, the client doesn't need to hold onto the build itself, just whatever ingredients it had that let it create the build in the first place.

ohAitch:
I mean, this sounds like we need to actually implement a non-nop ++slim,
which can deal with cores and stuff. In particular, this same problem comes
up with requests dojo /doesn't/ send: presumably it shouldn't just delete
all variables on every(automatic, unprompted) reboot? (Which seems even
worse than the described "we said urbit is a server you can trust,
unfortunately it had to reboot, so all your open http requests and pending
repl calculations are dead now", but otherwise you're just punting the
stale kernel |mass to userspace)

Don't suppose there's any way to revision ford cache by reboot "eras", and
increasingly nag the user to kill requests manually if an era sticks around
for more than a day or two?

ohAitch:
(Also, by "nondeterministically" I didn't mean "ask Ford for a once build
that won't complete" so much as "ask Ford for a once build that does
complete, then ask again for same build and it doesn't complete" - but I
think the answer is still yes, bc gall/ames "scry" does the date-check
thing.
i.e. while [(scry clay/future) (scry ames/now)] will always fail, [(scry
ames/now) (scry clay/future)] will occasionally succeed if you have a fresh
enough now)

pilfer-pandex:
Based on this discussion, I am inclined to think that putting the retry logic in the clients would be unfortunate, perhaps even very unfortunate. But I don't know if it can be avoided in the current system design.

It seems to me that the reason we can't have ford rebuild them itself is that the schematics can contain old kernel state in them, so we don't want to hold onto them, and want to punt to the client to do whatever complex logic it has to do to recreate them on top of the new kernel state (how do they even do this?)

Tentatively, I don't like ted's proposal to hold onto all roots that can't contain kernel data because it complicates the API. The reason to have ford do the rebuilding itself is to remove concerns from clients and their authors. To me, if "you don't have to worry about X" is 0 points of concerningness, and "you have to worry about X" is 10 concerning, then "you usually don't have to worry about X, except in special case Y" is not 2 concerning, but rather 8 concerning, and sometimes even 15 concerning.

As it stands, it seems the schematic type is rather complicated anyway. I don't know if it's possible, but if we could eliminate the cases that store gates, perhaps we could simplify. (This might result in some loss of functionality, like taking us from support of monadic builds, to merely applicative builds. But until I know how to do a kernel update underneath a core, which I may never know how to do, it's unclear to me that supporting a fully monadic build language is semantically reasonable.)

pilfer-pandex:
Anton, if we could get dojo state to survive reloads, that would be spectacular. My main problem with all the existing repl technology is that when you reload a module you lose (or silently fail to properly update) all your existing bindings, making the edit, save, load-in-repl, test-in-repl development workflow very clunky.

ohAitch:
For an example of a dojo request the user can't "just retry":

    */=archives=/keep/it/safe +~net/strip-ads-and-stuff +
https://im-about-to.go-down.lol

Suppose ~net is physically on mars, and thus has 20 minutes of ping.
Suppose further that the possessor of the dojo is a grandchild of zod, and
zod chooses sometime within those 20 minutes to add a rune to hoon.hoon -
to keep you safe, citizen!
Now there is definitely a technical sense in which the site contents will
be preserved in the event log, and ready to recover if you really want to
go spelunking; hopefully, "spelunking in the event log" is a sign of
"serious urbit-level error condition", which I believe the semantics of
this PR to be.

(For the eyre equivalent, imagine some Arctic
thing-that-should-not-be-connected-to-the-Internet POSTing megabytes of
atmospheric data over a sat-phone; it can't just buffer that indefinitely,
and once again recovering the request urbit-side will involve a log spelunk)

ohAitch:
Wait does a |reboot in the *currently deployed* system nuke all dojo state?
D:

ohAitch:
Oh no I see what you mean by "fail to properly update", any |= you've saved
would have to be recursively rerun. Maybe worthwhile to save all variables
as schematics in case of library update, hm

pilfer-pandex:
Right. For us to do it properly, we'd have to regard all bindings as like entries in a spreadsheet or something. This might actually be worth thinking about. If we did it fully throughout, we'd solve the problem of kernel updates, code changes, and so on once and for all, and rather elegantly at that.

But:
- this might be hard
- it would take time
- it would radically change the system
- and I don't know how to make types work with it yet

cgyarvin:
I continue to believe that ford should only build from source — no vases.

ohAitch:
Regardless of how much of this is in ford, how should dojo handle the
following input?

    /+  somelib
    =foo  ~&(%foo =/(a (slow-thing:somelib 12) |=(b=@ (~(compute somelib a)
b))))
    =bar  ~&(%bar =/(a (slow-thing:somelib 13) |=(b=@ (~(compute somelib a)
(foo b)))))
    |reset
    (bar 14)

(Ideally without leaking old kernels)

cgyarvin:
Yeah that’s basically what I was suggesting

ohAitch:
To be concrete, the suggested output here is "-find-limb.bar"?

This doesn't seem super compatible with the "a computer you don't have to
reboot" value proposition: obviously in a system like that automated kernel
updates are impractical, and everyone is y'know /using/ their dojo
sessions, so isn't going to manually allow updates either. I guess you can
just write all your software against a 5-10 year old zuse+ford and hope for
the best?

joemfb:
The broader discussion here is important, but seems to be going farther afield... %gall's handling of reset notifications (and, by extension, dojo's) should be discussed separately. I'm not sure where it should take place to be honest, but just opening a separate issue would be better IMO.

I think that it's fine for %ford, upon receiving a kernel upgrade notification (`%vega`) to a) delete any stored build artifacts for once-builds, b) clear it's cache, and c) rebuild any live builds that it can. I don't think it's fine for %ford to send %incomplete build responses.

A quick grep suggests the `%slit` schematic is unused. If so, can it be removed? And do we know if any live builds are are using the `%$` schematic literal? If they are not, can we just disallow `%$` in live builds? That would remove the need for any reset-triggered `%incomplete` responses for live builds.

Also, does %ford explicitly track dependencies on the kernel? If so, can we use that tracking to trigger rebuilds of live builds?

cgyarvin:
This is all true except for the part about types

pilfer-pandex:
Anton, the idea is that

```
/+ somelib
=foo ~&(%foo =/(a (slow-thing:somelib 12) |=(b=@ (~(compute somelib a) b))))
=bar ~&(%bar =/(a (slow-thing:somelib 13) |=(b=@ (~(compute somelib a) (foo b)))))
|reset
(bar 14)
```

have the same semantics as

```
|reset
/+ somelib
=foo ~&(%foo =/(a (slow-thing:somelib 12) |=(b=@ (~(compute somelib a) b))))
=bar ~&(%bar =/(a (slow-thing:somelib 13) |=(b=@ (~(compute somelib a) (foo b)))))
(bar 14)
```

pilfer-pandex:
Also I should report the outcome of a live discussion with @belisarius222, @eglaysher, and @vvisigoth:

It turns out that there are only two places in the system where we use live builds:
- Gall
- "Security drivers" in eyre

The latter has a vase inside its schematic, which means the schematic contains a kernel copy.

Our current thinking is that the best solution is to:
- Ban schematic types that can contain vases in live builds, so that the schematics for live-builds are "reload-safe."
- on %vega:
    - wipe all state except for live root builds
    - send incompletes to in-progress once builds
    - inside ford, and transparently to clients, rerun all root live builds, as though they had been submitted afresh from clients.

We determined that the correct time to land these changes is after Elliot's new eyre, which, somehow, shall not use live builds in this problematic manner. Accordingly this pr is on hold.

joemfb:
This sounds reasonable.

I want to point out that all the vanes receive the `%vega` notification, so the responsibility to reload state that would otherwise hold onto the old kernel is and should be shared across the system. If there are situations where %ford can't be sure on it's own that it's safe to discard some state, it's possible that the client of the build can supply the answer. It may be that %ford should just hang onto state in that case, and wait for it to be cleared or superseded by a request from the client. I'm not sure if this changes the analysis here, but it's relevant to the larger goal of prevent space leaks.

ohAitch:
Insofar as we can agree that "I don't think it's fine for %ford to send
%incomplete build responses.", the details of how dojo(which doesn't send
%live builds) continues not to have to handle %incomplete is indeed out of
scope. Insofar as this PR is on track to be accepted, in its current
form("on %vega, wipe all state except for live root builds") where it
introduces a new obligation for dojo to handle %incomplete, the discussion
seems pretty in-scope.

(Specifically, it is my opinion that it shouldn't handle %incomplete by
deleting user data, and my understanding of the pro-%incomplete side is for
dojo to handle %incomplete by deleting user data)

ohAitch:
> situations where %ford can't be sure on it's own that it's safe to
discard some state

I would feel a lot better about this PR if instead of %incomplete, the once
builds would receive %please-cancel-this, with ford otherwise hanging on to
and continuing to compute the build. If the builder actually has vaseless
source data, they can cancel the original and start a new once build; if it
doesn't and the request is taking forever, you can prompt the user to
determine if the operation is reasonable to cancel. Something similar to
the appplicationDidReceiveMemoryWarning mechanism in iOS perhaps.

(If the concern here is the old kernel polluting the build cache, I repeat
the suggestion to version said cache by %vega number, clearing it when you
run out of straggling once builds of the relevant version)

PR #1074, opened 2019-02-17
pin to local time because using local desk

philipcmonk:
This worked once upon a time in another branch, and it's pretty clearly correct, but I haven't tested it directly against current hotfix.

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -1757,8 +1757,9 @@
     :*  hen  %pass
         [%foreign-x (scot %p our) (scot %p her) syd car (scot cas) pax]
         %f  %build  live=%.n  %pin
-        ::  XX perhaps should be now as in ++validate-plops
-        (case-to-date cas)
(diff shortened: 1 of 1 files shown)


sys/vane/clay.hoon
```diff
@@ -1757,6 +1757,7 @@
     :*  hen  %pass
         [%foreign-x (scot %p our) (scot %p her) syd car (scot cas) pax]
         %f  %build  live=%.n  %pin
+        ::  XX perhaps should be now as in ++validate-plops
         (case-to-date cas)
```
joemfb: It does seem like this should also be now. (This comment also needs a trailing empty line.)

joemfb:
I think this should be tested further before merging. I'll try some local fake-ship experiments. Should be easier to replicate now that the problem is clear (ie, wait a little while before booting a fake ~marzod).

joemfb (approved):
LGTM. And sorry for the delay

PR #1079, opened 2019-02-21
More accurately represent Ethereum RPC filter topics

Fang-:
[Relevant Ethereum JSON RPC docs section.](https://github.com/ethereum/wiki/wiki/JSON-RPC#a-note-on-specifying-topic-filters)

Rather than a flat list of topics to watch for, it is an indexed/ordered list of topics to watch for in those specific positions. As such, items in that list may either be:
- zero (null, matches any topic in that position)
- non-zero (matches one specific topic in that position)
- a list of topics (matches any one of those topics in that position)

Additionally, topics are always specified using 32 bytes, so there is no need to specify length explicitly.

### DIFF
--- a/sys/vane/jael.hoon
+++ b/sys/vane/jael.hoon
@@ -2121,7 +2121,7 @@
       ==
     =+  cuz=[block-number.place log-index.place]
     ::
-    ?:  =(event.log changed-dns:azimuth-events)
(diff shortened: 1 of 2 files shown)


Fang-:
image
god I love MS-era Github

@philipcmonk and I discussed the solution for this out of band. @philipcmonk, please confirm this is in line with that and looks sane.

PR #1081, opened 2019-02-22
Implement %eth-watcher, an app for tracking Ethereum events

Fang-:
This is more or less a port of jael's Ethereum watching logic, without the Azimuth-specific transformations and with some extra niceties to make this generically usable.

Poke with `[%watch %sometag config:eth-watcher]` to initialize,
then subscribe at `/sometag` to receive updates in the shape of
`[%snap snapshot:eth-watcher]` for initial and on-reorg logs,
`[%logs loglist]` for logs as they happen.

I haven't *thoroughly* tested the peer/subscription functionality yet, but an app that uses this to build Azimuth statistics is next on my to-do list. That'll both hit that, and tell me whether there's any functionality that is still missing from this.
So, ready for review, but might see a small addition or two.


app/eth-watcher.hoon
```diff
@@ -0,0 +1,556 @@
...
+  ?-  what
+      %debug
+    ~&  [%log-lent (lent logs.eye)]
+    ~&  [%last-heard last-heard-block.eye]
+    [~ +>.$]
+  ::
+      %debug-init
+    =-  done:(init:watcher name -)
+    ^-  config
+    :*  (need (de-purl:html 'http://104.198.35.227:8545'))  ::  parity
```
dpc: What are these IPs?

ohAitch:
Whatever they are, if they don't have existing DNS they should
get("external-foo."?) names under urbit.org

app/eth-watcher.hoon
Fang-: Tlon-hosted Ethereum nodes. @ohAitch good Q, not entirely sure if they have domains pointing to them. Not sure if we want to either. I should clean out/up most of the debug stuff before merge.

ohAitch: `.tlon.io` then :P

It seems pretty valuable for code readability; is the concern here making them harder to find and DoS? ~sometimes the person you end up making the system secure against… is yourself~

dpc: https://blog.ethereum.org/2015/08/29/security-alert-insecurely-configured-geth-can-make-funds-remotely-accessible/

> ADVISORY: Never enable JSON-RPC interface on an internet-accessible machine without a firewall policy in place to block the JSON-RPC port (default: 8545).

Maybe things changed between now and then, but I generally feel uneasy about hosting a publicly accessible JsonRPC port for anything. Somewhere down the line there will be some security vuln. available, and at very least someone will be able to eg. serve fake data to other users relying on this host.

ohAitch: I mean, they'll be able to serve _stale_ data, if a single ethereum node can serve you _false_ data about the blockchain we have bigger problems. Unless this app isn't checking signatures / timestamps / difficulty?

(though finding a self-hosted JSON-RPC wrapper that does claim to be safe for public consumption does seem like a good idea, having to reprovision the server in question sounds annoying at least)

dpc: @ohAitch 

Worst case scenario. There's is a 0-day bug in parity/geth RPC, someone gains shell access to this host. They totally replace the geth/parity with their own version, that servers *any* data they desire.


RPC ports are just lot of API surface. A lot of commands, with varying functionality - accessing filesystems etc., and since it's recommended to keep it for localhost consumption, it doesn't get as much scrutiny.

Any software that only consumes RPC data and does not directly connect with p2p network, is at a mercy of the node they are querying giving them correct data, as there is nothing to cross-match it with.

Even if this host is not compromised by an outsider, all users consuming this RPC, are trusting Tlon administrators.

Fang-: @dpc that advisory was and still is about cases where you have wallets accessible/created via the RPC interface. Having wallets with *anything* attached to them (funds, permissions, whatever) connected to the internet is obviously dangerous and stupid. Exposing an RPC node just for reading the blockchain is fine and not uncommon.

@philipcmonk care to confirm my sanity here?

philipcmonk: We use http://eth-mainnet.urbit.org:8545/ for preboot validation.

That advisory doesn't apply, as @Fang- says, however @dpc 's concern about needing to trust the node to give valid information is valid.  This is a problem with *any* application that isn't running a full node inside (including metamask, mycrypto, augur, etc).  They can't steal your money directly, but they can fool you into thinking the blockchain says something other than what it does.

We should of course switch to HTTPS -- that would reduce the MITM attack surface.  But if the server the node is on is compromised, then all bets are off.  Each user should decide for themselves what level of security do they want.  They can run their own node(s) and point this at them, or they could use Infura if they trust them (if we had a websocket client, or else getFilterChanges doesn't work).  For now, Tlon hosts a couple nodes, and I think that's fine for a default.  It wouldn't be crazy to find several sources and verify they're all giving the same information.  This is definitely worth revisiting some time.

Fang-: Note that the addresses of Tlon's nodes aren't in this app as defaults, but as hardcoded values for a hardcoded config that's used during debugging. In actual use, the user needs to decide which node address to pass in, which is where the considerations mentioned above come in.

app/eth-watcher.hoon
```diff
@@ -0,0 +1,556 @@
...
+  ::
+  ++  wait-poll
+    =+  wen=(add now ~m4)
+    %-  put-move(poll-timer.eye wen)
+    [%wait name^/poll wen]
+  ::
+  ::  +cancel-wait-poll: remove poll reminder
+  ::
+  ++  cancel-wait-poll
+    ?:  =(*@da poll-timer.eye)  ..cancel-wait-poll
```
philipcmonk: If something is nullable, it should be a `unit`.  Otherwise it's easy to assume it exists when it doesn't.

app/eth-watcher.hoon
ohAitch: I'm taking that as a yes on `this app isn't checking signatures / timestamps / difficulty`: we're speaking over RPC to Tlon software that claims to be tracking the Azimuth contract
Though if it's only doing so for debugging that does sound fine.

Back to the original "what are there IPs" question: any reason not to DNS them as `test-eth-parity.urbit.org` etc?

app/eth-watcher.hoon
```diff
@@ -0,0 +1,556 @@
...
+    =.  +>.$  (fan-diff snap+snap)
+    ::  restore state and kick new fetch cycle
+    ::
+    %=    get-latest-block
+        last-heard-block.eye  last-heard-block.snap
+        heard.eye         heard.snap
+        logs.eye          logs.snap
+    ==
+  --
+--
```
philipcmonk: need a trailing newline

sur/eth-watcher.hoon
```diff
@@ -0,0 +1,39 @@
...
+::
+++  snapshot
+  $:  last-heard-block=@ud
+      heard=(set event-id:ethereum)
+      logs=loglist
+  ==
+::
+++  loglist
+  (list event-log:rpc:ethereum)  ::  newest first
+--
```
philipcmonk: need a trailing newline.

lib/eth-watcher.hoon
```diff
@@ -0,0 +1,38 @@
...
+  ::  otherwise, keep looking through :old
+  ::
+  ?:  ?|  (gth block.new-place block.old-place)
+          ?&  =(block.new-place block.old-place)
+              (gth log.new-place log.old-place)
+          ==
+      ==
+    [i.new ^$(new t.new)]
+  [i.old $(old t.old)]
+--
```
philipcmonk: need a trailing newline

lib/eth-watcher.hoon
```diff
@@ -0,0 +1,38 @@
+::  eth-watcher utilities
```
philipcmonk: Why are these two in a separate lib?  Do we expect clients to use them?

app/eth-watcher.hoon
```diff
@@ -0,0 +1,556 @@
...
+  [~ ..prep(+<+ u.old)]
+::
+++  poke-noun
+  |=  [what=?(%debug %debug-init %save %load) =name]
+  ^-  (quip move _+>)
+  =+  eye=(fall (~(get by eyes) name) *eye)
+  ?-  what
+      %debug
+    ~&  [%log-lent (lent logs.eye)]
+    ~&  [%last-heard last-heard-block.eye]
```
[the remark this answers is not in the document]
Fang-: Ah, good call on ++scry, this really should allow scrying for logs under labels also. On it!

app/eth-watcher.hoon
philipcmonk: The parity one is DNSed to eth-mainnet.urbit.org.  But yeah, the user should be getting their own node from somewhere.

@ohAitch that's correct, we don't implement a light node in this app.  Even if we did, that still wouldn't guarantee canonicality unless we could trust its peers.  Even a full node is only as good as its peers.  If its peers disagree, then it can choose the right one by checking difficulty, but eg if it only has one peer it won't have the right info.  It needs to have at least one honest peer.

lib/eth-watcher.hoon
Fang-: I imagine clients may want `++store-new-logs`, yes. `++log-to-id` is just there to support it, though maybe it belongs in zuse instead?

app/eth-watcher.hoon
Fang-: "Need"? Is there something that would be tripped up by not having trailing newlines?

app/eth-watcher.hoon
```diff
@@ -0,0 +1,556 @@
...
+  ::
+  ::  +|  outward
+  ::
+  ++  wipe
+    =>  cancel-wait-poll
+    =>  cancel-subscribers
+    :-  (flop moves)
+    ..watcher(eyes (~(del by eyes) name))
+  ::
+  ++  done
```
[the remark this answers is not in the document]
Fang-: We don't nest cores here, and likely never will, so I think just "done" is fine. (Also haven't even seen that get confusing, nested cores usually separate fairly cleanly, at least in their open-close arms.)

app/eth-watcher.hoon
```diff
@@ -0,0 +1,556 @@
...
+++  sigh-json-rpc-response
+  |=  [wir=wire res=response:rpc:jstd]
+  ^-  (quip move _+>)
+  ?>  ?=([@ *] wir)
+  =<  done
+  %-  sigh-json-rpc-response:(open:watcher i.wir)
+  [t.wir res]
+::
+++  watcher
+  |_  $:  =name
```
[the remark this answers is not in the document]
Fang-: I've seen that before in places. I prefer `|_` in that it very clearly communicates that "these things are used/changed inside this core". I agree this does leave the sample there easily touchable, which may add some confusion back in...

app/eth-watcher.hoon
philipcmonk: Yeah, isn't that what sunk ~zod one of those times?  We might manage to parse it for gall apps, but at least hoon.hoon really needs a trailing newline or it won't compile.  Plus, a text file with no trailing newline is really malformed.

Issue #585, opened 2018-01-18
=* tistar should allow for optional type annotation


Fang-:
`=*` is a rune strictly used for code readability, and being able to add in a type annotation seems to be in line with that.

Should help with sanity checks for both the writer ("I expect this thing here to be of this type") and the reader ("what type is the data that `p.q.r.s.dat` refers to again?").

cgyarvin:
It definitely should.  It's one of those odds and ends I kept meaning to
get around to.  Actually a good little project for those who don't fear the
compiler.

mikolajpp:
Just signalling, ~palfun got me on this. We will see what happens.
Anyway, I always kind of liked compilers, especially to-nock ones ;-)

cgyarvin:
Good luck!  Probably not that hard :-)

joemfb:
@mikolajpp, are you working this one? I'm looking for a target, so I'll take it if you're not.

cgyarvin:
Lmk if you need help with compiler staging...

mikolajpp:
@joemfb feel free. Working, but not in an immediate sense, as in I am reading through hoon.hoon for the first time...

joemfb:
So, it appears I do need help with staging...

I've added a `%tstc` hoon to demonstrate what `%tstr` should become:
[arvo](https://github.com/urbit/arvo/compare/master...joemfb:tstr-toro) and
[vere](https://github.com/urbit/urbit/compare/master...joemfb:tstr-toro).

I can sync those changes, `|reset` or `+solid`, and everything works as expected:

```
> =`(a our a)
~zod
> =`(a=@ our a)
0
> =`(a=@ our =`(b=@p (mix (lsh 5 1 (end 5 1 eny)) a) [a b]))
[0 ~danweg-bosrym-dozzod-dozzod]
```

However, when I update the `%tstr` implementation to match `%tstc` in hoon.hoon
(and remove `%tstc`), I can no longer `|reset` or `+solid`. Specifically, I get
`%find` errors compiling zuse and the vanes -- the same errors regardless of
whether the changes are in vere, surprisingly enough (at least to me). I'm sure
we've discussed the staging procedure before (and I thought I had successfully
done it before), but I'm at a bit of a loss.

For the sake of clarity, here were my steps:

- boot a fresh fake-zod from master
- additive changes to vere
- additive changes to hoon.hoon, |reset, success
- mutative changes to hoon.hoon, no |reset, +solid fails
[...]

PR #1084, opened 2019-02-26
Tiebout - Apple Push Notification Server App

loganallenc:
Simple gall app for sending push notifications to an urbit user's iPhone when they receive a Hall message from a circle that has been subscribed to these notifications.


vvisigoth:
@loganallenc remember to request review (I just added @ixv and @Fang- )

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+  ^-  (quip move _this)
+  ?~  wir
+    (mean [leaf+"invalid wire for diff: {(spud wir)}"]~)
+  ?+  i.wir
+    (mean [leaf+"invalid wire for diff: {(spud wir)}"]~)
+      %our
+    ?<  ?=(~ t.wir)
+    :_  this
+    =/  doc/dock  [our.bol %hall]
+    [ost.bol %peer /our/[i.t.wir] doc /circle/[i.t.wir]/config/grams]~
```
Fang-: Maybe we want to re-subscribe from the last-heard message going forward? I think the logic below ensures we don't send notifications for whatever old junk we may get here, but still.

Pretty sure this just gives you the last `100` messages, or all messages from the last `~d1` or something along those lines. Being more specific allows you to catch up from even further back when needed, and doesn't get you more info than you need.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+      {%our @ @}
+    ?>  ?=(%circle -.rum)
+    =/  nom/name:hall  i.t.wir
+    ?+  -.rum.rum
+      [~ this]
+    ::
+    ::  %gram: send notification if envelope is lower than read number
+    ::
+      %gram
+    =/  red  (~(get by circles.tie.sta) nom)
```
Fang-: Indent this stuff (in this `?+`) like you did in earlier code. Current indentation is a tad hard to read.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+    ?~  red
+      (mean [leaf+"invalid circle for diff: {(spud wir)}"]~)
+    ?:  (gth num.nev.rum.rum u.red)
+      :_  this(circles.tie.sta (~(put by circles.tie.sta) nom u.red))
+      (message-to-notification u.red nev.rum.rum)
+    :_  this
+    (message-to-notification u.red nev.rum.rum)
```
Fang-: Wait, huh. The comment says "send notification if ... lower than read number", but it seems to send a notification either way, only updating state in one of the cases? This seems wrong.

(And if it's right, at least do `:-  (msg-to-not ...)  ?:  ...`.)

(There's also questions here about what we want to happen in the "message was edited case", but we probably want to do nothing? Editing isn't really supported anywhere anyway.)

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+    ::
+    ::   %config: set our read number
+    ::
+      %config
+    ?+  -.dif.rum.rum
+      [~ this]
+          %read
```
Fang-: Indents seem weird here. I think `%read` and onward is too deep? Also separate these with a `::` line for absolute clarity.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+        [~ this(circles.tie.sta (~(put by circles.tie.sta) nom red))]
+      ==
+    ==
+  ==
+::
+::  generate notification move from hall message
+::
+++  message-to-notification
+  |=  [red=@ud env=envelope:hall]
+  ^-  (list move)
```
Fang-: This seems to always produce a list of either zero moves, or one move. Sounds like a `(unit move)` to me.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+  ?:  =(aut.gam.env our.bol)
+    ~
+  =/  pay  %-  my  :~
+    alert+s+'New message from {(cite:title aut.gam.env)}'
+  ==
```
Fang-: This could do with a type for `pay`. `my` isn't commonly used, `~(gas in *(map @t json))` is more verbose/clear on what it does (and type-checks early!). For this case, you could also `~(put by *(map @t json))` since you only have one element.

Or just do the dumbest thing and say `[alert+s+'etc' ~ ~]` and cast that to a `map`. (^:

(The most aggressive thing to do here, of course, is to tell @pilfer-pandex to put "map/set syntax" on the global hoon wishlist.)

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+::  +create-apns-request: create hiss with payload for APNs
+::
+++  create-apns-request
+  |=  not=notification
+  ^-  hiss
+  =/  furl=@t  (crip (weld (trip baseurl.tie.sta) (trip token.not)))
+  =/  url=purl  (need (de-purl:html furl))
+  =/  jon=json  :-  %o  %-  my  :~
+    aps+o+payload.not
+  ==
```
Fang-: [o, my payload](https://i0.kym-cdn.com/entries/icons/original/000/023/390/O.png)

Same map building advice from above applies here (and below). But also, there's stdlib functions dedicated to construction json. Look at `++en-json`.

(Also chaining tall-form together on the same line like this is generally bad style.)

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+      (send-notify +.act)
+  ==
+::
+::  +add-circle: add circle and subscribe for updates
+::
+++  add-circle
+  |=  nom=name:hall
+  ^-  (quip move _this)
+  :_  this(circles.tie.sta (~(put by circles.tie.sta) nom 0))
+  [ost.bol %peer /our/[nom] [our.bol %hall] /circle/[nom]/config/grams]~
```
Fang-: There's a lot of move construction happening in separate arms. There's common themes, like `ost.bol` at the head, or `[our.bol %hall]` being used for the `dock`. You may or may not want to write arms that deal with the common parts of move construction for you.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+    alert+s+'New message from {(cite:title aut.gam.env)}'
+  ==
+  =/  not/notification  :+
+    token.tie.sta
+    'com.tlon.urbit-client'
+    pay
+  ?:  (lte num.env red)
+    ~
+  =/  doc/dock  [king.tie.sta dap.bol]
+  [ost.bol %poke /ask-king doc %tiebout-action [%notify not]]~
```
Fang-: Ah, so kings route your notifications for you? Is this because we don't want any old ship to access our push endpoint? idk about loading it all onto ~dopzod, we may want a dedicated ship for this, but the load probably isn't *that* big?

mar/tiebout-action.hoon
```diff
@@ -0,0 +1,40 @@
...
+    ++  action
+      %-  of:dejs
+        :~  token+so:dejs
+            add-circle+sa
+            del-circle+sa
+        ==
+    ++  sa            :: string as ta
+      |=  jon=^json
+      ?>  ?=([%s *] jon)
+      (scot %tas p.jon)
```
Fang-: If you want a `@tas` from the `@t` `p.jon`, you can either just cast it, or `?>  (sane %tas p.jon)` and then cast, or similar. `(scot %tas` just gives you the "rendered" version, which may not be what you want.

mar/tiebout-action.hoon
```diff
@@ -0,0 +1,40 @@
...
+++  grab
+  |%
+  ++  noun  action:tiebout
+  ++  json
+    |=  jon=^json
+    ^-  action:tiebout
+    %-  (hard action:tiebout)
+    =<  (action jon)
+    |%
+    ++  action
```
Fang-: `|^` exists and might be nicer here.

mar/tiebout-action.hoon
```diff
@@ -0,0 +1,40 @@
...
+  ++  tank  >act<
+  --
+::
+++  grab
+  |%
+  ++  noun  action:tiebout
+  ++  json
+    |=  jon=^json
+    ^-  action:tiebout
+    %-  (hard action:tiebout)
```
Fang-: Do you *need* to hard this? If you write your json parsing correctly it should just nest.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
+::
+::::  /app/tiebout/hoon
+::
+/?  309
```
Fang-: Woah, I haven't seen one of these used in forever. I don't actually know how this operates across breaches that pass by this version etc. (Not that I've ever seen zuse version be changed in the first place.)

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+/-  hall
+/+  tiebout
+=,  tiebout
```
Fang-: I learned recently that you can do `/+  *tiebout` to not put a face on it, so you don't need to `=,` afterwards.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+=,  eyre
+::
+::  state:
+::
+~%  %tiebout  ..^is  ~
```
Fang-: Why is this here (and in the `/lib`)? Is this gonna be jetted or something? I haven't ever seen this in apps before.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+      (set-token +.act)
+    $baseurl
+      (set-baseurl +.act)
+    $add-circle
+      (add-circle +.act)
+    $del-circle
+      (del-circle +.act)
+    $notify
+      (send-notify +.act)
+  ==
```
Fang-: Old syntax!

The indentation is backwards here, I'd just single-line each of these cases anyway. They're short enough.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+/?  309
+/-  hall
+/+  tiebout
+=,  tiebout
+=,  eyre
+::
+::  state:
+::
+~%  %tiebout  ..^is  ~
+|_  [bol=bowl:gall sta=state]
```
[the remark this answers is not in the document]
ixv: I disagree with this, having a face for state means makes things clearer. I've heard other people argue both sides of this before. Perhaps this is something that should be resolved by a style guide. As someone who has had to update someone else's =* lark expression, I'm against that solution.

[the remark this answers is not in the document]
Fang-: Agreed that `=*`ing some lark is bad. I've felt your pain before. (But never in this context, because the lark here practically never changes.)

ime this is a non-issue for top-level app cores though. Have never ever needed a face on app state. Clarity is resolved by "common pattern" and having the state structure in the same file.

Surprised to hear others have argued for adding a face in this context, would be curious to hear from them.

lib/tiebout.hoon
```diff
@@ -0,0 +1,58 @@
+::
+::::  /lib/tiebout/hoon
+  ::
+/?  309
+/-  hall
+=,  eyre
+::
+~%  %tiebout-lib  ..is  ~
```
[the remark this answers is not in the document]
loganallenc: no, removing

lib/tiebout.hoon
```diff
@@ -0,0 +1,58 @@
...
+  $%  [%hall-rumor rumor:hall]
+      [%tiebout-action action]
+  ==
+::
++$  poke
+  $%  [%tiebout-action action]
+  ==
+::
++$  state
+  $%  [%0 tie=tiebout-zero]
```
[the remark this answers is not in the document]
loganallenc: good idea

app/tiebout.hoon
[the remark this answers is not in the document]
loganallenc: Keeping the face.

app/tiebout.hoon
loganallenc: removing

app/tiebout.hoon
loganallenc: cool!

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+    ::  XX send message to users inbox
+    [~ this]
+  ?~  wir
+    (mean [leaf+"invalid wire for diff: {(spud wir)}"]~)
+  ?+  i.wir
+    (mean [leaf+"invalid wire for diff: {(spud wir)}"]~)
+      %our
+    ?<  ?=(~ t.wir)
+    :_  this
+    ~
```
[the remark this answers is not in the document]
loganallenc: yup

app/tiebout.hoon
loganallenc: fixed

app/tiebout.hoon
loganallenc: done

lib/tiebout.hoon
```diff
@@ -0,0 +1,58 @@
+::
+::::  /lib/tiebout/hoon
+  ::
+/?  309
+/-  hall
+=,  eyre
+::
+~%  %tiebout-lib  ..is  ~
+|%
++$  move  [bone card]
```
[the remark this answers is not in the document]
loganallenc: Moved. Thanks!

app/tiebout.hoon
pilfer-pandex: Yes, this is certainly on my radar.

Fang-: Also wait a second, `'cords'` don't support string interpolation! You need to `(crip "tape {etc}")` here.

app/tiebout.hoon
loganallenc: Yes, msg-to-not contains the conditional that checks whether to send a notification. Renamed for clarity.

app/tiebout.hoon
loganallenc: I like the code style of sending back a list much more. I'd need to add another layer of conditional checks in diff-hall-rumor otherwise.

app/tiebout.hoon
loganallenc: I don't like gas in map. It's less clear to me than my.

app/tiebout.hoon
loganallenc: Will fix tall-form chaining.

mar/tiebout-action.hoon
loganallenc: Nest-failed in a weird way, skipping it.

app/tiebout.hoon
Fang-: Yeah, I guess `(list move)` is still fair game here. The contract between "notification" and "moves" is a bit weird, so maybe comment, but approved either way.

mar/tiebout-action.hoon
Fang-: Happy to help you figure it out if you want to, but I won't press you on it.

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+  [~ this]
+::
+::  +diff-hall-prize: receive new circle data
+::
+++  diff-hall-prize
+  |=  [wir=wire piz=prize:hall]
+  ^-  (quip move _this)
+  ~&  prize+[wir piz]
+  ?~  wir
+    (mean [leaf+"invalid wire for diff: {(spud wir)}"]~)
```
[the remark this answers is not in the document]
loganallenc: fixed

app/tiebout.hoon
```diff
@@ -0,0 +1,253 @@
...
+  ?+  wir
+    (mean [leaf+"invalid wire for diff: {(spud wir)}"]~)
+  ::
+  ::  %our: set config of circle and iterate through messages, sending
+  ::  notifications for all messages where number is higher than our last-read
+  ::
+      {%our @ @}
+    ?>  ?=(%circle -.piz)
+    =/  nom/name:hall  i.t.wir
+    =/  red/@ud  red.loc.cos.piz
```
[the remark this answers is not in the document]
loganallenc: Can you include type information in `=*`?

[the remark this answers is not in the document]
Fang-: Yes! Since #585 you can do `=*  foo=bar  baz`. I actually forget this made it in already, and should've been using it more. Neat!

app/tiebout.hoon
```diff
@@ -0,0 +1,236 @@
...
+::  +reap: recieve acknowledgement for peer
+::
+++  reap
+  |=  [wir=wire err=(unit tang)]
+  ^-  (quip move _this)
+  ?~  err
+    [~ this]
+  ?~  wir
+    (mean [leaf+"invalid wire for diff: {(spud wir)}"]~)
+  ?+  i.wir
```
Fang-: I thought I left a comment about this in the initial review but can't find it, so maybe it got eaten? You already do a saner `?+` in the `++diff-hall-*` cases, you'll want to apply that here and in `++quit` too.

(Also indenting for this particular one is borked.)

app/tiebout.hoon
```diff
@@ -0,0 +1,236 @@
...
+  ^-  (list move)
+  ?:  =(aut.gam.env our.bol)
+    ~
+  =/  pay  %-  my  :~
+    alert+s+'New message from {(cite:title aut.gam.env)}'
+  ==
+  =/  not/notification  :+
+    token.sta
+    'com.tlon.urbit-client'
+    pay
```
Fang-: You're probably fine putting these between `[]` on one line. If you feel that's too long, proper `:+` form would be

```hoon
:+  token.sta
  'com.tlon.urbit-client'
pay
```

sur/tiebout.hoon
```diff
@@ -0,0 +1,62 @@
+::
+::::  /lib/tiebout/hoon
```
Fang-: Pretty sure filepaths in file headers is outdated style.

And as mentioned before, lots of the structures (cards, state) here really belong in the `/app` file instead, since they aren't shared to/used by other applications/libraries interfacing with Tiebout.

loganallenc:
Hey man! I removed the lib from the repository, made the indentation change and removed the `?~  i.wir` you've mentioned. I also added the change to the +:  for the notification that you mentioned. This should be ready to go now. Thanks!

app/tiebout.hoon
```diff
@@ -0,0 +1,269 @@
...
+::
+++  diff-hall-rumor
+  |=  [wir=wire rum=rumor:hall]
+  ^-  (quip move _this)
+  ?+  wir
+    (mean [leaf+"invalid wire for diff: {(spud wir)}"]~)
+  ::
+  ::  %our
+  ::
+      {%our @ @}
```
Fang-: While these are technically correct and will successfully match `/our/whatever` aka `[%our %whatever ~]`, it's better to describe a limited-length wire by actually ending with `~` in the type description.

PR #1085, opened 2019-02-27
added initial image support to udon parser

jfranklin9000:
This patch implements `alt-text` markdown to udon.
The `alt-text` is displayed if `image-url` doesn't exist.

One small issue is `` produces an `alt=""` tag.
This is fine if `image-url` exists, but ``
shows the broken image icon in Firefox (desired), while Chrome
shows nothing (arguably correct).  I'd like to see the broken
image icon and will look into not adding the `alt` tag in this
case.

Another issue concerns `alt-text`.
This affects link markdown too: `[some-content](link-url "title-text")`.
`https://urbit.org/docs/learn/arvo/arvo-internals/udon/` does
not mention title text markdown so it doesn't exist in udon.
However, the parser will consume it and produce
`<a href="link-url &quot;title-text&quot;">some-content</a>`
instead of using a `title` tag.  I will look into proper title text
handling for links and images.

I don't think either of these issues should prevent a pull.


jfranklin9000:
The first issue can be fixed by replacing
```
            $mage  [[%img [%src q.nex] [%alt p.nex] ~] ~]~
```
with
```
            $mage  [[%img [%src q.nex] ?~(p.nex ~ [%alt p.nex]~)] ~]~
```
I'll commit, squash and push later today.

jfranklin9000:
I've pushed the no-alt-text fix.

rmariani:
This is great! Having to use Sail for images in otherwise-Udon files was annoying.

PR #1093, opened 2019-03-04
Update leap second lists

Fang-:
I found a good source for the "leapsecond days", so I added that in as a comment, for future reference.

I'm unsure where the "leapsecond dates" was sourced from, having a hard time finding a source that uses those exact times, so I just did the "second at end of day" thing that the previous two leap seconds did.

(The list under `+lef` is an argument in favor of #1065. (^: )


cgyarvin:
I think we should kill this feature and move to Google style time distortion

belisarius222:
I know almost nothing about this subject. @Fang- or @cgyarvin any idea who would be more qualified to review this PR?

Fang-:
@belisarius222 no idea. I just tagged you randomly, this change as presented only requires the ability to fact-check. For further discussion around the best approach to take (and how to do so!) we should definitely ask some time wizards.

Reference on (leap) time smearing [here](https://developers.google.com/time/smear).

Fang-:
(For the record I think we're fine merging this in as a quick fix, and then discussing the proper approach to leap seconds in a dedicated issue.)

ohAitch:
Don't feel strongly about this, but I thought it was urbit's considered
position not to respect leap seconds after the initial launch? (Super
surprised to see ~2015 on that list for that matter, but at least then we
could meaningfully have been described as not yet launched)

jtobin:
@Fang- I peeked at this and it looks reasonable enough.  Still want it merged?

Fang-:
I guess the decision on this is in @pilfer-pandex's hands?

Even if we want to do leap second smearing eventually, this should keep us accurate in the mean time. But curious to hear if anyone still has strong feelings about the "position not to respect leap seconds after the initial launch", which I only recently learned of.

jtobin:
Yeah, I agree that the issue of whether or not to adopt leap smearing (which is probably better) is separate from this quick fix of sorts.

jtobin:
@pilfer-pandex tagging you to check on this one.  My thinking here is that:

* This is fine for now
* Someone should probably implement Google-style time smearing

But the latter should be handled in another issue.

I'll merge this tomorrow unless there are any complaints!

sys/zuse.hoon
```diff
@@ -6882,28 +6882,33 @@
       ::                                                ::  ++lef:yu:chrono:
       ++  lef                                           ::  leapsecond dates
         ^-  (list @da)
-        :~  ~2015.6.30..23.59.59   ~2012.6.30..23.59.59
```
pilfer-pandex: One minor comment: For better diff legibility around things like this in the future, we should probably have one entry per line

sys/zuse.hoon
```diff
@@ -6882,28 +6882,33 @@
...
-      ::                                                ::  ++les:yu:chrono:
-      ++  les                                           ::  leapsecond days
+      ::
+      ::  +les:yu:chrono: leapsecond days
+      ::
+      ::    https://www.ietf.org/timezones/data/leap-seconds.list
+      ::
+      ++  les
         ^-  (list @da)
-        :~  ~2015.7.1  ~2012.7.1  ~2009.1.1  ~2006.1.1  ~1999.1.1  ~1997.7.1
```
pilfer-pandex: likewise here

PR #1094, opened 2019-03-04
removed hard calls on json blobs

ixv:
Stop harding json blobs in eyre; clam them instead. Mitigates the problem of eyre choking on large json blobs that landscape passes.

### DIFF
--- a/sys/vane/eyre.hoon
+++ b/sys/vane/eyre.hoon
@@ -990,7 +990,7 @@
         ?.  ?=($json p.p.cuf)
           :: ~>  %slog.`%*(. >[%backing p.p.cuf %q-p-cuf]< &3.+> (sell q.p.cuf))
           (back-turbo tee %json p.cuf)
-        (get-rush:(ire-ix p.tee) q.tee ((hard json) q.q.p.cuf))
+        (get-rush:(ire-ix p.tee) q.tee (,json q.q.p.cuf))
       ::
           $quit  (get-quit:(ire-ix p.tee) q.tee)
       ==
@@ -1066,7 +1066,7 @@
         =/  cay=cage  (result-to-cage:ford build-result.result.sih)
         %+  get-rush:(ire-ix p.tee)  q.tee
         ?>  ?=($json p.cay)                    ::  XX others
-        ((hard json) q.q.cay)
+        (,json q.q.cay)
       ==
     ==
   ::


ohAitch:
I thought clams got removed from hoon?

philipcmonk:
We use "clam" to refer to irregular `,`, which apparently expands to `$;` now.  It's not a clam in the old sense of "don't crash", but it uses the same syntax and you use it in most of the same instances.  It should be semantically equivalent to hard but faster.  In practice, bugs.

philipcmonk (approved):

ohAitch:
Huh, what is hard(or just calling json as a gate) doing that $; isn't?

philipcmonk:
Equality check at the end.  This can be very expensive.

"calling json as a gate" is equivalent to `$;`.  It's just that "calling a mold as a gate" is a really bad syntactic antipattern, and I think it should be disallowed.  It makes you have to think about why `((list [@ @]) ~)` doesn't work.  We can't at parse time distinguish a mold from a value, so we need to apply Philip's Comma Rule[0].  Much better to simply prefix all clams with `,` which has the added benefit of being more explicit about what you're doing.

[0] If (1) it's not a function (unlike pair), (2) you can't tell by context that it's a type (unlike the first argument to |= ), and (3) it's ambiguous as context-free as to whether it's a type (unlike * or @ or ^), then you need to prefix it with a comma.

ohAitch:
Interested in the documentation strategy for that proposal: "a hoon
type(mold) is just a gate. You are not however allowed to call this gate"

It sounds like the equality check should just be removed from hard? Unless
2019 clams are *inconsistent* about whether they'll crash on mistyped values

philipcmonk:
> "You are not however allowed to call this gate"

I mean, calling it a gate has always been a lie, it's just a question of degree.  You can't just do `(,|=(* +<) 5)` for example.  OTOH you can do:

```
=+  m=|=(* ?^(+< +< +(+<)))  (,m 5)
```

It may be implemented as a gate somewhere deep down, but you shouldn't pretend it's just a gate to the user.

> Unless 2019 clams are *inconsistent* about whether they'll crash on mistyped values

Bingo: `((list) [1 2 3 ~])`

ohAitch:
Sorry, I don't have a repl at hand; are you saying that shouldn't crash,
but does? Crashes but doesn't crash if you put a , on the (list)? Generally
behaves any differently from (list *)?

philipcmonk:
```
> ((list) [1 2 3 ~])
~[0 0 0]
> (,(list) [1 2 3 ~])
~[0 0 0]
> ((list *) [1 2 3 ~])
~[1 2 3]
> ((list *mold) [1 2 3 ~])
~[1 2 3]
> (,(list ,*mold) [1 2 3 ~])
~[1 2 3]
> (,(list *mold) [1 2 3 ~])
[1 2 3 0]
```

ohAitch:
Thanks!

That definitely looks like a bug yeah; is it filed? It does not however
seem related to ++hard, though using hard certainly downgrades it to a less
severe "it is *detected* that the new mold system is corrupting data"; I'm
guessing +* is putting in a hardcoded compile time equivalent of *gate
instead of *mold? (And the last one is just weird, though I definitely
wouldn't describe it as "using , is a good idea")

philipcmonk:
This is definitely a bug.  As for whether all `++hard` calls should turn into clams, the question is "do you trust that clams don't corrupt data, given that nobody's ever checked it very well and we know it does in at least one case"?

ohAitch:
I mean, either it gets checked very well, or should be reverted to the old
system known *not* to be broken and with identical semantics in the ++hard
case

PR #1096, opened 2019-03-05
added ++arm:core, +-arm:core, +$arm:core, +*arm:core to udon parser

jfranklin9000:
Before:

`++arm`
+-arm
+$arm
+*arm

`++arm`:core
+-arm:core
+$arm:core
+*arm:core

After:

`++arm`
`+-arm`
`+$arm`
`+*arm`

`++arm:core`
`+-arm:core`
`+$arm:core`
`+*arm:core`

This has PR #1091 functionality in it (parsing `:core`) and can replace it.

Should this be submitted against `next` instead of `master`?


jfranklin9000:
Without this patch
```
[++abs`:fl`](../library/3b/#--abs-fl-)
```
from `docs/reference/alphabetical-arms.udon` displays split:
split

If this patch is accepted I'll submit a patch to docs that does
```
[++abs:fl](../library/3b/#--abs-fl-)
```

It not I'll submit a patch to docs that does
```
[`++abs:fl`](../library/3b/#--abs-fl-)
```

joemfb:
bumping for @pilfer-pandex

PR #1098, opened 2019-03-06
Fix topics decoding to work with zero values

Fang-:
Because `+cat` doesn't care to left-shift if you append a zero value.

Previously:

```hoon
> (decode-topics:rpc:ethereum ~[0x0 0x1] ~[%uint %uint])
[0 1]
> (decode-topics:rpc:ethereum ~[0x1 0x0] ~[%uint %uint])
[0 1]
```

Now:

```hoon
> (decode-topics:rpc:ethereum ~[0x0 0x1] ~[%uint %uint])
[0 1]
> (decode-topics:rpc:ethereum ~[0x1 0x0] ~[%uint %uint])
[1 0]
```

### DIFF
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -7758,10 +7758,9 @@
       ::  tox:  list of hex words
       |*  [tox=(lest @ux) tys=(list etyp)]
       =-  (decode-arguments (crip -) tys)
-      %+  render-hex-bytes  (mul 32 (lent tox))
       %+  roll  `(list @ux)`tox
-      |=  [top=@ux tos=@]
-      (cat 8 top tos)
+      |=  [top=@ tos=tape]
+      (weld tos (render-hex-bytes 32 top))
     ::
     ++  decode-results
       ::  rex:  string of hex bytes with leading 0x.
--- a/sys/zuse.hoon
+++ b/sys/zuse.hoon
@@ -7755,23 +7755,19 @@
     ::
     ::  decoding
     ::
-    ++  decode-topics
-      ::  tox:  list of hex words
-      |*  [tox=(lest @ux) tys=(list etyp)]
-      =-  (decode-arguments (crip -) tys)
-      %+  roll  `(list @ux)`tox
-      |=  [top=@ tos=tape]
-      (weld tos (render-hex-bytes 32 top))
+    ++  decode-topics  decode-arguments
     ::
     ++  decode-results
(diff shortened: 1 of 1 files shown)


ohAitch:
I think ++can is the canonical way to assemble fixed-width blocks?

philipcmonk:
Ah, yes, `++can` is better.

ohAitch:
Or actually `++rep`, if they're all just `lsh 1`?

I'm also somewhat confused what the surrounding code is doing: is `render-hex-bytes` literally just uncurried `x-co:co`? In that case, it's taking a list of `@uxI`(uint256), printing them as a tape of hex characters, converting that into a cord of hex characters, and then parsing back from hex into a tuple of the exact same numbers??

ohAitch:
Yeah it looks like every single call to decode-topics accepts `~[%uint]`, `~[%uint %uint]`, or `~[%uint %address]` where `%address` is an alias for `%uint`. While I'm all for future compatibility, I feel like in this case it would be reasonable to remove the function in question entirely and replace it with a series of length checks: `?>  ?=([who=@ wer=address ~] topics.log)` and similar.

(I'm also not sure why `azimuth-events` is a core instead of a `(map @uxI term)` which you could apply to `event.log` once and then idiomatically `?-`. Or well, not literally `term`, `%- :~ my` result so that you can actually catch typoed(as currently) or unhandled(but known) event types)

ohAitch:
…why is the compiler allowing lone terms in a $% o.o https://github.com/urbit/arvo/blob/master/sys/zuse.hoon#L7608-L7617
(I think the rune you technically want here is `$?`, possibly with a `$@` split to `$%` tho tbh giving the commented static/dynamic precedence makes sense)

Fang-:
I completely forgot about `rep`. Serious tunnel-vision going on here.

Thanks for the feedback @ohAitch! I have touched up the logic here to not do superfluous `@t <--> (list @)` conversions.

Found an issue in the encoding logic while I was trying to test my changes there. We probably want more thorough tests for encoding, and tests _at all_ for decoding. Official test data (which the current tests are sourced from) don't reach further than simple cases. Currently have a Q out to Ethereum core devs for recommended test data, but if that comes up blank, I might just pull in [these](https://github.com/ethereumjs/ethereumjs-abi/blob/master/test/index.js).

Tests aside, this PR is ready for re-review.

PR #1104, opened 2019-03-12
Fixed comet sigils on landscape

ixv:
Comets just display the sigil of ~zod


vvisigoth:
Any way we can use the sigil for an invalid syllable instead of ~zod?

Screenshot 2019-03-12 15 18 02

Fang-:
imo using ~zod's sigil for comets is pretty elegant, since that otherwise won't ever show up unless you actually are ~zod. Fine with the `x` "sigil", but it feels a bit negative.

vvisigoth:
@Fang- man of the people

rmariani:
A comet should/will carry negative connotations anyway

Fang-:
@rmariani that is entirely for communities themselves to decide. Common? Probably. But not guaranteed.

ixv:
Made the sigil an X. Also reverted a landscape PR that was displaying fora posts in the inbox weirdly.

ohAitch:
Huh, what is the reasoning behind not using the comet name as an
actual visual key fingerprint?

galenwp:
I don't think an X is the right solution — 128 bits is a lot to represent. They don't look very good using the existing system.

vvisigoth:
Should we have a fallback comet sigil? Just a square?
—
~poldec-tonteg
http://urbit.org

joshuareagan:
How about an actual comet

Fang-:
> How about an actual comet

Hence, just use a literal circle, ~zod's sigil. Simple, clean, and still stands out.

vvisigoth:
How about we use the same word to refer to do diametrically opposite things?
—
~poldec-tonteg
http://urbit.org

ohAitch:
Controversial compromise idea: on talk(and possibly current landscape),
comets are represented in the general shape ~fobsum_barsum: the first and
last syllable pair, joined by an underscore to evoke part of the name being
elided. What if we use those to render a planet sigil, but then rotate it
45° and widen the middle slightly to still suggest the "x" / render as grey
instead of white/black / compose in a "signal noise" alpha mask (perhaps
based on the rest of the name) / sth, to represent the second-classhood of
the point?

ohAitch:
…or we can just not distinguish between typoed and non-typoed comets I guess

PR #1109, opened 2019-03-17
Landscape signin page

vvisigoth:
Pretty bare bones, but matches the mocks.


vvisigoth:
Specifically, this could have more advanced error states (currently doesn't check if it's a shipname, etc.)

ixv:
Might want to merge this into hotfix

ixv:
Reverting this because you included some of Fangs stuff that shouldn't be there

PR #1110, opened 2019-03-18
+ob overhaul

### LINKED ISSUE #1105: @p is not bijective
.. in other words, there exist multiple atom inputs that resolve to the same `@p` output.

For example:

```
> `@p`3.108.299.008
~morlyd-mogmev
> `@p`479.733.505
~morlyd-mogmev
```

@bencalder stumbled upon this problem early last week and brought it to our attention; I took a look into it and, after bashing around for awhile and getting some help, traced it to an off-by-one-like error in the implementation of a generalised Feistel cipher used to obfuscate `@p` values for planets.  Stars and galaxies aren't affected.

Specifically, the core of the problem is right here in `hoon.hoon`:

```
  ++  fice                                              ::  adapted from
    |=  nor/@                                           ::  black and rogaway
    ^-  @                                               ::  "ciphers with
    =+  ^=  sel                                         ::   arbitrary finite
    %+  rynd  3                                         ::   domains", 2002
    %+  rynd  2
    %+  rynd  1
    %+  rynd  0
    [(mod nor 65.535) (div nor 65.535)]
    (add (mul 65.535 -.sel) +.sel)
```

As per the [reference algorithm][algo], one needs to return `aR + L` when using an even number of Feistel rounds.  We use four rounds, but our implementation returns `aL + R`, i.e. what should be used for the *odd* parity case.  The last line should read:

```
    (add (mul 65.535 +.sel) -.sel)
```

Thus, as-implemented `fice` is not a proper permutation cipher, and we eventually get `@p` collisions for various inputs.

There appear to be roughly ~360 collisions between planets spawned by stars under any given galaxy, so probably on the order of ~100k in total (I'm basing this estimate on collisions observed between planets under each of `~zod`, `~nec`, and `~bud`, provided by @bencalder).  In other words: there are probably about ~200k planets affected in total, or about one in every 25000.  There are currently no active planets with conflicting `@p`'s, though.

This is pretty easily fixed, and I've done a lot of testing of a fixed-up version to make sure it's sound -- including an exhaustive test of the entire planetary `@p` space using [urbit-ob][objs] -- in order to completely rule out collision errors.  And this definitely needs to be fixed.

The problem, as one might expect, is that the "fix" changes every planet's `@p` (and, in turn, its associated sigil).  I'm `~nidsut-tomdun`, but am staring at the prospect of becoming `~dasref-tapteg`.

We've been discussing a few potential solutions for this internally, but want to get some public feedback before deciding on any of them.

[algo]: http://web.cs.ucdavis.edu/~rogaway/papers/subset.pdf
[objs]: http://github.com/urbit/ob-js

jtobin:
(Resolves #1105)

`+ob`, a core that implemented a permutation cipher for planet-sized atoms, contained a critical bug that allowed multiple inputs to encipher to the same output.  This created collisions between some planet-sized `@p` values.

The rewritten version corrects the bug, and also rewrites `+ob` to more clearly match both 1) the paper that the algorithm comes from, and 2) the parallel implementation in urbit-ob that has gone through a full-scale exhaustive test.

The rewritten version is also itself easier to test; this PR adds a small-scale exhaustive test, as well as small battery of sanity checks on `+ob` internals (including on `@p` values previously known to collide).

NOTE: One thing I'm observing here is that dojo may be lagging behind on changes to hoon.hoon.  For example, the following in `tests/sys/hoon/auras.hoon` passes:

```
    %+  expect-eq
      !>  ~dilwes-fadnel
      !>  `@p`23.554.048
    ::
    %+  expect-eq
      !>  ~fipfes-dilwes
      !>  `@p`529.511.092
```

but in dojo, I still get collisions:

```
> `@p`23.554.048
~dilwes-fadnel
> `@p`529.511.092
~dilwes-fadnel
```

Should I trust the test suite here?  I'd like to figure out what's going on, regardless.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -3865,9 +3865,9 @@
     =/  f  (prf (sub j 1) arr)
     ::
     =/  tmp
-    ?.  =((mod j 2) 0)
-      (mod (add f ell) a)
-    (mod (add f ell) b)
+      ?.  =((mod j 2) 0)
+        (mod (add f ell) a)
+      (mod (add f ell) b)
     ::
     $(j +(j), ell arr, arr tmp)
   ::
@@ -3882,14 +3882,14 @@
     =/  j  r
     ::
     =/  ahh
-    ?.  =((mod r 2) 0)
-      (div m a)
-    (mod m a)
+      ?.  =((mod r 2) 0)
+        (div m a)
+      (mod m a)
     ::
     =/  ale
-    ?.  =((mod r 2) 0)
-      (mod m a)
-    (div m a)
+      ?.  =((mod r 2) 0)
+        (mod m a)
+      (div m a)
     ::
     :: Similar to the comment in +fe, +fen differs from B&R (2002)'s "fe^-1"
     :: here in order to preserve the legacy cipher's behaviour on most inputs.
@@ -3898,14 +3898,14 @@
     :: correct those cases by swapping the values of 'ahh' and 'ale'.
     ::
     =/  ell
-    ?:  =(ale a)
-      ahh
-    ale
+      ?:  =(ale a)
+        ahh
+      ale
     ::
     =/  arr
-    ?:  =(ale a)
-      ale
-    ahh
+      ?:  =(ale a)
+        ale
+      ahh
     ::
     |-
     ?:  (lth j 1)
@@ -3917,9 +3917,9 @@
     ::  'f' modulo the same number before performing subtraction.
     ::
     =/  tmp
-    ?.  =((mod j 2) 0)
-      (mod (sub (add arr a) (mod f a)) a)
-    (mod (sub (add arr b) (mod f b)) b)
+      ?.  =((mod j 2) 0)
+        (mod (sub (add arr a) (mod f a)) a)
+      (mod (sub (add arr b) (mod f b)) b)
     ::
     $(j (sub j 1), ell tmp, arr ell)
   ::
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -3788,34 +3788,34 @@
   ++  feis
     |=  m=@
     ^-  @
-    (fee 4 65.535 65.536 (mul 65.535 65.536) m)
+    (fee 4 65.535 65.536 (mul 65.535 65.536) eff m)
   ::
   ::  +tail: reverse +feis.
   ::
   ++  tail
     |=  m=@
     ^-  @
-    (feen 4 65.535 65.536 (mul 65.535 65.536) m)
+    (feen 4 65.535 65.536 (mul 65.535 65.536) eff m)
   ::
   ::  +fee: "Fe" in B&R (2002).
   ::
   ++  fee
-    |=  [r=@ a=@ b=@ k=@ m=@]
+    |=  [r=@ a=@ b=@ k=@ prf=$-([j=@ r=@] @) m=@]
     ^-  @
-    =+  c=(fe r a b m)
+    =+  c=(fe r a b prf m)
     ?:  (lth c k)
       c
-    (fe r a b c)
+    (fe r a b prf c)
   ::
   ::  +feen: "Fe^-1" in B&R (2002).
   ::
   ++  feen
-    |=  [r=@ a=@ b=@ k=@ m=@]
+    |=  [r=@ a=@ b=@ k=@ prf=$-([j=@ r=@] @) m=@]
     ^-  @
-    =+  c=(fen r a b m)
+    =+  c=(fen r a b prf m)
     ?:  (lth c k)
       c
-    (fen r a b c)
+    (fen r a b prf c)
   ::
   ::  +fe:  "fe" in B&R (2002).
   ::
@@ -3823,7 +3823,7 @@
   ::    to support some legacy behaviour.  See urbit/arvo#1105.
   ::
   ++  fe
-    |=  [r=@ a=@ b=@ m=@]
+    |=  [r=@ a=@ b=@ prf=$-([j=@ r=@] @) m=@]
     ^-  @
     =+  j=1
     =+  ell=(mod m a)
@@ -3837,7 +3837,7 @@
         (add (mul arr a) ell)
       (add (mul ell a) arr)
     ::
-    =/  f  (eff (sub j 1) arr)
+    =/  f  (prf (sub j 1) arr)
     ::
     =/  tmp
     ?.  =((mod j 2) 0)
@@ -3852,7 +3852,7 @@
   ::    to support some legacy behaviour.  See urbit/arvo#1105.
   ::
   ++  fen
-    |=  [r=@ a=@ b=@ m=@]
+    |=  [r=@ a=@ b=@ prf=$-([j=@ r=@] @) m=@]
     ^-  @
     =+  j=r
     ::
@@ -3879,7 +3879,7 @@
     |-
     ?:  (lth j 1)
       (add (mul arr a) ell)
-    =/  f  (eff (sub j 1) ell)
+    =/  f  (prf (sub j 1) ell)
     ::
     ::  Note that there is a slight deviation here to avoid dealing with
     ::  negative values.  We add 'a' or 'b' to arr as appropriate and reduce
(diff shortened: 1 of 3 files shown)


sys/hoon.hoon
```diff
@@ -3747,68 +3747,159 @@
...
-    |=  vip/@
+  ::  +feis: a four-round generalised Feistel cipher over the domain
+  ::         [2^16, 2^32 - 1].
+  ::
+  ::    See: Black & Rogaway (2002), Ciphers for arbitrary finite domains.
+  ::
+  ++  feis
+    |=  m=@
+    ^-  @
+    (fee 4 65.535 65.536 (mul 65.535 65.536) eff m)
```
Fang-: This might be a good time to change these decimals into hexadecimals, here and in `++tail`. That'd be consistent with notation used higher up, which is easier to read.

jtobin:
@Fang- nice, I'll try whipping up a pill.

sys/hoon.hoon
```diff
@@ -3747,68 +3747,159 @@
...
+      c
+    (fen r a b prf c)
+  ::
+  ::  +fe:  "fe" in B&R (2002).
+  ::
+  ::    Note that this implementation differs slightly from the reference paper
+  ::    to support some legacy behaviour.  See urbit/arvo#1105.
+  ::
+  ++  fe
+    |=  [r=@ a=@ b=@ prf=$-([j=@ r=@] @) m=@]
```
belisarius222: Can you add some docs on the args here, please? I'm not sure what I'm looking at

sys/hoon.hoon
```diff
@@ -3747,68 +3747,159 @@
...
-  ++  rund                                              ::  reverse round
-    |=  {n/@ l/@ r/@}
-    ^-  {@ @}
-    :-  r
-    ?~  (mod n 2)
-      (~(dif fo 65.535) l (muk (snag n raku) 2 r))
-    (~(dif fo 65.536) l (muk (snag n raku) 2 r))
+    (feen 4 65.535 65.536 (mul 65.535 65.536) eff m)
+  ::
+  ::  +fee: "Fe" in B&R (2002).
```
belisarius222: Can you give a brief description of what this function is for?

sys/hoon.hoon
jtobin: Fixed in 0e97c1b9.

sys/hoon.hoon
jtobin: Fixed in fbfc520.

sys/hoon.hoon
jtobin: Fixed in fbfc520.

jtobin:
After scratching my head for awhile as to why a pill wouldn't work, I realised I was using the one in `zod/.urb`, *not* `zod/urb/put`.  Derp.

Using the correct one persisted my changes to dojo:

```
> `@p`145.391.618
~fipfes-hossev
> `@p`1.859.915.444
~hossev-roppec
```

joshuareagan:
fwiw @jtobin I made the exact same mistake when I was learning to make modifications to hoon.hoon. It took me way too long to realize what I was doing wrong.

sys/hoon.hoon
```diff
@@ -3750,65 +3750,187 @@
...
+  ::    a, b: parameters such that ab >= k
+  ::    k:    value such that the domain of the cipher is [0, k - 1]
+  ::    prf:  a gate denoting a family of pseudorandom functions indexed by
+  ::          its first argument and taking its argument second as input
+  ::    m:    an input value in the domain [0, k - 1]
+  ::
+  ++  fee
+    |=  [r=@ a=@ b=@ k=@ prf=$-([j=@ r=@] @) m=@]
+    ^-  @
+    =+  c=(fe r a b prf m)
```
joemfb: I think `=/(a b c)` is always superior to `=+(a=b c)`, and would be inclined to switch here and elsewhere in this PR. I'm curious to hear if anyone disagrees.

sys/hoon.hoon
```diff
@@ -3750,65 +3750,187 @@
...
+      ?:  =(arr a)
+        (add (mul arr a) ell)
+      (add (mul ell a) arr)
+    ::
+    =/  f  (prf (sub j 1) arr)
+    ::
+    =/  tmp
+    ?.  =((mod j 2) 0)
+      (mod (add f ell) a)
+    (mod (add f ell) b)
```
joemfb: The expression after `tmp` should be indented.

sys/hoon.hoon
```diff
@@ -3750,65 +3750,187 @@
...
+  ::
+  ++  fen
+    |=  [r=@ a=@ b=@ prf=$-([j=@ r=@] @) m=@]
+    ^-  @
+    =+  j=r
+    ::
+    =/  ahh
+    ?.  =((mod r 2) 0)
+      (div m a)
+    (mod m a)
```
joemfb: All the `=/` expressions here need to be indented.

sys/hoon.hoon
jtobin: I think you're probably right.  In general I prefer using less runes to more, and standardising on `=/` makes sense to me.

jtobin: Fixed in fdb9d0f5.

sys/hoon.hoon
jtobin: Fixed in c65f882.

sys/hoon.hoon
jtobin: Fixed in c65f882.

sys/hoon.hoon
```diff
@@ -3750,65 +3750,187 @@
...
+  ::    Reverses a Feistel cipher constructed with parameters described in
+  ::    +fee.
+  ::
+  ++  feen
+    |=  [r=@ a=@ b=@ k=@ prf=$-([j=@ r=@] @) m=@]
+    ^-  @
+    =/  c  (fen r a b prf m)
+      ?:  (lth c k)
+        c
+      (fen r a b prf c)
```
joemfb: This indentation is wrong now (as is `+fee` above). Sorry for causing confusion!

`=/` is "3-fixed" -- it consists of three separate sub expressions. In "flat" or "wide" form, that is simply `=/(a b c)`. In tall form:

```
=/  a  b
c
```

or, if be is substantial:

```
=/  a
  b
c
```

(Note that `a` in these examples can be just a face, or a face and a mold: `=/(a 1 a)` or `=/(a=@ud 1 a)`).

sys/hoon.hoon
```diff
@@ -3750,65 +3750,187 @@
...
+  ::    Note that this implementation differs slightly from the reference paper
+  ::    to support some legacy behaviour.  See urbit/arvo#1105.
+  ::
+  ++  fe
+    |=  [r=@ a=@ b=@ prf=$-([j=@ r=@] @) m=@]
+    ^-  @
+    =/  j  1
+    =/  ell  (mod m a)
+    =/  arr  (div m a)
+    |-
```
joemfb: I meant to mention this earlier: the idiom for `|-` is to always follow it with a `^-` cast. I would move the cast at line 3842 down here.

```
|-  ^-  @
```

sys/hoon.hoon
```diff
@@ -3750,65 +3750,187 @@
...
+      ?:  =(ale a)
+        ahh
+      ale
+    ::
+    =/  arr
+      ?:  =(ale a)
+        ale
+      ahh
+    ::
+    |-
```
joemfb: This `|-` needs a `^-` cast.

sys/hoon.hoon
jtobin: Right, ok.  That was my bad -- will fix.

jtobin: Fixed in e1fe53a.

sys/hoon.hoon
jtobin: Thanks for that one -- fixed in 27fad35.

A note: it might be worth putting together a contributor style reference of sorts, rather than the more exhaustive guide that currently exists.  I've jotted that idea down, so might get around to it before terribly long.

sys/hoon.hoon
jtobin: Fixed in 27fad35.

sys/hoon.hoon
joemfb: That'd be great.

jtobin:
I'll merge this in -- don't want to sidestep @belisarius222, but I reckon the changes meet his approval, so I'll be bold.

PR #1115, opened 2019-03-21
Ncurses support

BernardoDeLaPlaz:
creates new blits that are used by commit  4f944f810d082f30f81caaabe4ddf54a04158b18 on urbit / vere

### DIFF
--- a/sys/vane/dill.hoon
+++ b/sys/vane/dill.hoon
@@ -172,62 +172,74 @@
         ?~  wol  +>.^$
         $(wol t.wol, +>.^$ (from %out (tuba i.wol)))
       ::
-      ++  dump                                          ::  pass down to hey
+      ++  dump                                ::  pass down to hey
         |=  git/gift:able
         ?>  ?=(^ hey.all)
         +>(moz [[u.hey.all %give git] moz])
       ::
-      ++  done                                          ::  ++ done
-        |=  git/gift:able                               ::       sample: 'git' of type gift- tagged union ( zuse.hoon:616 )
-                                                        ::               (note that all will be of type $blit in practice)
-        +>(moz :_(moz [hen %give git]))                 ::       return a copy of the subject, but add a new move
-                                                        ::       to the list of moves already stored there at face 'mov'
-                                                        ::       
-      
-      ++  from                                          ::  ++ from
-        |=  bit/dill-blit                               ::       sample: 'bit' of type dill-blit (zuse.hoon:686 - a tagged union )
-        ^+  +>                                          ::       returns a modified subject
-        ?:  ?=($mor -.bit)                              ::    if tag in 'bit' == $mor (collection of blits), then
-          |-  ^+  +>.^$                                 ::       form core recursion point that returns subject
-          ?~  p.bit  +>.^$                              ::       when done, return everything
-          $(p.bit t.p.bit, +>.^$ ^$(bit i.p.bit))       ::       until then, process each sub-blit
+      ++  done                                ::  ++ done
+        |=:able                               ::     sample: 'git' of type gift-
+                                              ::     tagged union (zuse.hoon)
+                                              ::     (note that all will be of
+                                              ::     type $blit in practice)
+                                              
+        +>(moz :_(moz [hen %give git]))       ::     return a copy of the
+                                              ::     subject, but add a new move
+                                              ::     to the list of moves
+                                              ::     already stored there at
+                                              ::     face 'mov'
+
+      ++  from                                ::  ++ from
+        |=  bit/dill-blit                     ::     sample: 'bit' of type
+                                              ::     dill-blit (zuse.hoon - a
+                                              ::     tagged union )
+        
+        ^+  +>                                ::     returns a modified subject
+        
+        ?:  ?=($mor -.bit)                    ::    if tag in 'bit' == $mor
+                                              ::    (collection of blits), then
+          |-  ^+  +>.^$                       ::    form core recursion point
+                                              ::    that returns subject
           
-        ?:  ?=($out -.bit)                              ::    if tag in 'bit' == $out (output on screen), then
-          %+  done  %blit                               ::       add blits onto the 'moves' list 
+          ?~  p.bit  +>.^$                    ::    when done, return everything
+          $(p.bit t.p.bit, +>.^$ ^$(bit i.p.bit)) :: until then, process each
+                                              :: sub-blit
+          
+        ?:  ?=($out -.bit)                    ::    if tag in 'bit' == $out add
+          %+  done  %blit                     ::    blits onto the 'moves' list
           ;:  weld
             `(list blit)`~[[%lin p.bit]]
             `(list blit)`~[[%mor ~]]
             see
           ==
-        ?:  ?=($klr -.bit)                              ::    if tag in 'bit' == $klr (styled output on screen)
-                                                        ::       p data is of type stub (hoon.hoon:426)
+        ?:  ?=($klr -.bit)                     ::    if tag in 'bit' == $klr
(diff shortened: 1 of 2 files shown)


BernardoDeLaPlaz:
1) ok, have shortened all lines to <= 80
2) removed all line numbers
3) the wid that defaults to 80 is overwritten by term.c which has function u3_term_ef_winc() which is triggered by SIGWINCH and which injects new data in via u3v_plan()

Fang-:
Neat, thanks!

> which injects new data in via u3v_plan()

I have no idea how `u3v_plan` works, but maybe window re/size should be an explicit event (either now or in future revisions, don't let me scope-creep you too hard). That would allow dill to notify userspace to force a re-draw accounting for the new size.

BernardoDeLaPlaz:
> I have no idea how u3v_plan works, but maybe window re/size should be an explicit event (either now or in future revisions, don't let me scope-creep you too hard). That would allow dill to notify userspace to force a re-draw accounting for the new size.

u3v_plan() is how the C code injects an event into Arvo.

the event that gets injected is [ $blew [80 25] ] (or whatever the actual screen resolution is)

and this seems to get processed in dill.hoon in ++call where it is dispatched to ... I'm not exactly sure where, bc my arvo knowledge is not complete.

Fang-:
Ohhh, I see now. That makes total sense, and it looks like `%blew` already works as a CLI event then. (Pretty sure dill sends it off to whoever has running `sole` subscriptions.) Perfect, no work needed there! Thanks for explaining, @BernardoDeLaPlaz. (:

yosoyubik:
I gather that this is still a work in progress but I was so excited to try it out that I couldn't stop myself :)

I think that the last commit has introduced some typos (Lines 181 and 272) but after fixing them 
- L181: `|=:able` -> `|=  git/gift:able`
- L272: `:*  [hen %pass /heft/ames %a %wegh ~ ~]` -> `:*  [hen %pass /heft/ames %a %wegh ~]` 

I get a `nest-fail` & `%load-lost` after doing a `|reset`, not sure where the error is or means...

joemfb:
@yosoyubik, that error most likely means that the state of dill is changed incompatibly by the `|reset` (ie, the type of the old state does not *nest* in the type of the new state, preventing the `+load` arm - whereby the state is "installed" in a vane - from being called in a type-safe manner).

The compatible way to fix this is add a conversion from the old state to the new state in the `+load` arm, the easy way is just to make a new pill without loading the vane into a running ship, and then boot into that (turning off automatic vane loading with `|autoload`, copying in the changed `sys/` files, and making a new pill with `+solid`). See https://github.com/urbit/urbit/blob/master/CONTRIBUTING.md for some more details.

As for this state of this PR: I owe it a review ...

yosoyubik:
Thanks @joemfb I oversaw the instructions @BernardoDeLaPlaz added here: https://github.com/urbit/urbit/pull/1228#issue-263270629

sys/vane/dill.hoon
```diff
@@ -172,111 +172,104 @@
...
-          ?~  p.bit  +>.^$
-          $(p.bit t.p.bit, +>.^$ ^$(bit i.p.bit))
-        ?:  ?=($out -.bit)
-          %+  done  %blit
-          :~  [%lin p.bit]
-              [%mor ~]
-              [%lin see]
-              [%hop pos]
+      ++  done                                ::  ++ done
+        |=:able                               ::     sample: 'git' of type gift-
```
Fang-:
```suggestion
        |=  git/gift:able                     ::     sample: 'git' of type gift-
```

sys/vane/dill.hoon
```diff
@@ -172,111 +172,104 @@
...
+          `blit`[%fog %g]         :: reset foreground
+          `blit`[%bog %k]         :: reset background
+          `blit`[%eff ~]          :: reset effects to normal                
+        ==
       ::
       ++  heft
         %_    .
             moz
-          :*  [hen %pass /heft/ames %a %wegh ~]
+          :*  [hen %pass /heft/ames %a %wegh ~ ~]
```
Fang-:
```suggestion
          :*  [hen %pass /heft/ames %a %wegh ~]
```

sys/zuse.hoon
```diff
@@ -678,11 +678,16 @@
     $%  {$bel ~}                                        ::  make a noise
```
Fang-: This says `old blit`. Theres `++ dill-blit :: new blit` further down. Do you know what the deal is with that? This seems like a good time to unify those, or clean them up, whichever's desired.

`/mar/bill/blit` exists, and it uses `++dill-blit`, so maybe that's where these changes should go, and have `++blit` be removed? Using `++dill-blit` exclusives, like `%url`, crashes with abort.

Fang-: Actually, that second paragraph is incorrect issue analysis. `%url` appears in both. But it crashes nonetheless!

sys/zuse.hoon
```diff
@@ -678,11 +678,16 @@
...
         {$clr ~}                                        ::  clear the screen
         {$hop p/@ud}                                    ::  set cursor position
```
Fang-: Should this maybe be `[x=@ud y=@ud]` now? You can't partially change lines otherwise.

Fang-:
> I've asked for but gotten zero bug reports / feature requests.

Woah hey hang on! I listed a bunch of things on urbit/urbit#1228. Are those not valid bug reports?

> someone to coordinate the two PRs

These should just go on @jtobin's personal `next` list.

I'd be happy to help out with these PRs, since I too would love to get them in. I think I'll have *some* time available for this soon, but have a healthy fear of the Vere changes for this. Am talking with @vvisigoth today and will bring it up.

Fang-:
> I've asked for but gotten zero bug reports / feature requests.

Woah hang on! I certainly did report a list of bugs on urbit/urbit#1228. Don't those count?

> (b) someone to coordinate the two PRs

I think this falls into @jtobin's hands? He'll have to put this on his `next` list once it's ready.

Since I, too, would love for this to make it in, I'd be happy to help out where I can. I _think_ I should have some time available for this on the short term. Vere seems a little scary here, but maybe it's not that bad. Am talking with @vvisigoth today and will bring it up.

BernardoDeLaPlaz:
@Fang-  my apologies; I was heads down when those went in and didn't see them!

BernardoDeLaPlaz:
@jtobin 

I wrote the C and Hoon code before I knew about breaching vs non-breaching changes.  The upshot is that the project is not easy to roll out.  Also, I've heard from a few people that they'd like it to work in manner X or manner Y. ... but X and Y are not well defined.

So the upshot is that 
(a) it may or may not be what people want.  I've asked for but gotten zero bug reports / feature requests.
(b) it can only be merged in as a breaching change, coordinating the Arvo and Urbit PRs.

So, yes, it's stagnating.

I'd LOVE to push it across the finish line, but the next steps ( (a) - crisp bug reports or a stated willingness to accept it as is; (b) someone to coordinate the two PRs ) are things that I need someone else to do.

Also, I'm currently bogged down in nix / gdb hell on the 64 bit vere project, and don't feel like I can take my eyes off of that right now.

Thoughts?

cc @vvisigoth

jtobin:
@BernardoDeLaPlaz Roger.  Given that this seems to need a bit more work, and that it appears it might not actually be wrapped up for some time, for now I think the best thing to do is to close the active pull request.  Feel free to just reopen it or submit a different one when you reckon the work is about ready to be merged.

Re: coordinating PRs, fear not: just fire them at the requisite repos and I will sort them out for you. 😄

Fang-:
But I imagine these PRs will remain the primary location for discussion of the actual dev happening here? Kinda weird to do that in a closed PR imo. Oh well.

PR #1118, opened 2019-03-22
 adds a path for externally-routed packets to %ames

joemfb:
This is specifically for use with the reverse (reverse) tcp proxy in vere, but could be used by any external arvo caller to send / receive %ames packets.


sys/vane/ames.hoon
```diff
@@ -1451,6 +1451,16 @@
     ?:  ?=(%crud -.kyz)
       [[[hen [%slip %d %flog kyz]] ~] +>]
     ?:  ?=(%west -.kyz)
+      ::  a .cha path of /a/give/* is immediately acknowledged and
+      ::  routed directly out to unix
+      ::
+      ?:  ?=([%give *] q.kyz)
+        ~|  [%ames-bad-duct %give duct=gad.fox p.kyz]
```
belisarius222: This will always print out `duct=~` on failure. Not sure if that's a problem, but might be worth cleaning up.

joemfb: Thanks, it should probably just print the sender and route-path.

PR #1119, opened 2019-03-22
Landscape integration into lighter than eyre

ixv:
This PR includes:
- A generator that serves the web files responsible for landscape.
- The js changes corresponding to this landscape commit: https://github.com/urbit/landscape/commit/e33f5a02a75d5cd0457c03950136c980faf37fa0
- A prototype of the modulo app, which contains some app switching logic, but currently only creates the binding for the landscape generator on boot

cc @loganallenc


ixv:
This isn't ready yet by the way. Collections do not work cross-ship yet.

ixv:
Cross-ship collections now work, this is ready for review

vvisigoth:
Whoa.
—
~poldec-tonteg
http://urbit.org

app/landscape.hoon
```diff
@@ -0,0 +1,246 @@
...
+      [%'~landscape' %stream ~]
+    =/  stream-html=octs  (manx-to-octs (index stream))
+    :_  this
+    [ost.bol %http-response (html-response:app stream-html)]~
+  ::
+  ::  collections top level page
+  ::
+      [%'~landscape' %collections @t @t ~]
+    =/  shp/@p    (slav %p i.t.t.site.request-line)
+    =/  col/@da   (slav %da i.t.t.t.site.request-line)
```
joemfb: Old style mold syntax, plz `shp=@p`

app/landscape.hoon
```diff
@@ -0,0 +1,246 @@
...
+    ::  html format
+    =/  pax=path  /xship/top/(scot %da col)
+    :_  this
+    [ost.bol %peer pax [shp %landscape] pax]~
+  ::
+  ::  collections new post page
+  ::
+      [%'~landscape' %collections @t @t %new ~]
+    =/  shp/@p    (slav %p i.t.t.site.request-line)
+    =/  col/@da   (slav %da i.t.t.t.site.request-line)
```
joemfb: Here too, and elsewhere in this app

app/modulo.hoon
```diff
@@ -0,0 +1,212 @@
...
+::
+++  this  .
+::
+++  prep
+  |=  old=(unit *)
+  ^-  (quip move _this)
+  ?~  old
+    :_  this
+    :~  [ost.bow %connect / [~ /] %modulo]
+        [ost.bow %connect / [~ /'~landscape'] %landscape]
```
joemfb: Any particular reason this binding is created here instead of in `app/landscape`?

app/modulo.hoon
```diff
@@ -0,0 +1,212 @@
...
+  ^-  (quip move _this)
+  :_  this
+  [ost.bow %diff %json (session-as-json cur.sta session.sta order.sta)]~
+::
+++  session-js
+  ^-  octs
+::  ?~  cur.sta
+::    *octs
+  %-  as-octs:mimes:html
+  %-  crip
```
joemfb: `(as-octt ...)` is the same as `(as-octs (crip ..))`

app/modulo.hoon
```diff
@@ -0,0 +1,212 @@
...
+::  ;:  weld
+::      (trip 'window.onload = function() {')
+      "window.ship = '{(trip (rsh 3 1 (crip <our.bow>)))}';"
```
joemfb: I'd be inclined to write this as `{+:(scow %p our.bow)}`

lib/http.hoon
```diff
@@ -0,0 +1,35 @@
+::
```
joemfb: Is this file actually be used by anything, or was it spuriously re-added in https://github.com/urbit/arvo/commit/8eac489504589d23fe3108d9e59b4de4f1097bdb#diff-745c223f983f23c1162e3fc8932996b4 ?

ixv: spurious re-add

app/modulo.hoon
ixv: No, just leftover from the brief detour into making landscape a generator. Went back to an app for the sake of cross-ship posting, I'll change this.

sys/vane/rver.hoon
```diff
@@ -627,9 +627,7 @@
     }
   };
 
-  export function newChannel() {
-    return new Channel;
-  }
+  window.urb = new Channel();
```
eglaysher: My understanding is that this makes a single channel object on window, instead of giving the caller the option of how this library gets used. If the ES6 syntax for this isn't cross-browser compatible, is there a way to make this at least export newChannel() to callers and have callers use it instead of having a library modify a global object?

sys/vane/rver.hoon
```diff
@@ -1333,10 +1337,12 @@
       |=  [channel-id=@t json-text=wall]
       ^-  [(list move) server-state]
       ::
-      =/  channel=channel
-        (~(got by session.channel-state.state) channel-id)
+      =/  channel=(unit channel)
+        (~(get by session.channel-state.state) channel-id)
+      ?~  channel
+        [~ state]
```
eglaysher: Please add a ~& here; I'm still curious about actually figuring out why we can get into a state where the channel has stopped existing but is still receiving callbacks.

ixv: Is there anything in particular you want me to print out, or just a term to indicate it got here?

sys/vane/rver.hoon
loganallenc: In the current JS, Channel becomes a globally available type that any application can instantiate.
The ES6 syntax was only supported in Chrome. I see a good argument for not instantiating `window.urb = new Channel();` however we do not need to export the function newChannel(), as the Channel class will already be available.

sys/vane/rver.hoon
eglaysher: Indicating that it got there, though you might as well throw the channel-id in the logging statement. I'm not convinced it can get here after the changes to subscribe/unsubscribe, so traces showing it can would be super interesting.

sys/vane/rver.hoon
eglaysher: OK, then just remove the assignment here and use Channel where needed.

app/modulo.hoon
```diff
@@ -90,9 +90,11 @@
 ::  ?~  cur.sta
 ::    *octs
   %-  as-octt:mimes:html
-::  ;:  weld
+  ;:  weld
 ::      (trip 'window.onload = function() {')
       "window.ship = '{+:(scow %p our.bow)}';"
+      "window.urb = new Channel();"
```
eglaysher: Looks legit.

sys/vane/rver.hoon
```diff
@@ -1340,6 +1338,7 @@
       =/  channel=(unit channel)
         (~(get by session.channel-state.state) channel-id)
       ?~  channel
+        ~&  [%recieved-event-for-nonexistent-channel channel-id]
```
joemfb: "*i* before *e*, except after *c*..."

lib/http.hoon
Fang-: [):](https://github.com/Fang-/djay/blob/master/app/djay.hoon#L4)

Are we gonna get an alternative "build http request structure, but slightly easier" lib?

joemfb: Perhaps a `lib/client`, to go with the `lib/server` added in this PR? In any case worth noting that this PR (and the others targeting `lighter-than-eyre`) are just finishing the development of `lighter-than-eyre`. There will be final PRs here and on urbit/urbit, wherein the whole effort should receive (any necessary) critical reviews.

ohAitch:
Catching breaking changes incrementally as they occur is a lot easier than
trying to dig for them all at once - generally, code review effort is
superlinear in the amount of code that needs to be looked at, because after
combination the information on what changes depend on each other vs don't
is rather involved to rederive.

PR #1120, opened 2019-03-23
Aquarium and pH

philipcmonk:
A lot happening here.  I've sent out a usage document, which should probably become public documentation at some point.

Main changes:

- Aquarium created
- pH created
- Clay queues incoming changes so that no more than one change happens in a single event
- Various small sequencing changes in vanes, to improve robustness
- Clay stores the rift number that we had for the foreign ship with its desks.  We should use this information for individual breaches, but this doesn't attempt to solve that.


sys/zuse.hoon
```diff
@@ -7199,11 +7200,13 @@
...
         ::  ++  azimuth  0x308a.b6a6.024c.f198.b57e.008d.0ac9.ad02.1988.6579  ::  ropsten
-        ++  azimuth  0x223c.067f.8cf2.8ae1.73ee.5caf.ea60.ca44.c335.fecb  ::  mainnet
+        ::  ++  azimuth  0x223c.067f.8cf2.8ae1.73ee.5caf.ea60.ca44.c335.fecb  ::  mainnet
+        ++  azimuth  0x863d.9c2e.5c4c.1335.96cf.ac29.d552.55f0.d0f8.6381  ::  local bridge
         ::
         ::  launch: block number of azimuth deploy
         ::
-        ++  launch  6.784.800
+        ::  ++  launch  6.784.800  ::  mainnet
+        ++  launch  0  ::  local bridge
```
joemfb: I think these ethereum address/block changes should be backed out.

joemfb:
@philipcmonk, how long do some of these basic tests task? It'd be nice to wire something up for CI, as an example for future testing efforts if nothing else.

philipcmonk:
Haven't timed it, but something like 5-10 minutes on my machine.  It's significantly faster if you don't start collections, acme, dns, and hall.  If it's easy to wire up to CI, I'd say we should do it.  CI is changing in cc-release, so we also could just hook it up to CI there.

app/aqua-ames.hoon
```diff
@@ -0,0 +1,81 @@
+::  This needs a better SDN solution.  Every ship should have an IP
+::  address, and we should eventually test changing those IP
+::  addresses.
+::
+::  For now, we broadcast every packet to every ship and rely on them
+::  to drop them.
+::
+/-  aquarium
+=,  aquarium
+=>  $~  |%
```
belisarius222: extraneous `$~` ... I don't understand why this doesn't break everything, but it should still be removed

app/aqua-ames.hoon
```diff
@@ -0,0 +1,81 @@
...
+++  emit-aqua-events
+  |=  aes=(list aqua-event)
+  %_(this aqua-event-list (weld aqua-event-list aes))
+::
+++  poke-aqua-vane-control
+  |=  command=?(%subscribe %unsubscribe)
+  :_  this(subscribed =(command %subscribe))
+  (aqua-vane-control-handler our ost subscribed command)
+::
+++  diff-aqua-effects
```
belisarius222: A little comment explaining that this ignores effects other than %restore or %send might be helpful.

app/aqua-behn.hoon
```diff
@@ -0,0 +1,126 @@
...
+    =<  ..abet-pe(pier-data *pier)
+    ?~  next-timer
+      ..abet-pe
+    cancel-timer
+  ::
+  ++  handle-restore
+    ^+  ..abet-pe
+    =.  this
+      %-  emit-aqua-events
+      [%event who [//behn/0v1n.2m9vh %born ~]]~
```
belisarius222: Does this need to be the same `/0v1n.2m9vh` as in the ames handler? Maybe that constant should be split off into sur/aquarium/hoon.

app/aqua-behn.hoon
```diff
@@ -0,0 +1,126 @@
...
+      ?~  next-timer
+        ..abet-pe
+      cancel-timer
+    ?~  next-timer
+      (set-timer u.tim)
+    (set-timer:cancel-timer u.tim)
+  ::
+  ++  set-timer
+    |=  tim=@da
+    =.  tim  +(tim)  ::  nobody's perfect
```
belisarius222: Human foibles aside, why does this get incremented? Also, a comment saying that this uses the host Behn and explaining the overall aqua timer model might be useful.

app/aqua.hoon
```diff
@@ -0,0 +1,546 @@
...
+          [%aqua-events aqua-events]
+          [%aqua-boths aqua-boths]
+      ==
+    ::
+    +$  state
+      $:  %0
+          pil=pill
+          assembled=*
+          tym=@da
+          fleet-snaps=(map term (map ship pier))
```
belisarius222: Do :fleet-snaps and :piers refer to the same ships? If so, it might be cleaner to have :fleet-snaps be just a `(map term (set ship))` so there's less risk of the two getting out of sync.

app/eth-manage.hoon
```diff
@@ -34,7 +34,7 @@
       %look-ethnode
     :_  ~
     =/  pul
-      (need (de-purl:html 'http://eth-mainnet.urbit.org:8545'))
+      (need (de-purl:html 'http://localhost:8545'))
```
belisarius222: This looks like an artifact left in from local testing. I suspect it might break some production ships if we merge it.

sur/aquarium.hoon
```diff
@@ -0,0 +1,96 @@
+::
+::  Traditionally, ovo refers an event or card, and ova refers to a list
```
belisarius222: ovo refers to an event

sur/aquarium.hoon
```diff
@@ -0,0 +1,96 @@
...
+::  of them.  We have several versions of each of these depending on
+::  context, so we do away with that naming scheme and use the following
+::  naming scheme.
+::
+::  Every card is either a an `event` or an `effect`.  Prepended to this
```
belisarius222: either an `event`

sys/vane/gall.hoon
```diff
@@ -728,10 +728,16 @@
     ++  ap-fill                                         ::  add to queue
       ^-  {? _.}
       =+  suy=(fall (~(get by qel.ged) ost) 0)
-      ?:  =(20 suy)
-        [%| +]
-      ::  ~?  !=(20 suy)  [%ap-fill-add [[our dap] q.q.pry ost] +(suy)]
-      [%& +(qel.ged (~(put by qel.ged) ost +(suy)))]
+      =/  subscriber-ship  p:(~(got by sup.ged) ost)
+      ?:  &(=(20 suy) !=(our subscriber-ship))
```
belisarius222: Huh, looks like you turned off the culling if our own ship is the one subscribed. Does that still work for web stuff?

app/aqua-behn.hoon
joemfb: That value is `u3A->sen`, the "instance number" in vere. It's set to `(scot %uv (mug now))` on boot/restart. I think the idea is to ensure there's a new duct corresponding to the new process, but it's not used for effect routing (or for injecting `%wake` events). I don't think it's needed here. It might not be needed at all, but that's a larger conversation.

sur/aquarium.hoon
joemfb: You're both wrong: `ovo` is the conventional name for an `+ovum`, which is `(pair wire card)`. An "event" is `(pair @da ovum)`.

joemfb:
@philipcmonk, i think it's worth adding to CI. It's not particularly complicated to add now, and then re-add once arvo/next converges with vere/cc-release. Duplicating the `|mass` implementation (here: https://github.com/urbit/arvo/blob/next/.travis/test.js#L77) gets you most of the way, you just need to scrape for an error condition (a la https://github.com/urbit/runner-js/blob/master/lib/actions.js#L26). I can help if runner-js gives you trouble.

app/aqua-behn.hoon
philipcmonk: I just cargo-culted that.  The right answer is probably for aqua to keep track of that, change it on restart, and insert it into each of these wire as necessary.  For now, I've just copied the same string everywhere.

app/aqua-ames.hoon
philipcmonk: Ha, good catch.  It doesn't break anything because (1) $~ is a mold, so it's a core with the normal subject in its context and (2) Ford compiles files as a list of expressions implicitly tisgar'd together.

app/aqua-behn.hoon
philipcmonk: Various things break if you have two events at the same time.  I think this is unnecessary now because (1) behn no longer responds immediately to a timer set in the current event (it used to do that under certain circumstances), and (2) aqua enforces that each event happens at a different time anyway.  I'll remove it.

app/aqua.hoon
philipcmonk: `fleet-snaps` is a set of snapshots of the state of the ships.  It's intended to freeze a fleet so you can restore back to that point later on.  This is how pH can cache e.g. "start ~bud" so that you only have to do it once rather than once per test.

sys/vane/gall.hoon
philipcmonk: I tried a bunch of stuff to reduce the number of subscription updates in the queues between aqua/ph and aqua-*, and I couldn't come up with anything that would work reliably.

I'm not sure how web stuff uses this?  Is that the only way we clean up subscriptions created by web pages?

philipcmonk:
Added it to CI.  Now CI takes ~25 minutes.  Looks like the old builds took between 8 and 17 minutes (not sure what causes them to be longer; maybe changes to /sys?).

ohAitch:
Changes to sys/ definitely *should* take longer, as they build a new pill;
are there many that are not close to 8 or close to 17 minutes?

ohAitch:
Also, would it be possible to add [custom folds](https://github.com/macbre/travis-fold) for individual test phases showing more granular timings? Even 8 minutes is rather a lot.

(5 of those is the actual test script, 8 in the pill case; it looks like we're maybe building the pill twice??)

joemfb:
@ohAitch, we have to build a second pill to correctly capture userspace changes (which are now stored in the pill).

I think the right approach is to a) split the different test types across different CI jobs and b) cache and promote build artifacts more efficiently. The right place will be on top of cc-release, which now uses nix and make as the build system, and urb.py as the test runner. The right time will be when somebody is free to work on it.

ohAitch:
Hm, I don't see it being built at all in the 8 minute cases though, is this
a master vs cc-release difference? (And even if you have to add userspace,
recompiling everything from scratch seems superfluous given you could
instead just mutate the "userspace files" event at the end of the pill)

Agree that the Travis build matrix seems useful here, if most of the test
time is in fact straightforwardly parallelizable.

joemfb:
It's still only built on changes to sys, but when it's built it's build twice twice: first to stage the kernel changes, second to make the complete up-to-date pill that will be uploaded. This is of course stupid and inefficient, but was the easiest way to make the new pills not wrong. My grand plan is to let the pain of these slow builds increase until someone (maybe me!) snaps and makes it fast.

ohAitch:
In that case it is certainly correct to merge the 25-minute version B)

PR #1121, opened 2019-03-25
Add a readme

Fang-:
Comes with usage instructions largely copied from `urbit/urbit`, minimum viable developer instructions, minimum viable contributor direction, and minimum viable flair.

Happy to take suggestions for better copy right now, or punt on that and merge as is. This will definitely need a rewrite at some point.

Also, how do we want to deal with Arvo's lack of `.md` support? We'll want a readme **`.md`** for Github presentation reasons, right?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,58 @@
+# Arvo
+
+A clean-slate operating system.
+
+## Usage
+
+To run Arvo, you'll need [Urbit](https://github.com/urbit/urbit/). To install Urbit and run Arvo please follow the instructions in the [getting started docs](https://urbit.org/docs/getting-started/). You'll be on the live network in a few minutes.
+
+If you're doing development on Arvo, keep reading.
+
+## Documentation
+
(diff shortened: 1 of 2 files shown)


vvisigoth:
V nice. Yeah, I think we have to have the .md file for GitHub purposes. Too
bad %clay barfs when it sees this. I'm not sure what the answer is here
besides making a bogus .md mark that just renders as text?
—
~poldec-tonteg
http://urbit.org

ohAitch:
This is probably going to need special sync blacklisting similar to
LICENSE.txt, so we don't have an illegal-@ta /=home=/'README'/md? In which
case lack of arvo support should be fine.

ohAitch:
(Or possibly generic "if a file name is in all caps / not valid knot, don't
sync it", though that seems likely to be confusing and generally produce
"false" positives wrt user expectations) (I guess if there's a "file with
invalid name not synced" printf that migh be fine?)

PR #1125, opened 2019-03-28
lighter-than-eyre/:dns app integration (manual oauth)

joemfb:
This PR finalizes the integration of the :dns app with the new %lient/%rver lighter-than-eyre vanes, and adds a manual, explicit Oauth2 implementation to the app (since the security-driver dispatch mechanism no longer exists).

 Note that while functionality is complete and tested, the interfaces to the Oauth2 functionality will likely need additional polish.


eglaysher:
Immediate comment after skimming the heads of each file: I suspect you can delete lib/oauth1.hoon, lib/oauth2.hoon, lib/interpolate.hoon and lib/hep-to-cab.hoon.

ohAitch:
lib/oauth1 and lib/oauth2 target the removed eyre functionality yeah;
probably still good to keep around as reference until there's a less XXful
manual implementation which handles things like refresh tokens? Like the
/bottom/ core of each is not long for this world obviously, being shaped
for the eyre interface, but at the very least they seem like correct
long-term names for the respective standards' state machine code, as
opposed to a "library core" in the middle of an app header.

On a less related style note, confused by the misindented toplevel "=> |%"s
- is this a new ford thing?

joemfb:
@ohAitch, not sure it's mis-indented -- I just prefer explicit `=>` to the implicit `=~` of %ford.

As for the oauth libraries and security drivers, I had originally intended to replicate the pass-thru request functionality in an app, but that ended up being a larger effort than I was willing to take on (especially given only one client). I think tried to use the security driver directly as a library, but it too was more trouble than I found it to be worth.

I'm inclined to share your concern about preserving obsoleted implementations as a reference if nothing else, but also respect @eglaysher's aim of keeping dead code out of the repository. It's not very difficult to recover deleted files from the repository history, and the majority of the deletions are isolated to a couple of pull requests.

Finally, I'll note that this PR is from a short-lived feature branch to a longer-running feature branch. The final integration of the larger lighter-than-eyre effort will be a PR against the `next` branch, in the relatively near future.

ohAitch:
As the original implementer of the eyre functionality, can confirm that
doing it in userspace was difficult which is why I gave up and stuck it in
eyre :P

But that was serving a far more broadly scoped "one of urbit's top
priorities is connecting to external services, you can even interact with
API URLs directly in the dojo and it Just Works" - we've definitely moved
away from that, and I'm on board with deleting the top-level `sec/`
directory(+ associated machinery) on that basis. This doesn't preclude
factoring the particulars of oauth into a(n explicitly invoked) library
however; though you're probably right that waiting for a second library
consumer would make sense.

ohAitch:
As for the =>, I think in that case I would suggest an explicit non-ford
=~, but regardless "first child of the rune is outdented to the same level
as the second child" is definitely not correct tall form

joemfb:
Somehow I had come to think that

```
=>  |%
++  foo
...
```

had become a blessed style, but I looked around and couldn't find examples. I tried an explicit `=~`, but didn't really like it, especially the explicit `==` terminator. So I've switched back to the implicit `=~`.

ohAitch:
Fwiw  %- foo :~  is doing something similar, and occurs in a number of
places, as does  |. ^+ foo

Like in general it's reasonable to not do 100% strict backstep form, but
the "spirit" is like, "all the children are on different indentation levels
/ easy to visually distinguish from one another"

PR #1127, opened 2019-03-29
add +grad to md mark

philipcmonk:
The md mark didn't have a `+grad`, so it crashed if you ever changed an md file.

It also lacked a trailing newline.


joemfb:
Longstanding complaint of yours truly: #175

PR #1135, opened 2019-04-09
Behn pass errors to client vanes

belisarius222:
This sets the groundwork for future work for vanes to handle async errors. I just added stubs to the various vanes for now, but at least they now have the error information.

A `|reset` passed, but I have not done more detailed testing on this. Not sure if there's a convenient way to test. Maybe pH?

### DIFF
--- a/app/aqua-behn.hoon
+++ b/app/aqua-behn.hoon
@@ -58,12 +58,12 @@
 ::  Received timer wake
 ::
 ++  wake
-  |=  [way=wire ~]
+  |=  [way=wire error=(unit tang)]
   ^-  (quip move _this)
   =.  this  apex  =<  abet
   ?>  ?=([@ *] way)
   =/  who  (,@p (slav %p i.way))
-  abet-pe:(take-wake:(pe who) t.way ~)
+  abet-pe:(take-wake:(pe who) t.way error)
 ::
 ++  pe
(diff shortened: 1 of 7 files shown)


belisarius222:
I resolved a merge conflict in Clay by checking both the error in the `%wake` move and whether the wire is `%tyme`.

belisarius222:
@philipcmonk I think I fixed Aqua; we'll see what the tests say. I added `%crud` to `+unix-task` in Zuse, and I modified `:aqua-behn` to send one as an aqua-event when it receives a Behn error. If you'd prefer different handling for that case, let me know and I can change it.

belisarius222:
Tests passed; ready for review.

joemfb:
@belisarius222, since this is going into `next`, do you want to coordinate the `%crud` protocol change in vere at the same time? I could do that if you'd like.

belisarius222:
@joemfb I'm down, although I think we should get this in as-is first. I think it'll be a little cleaner that way, although I could be convinced otherwise.

joemfb:
It looks to me like all the apps that set timers will fail if the are called with an error `%wake`. It looks like gall just prints an error in this case, but doesn't crash -- if that's right, %behn will still be unblocked in the event of app timer failure, which is the most important result.

PR #1139, opened 2019-04-12
suppresses %clay file-change printfs for initial filesystem

joemfb:
This is not a pressing issue -- consider it a UX feature request.

I've long found these printfs to be superfluous, but it's gotten much worse (imo) with the growth of userspace. This is really highlighted when you boot ships in :aqua.

The printfs are replaced with

```
clay: committed initial filesystem (hoon)
```

for `base/1`,  and

```
clay: committed initial filesystem (all)
```

for `base/2`

### DIFF
--- a/lib/ph.hoon
+++ b/lib/ph.hoon
@@ -330,7 +330,7 @@
       ::
       :~
         ?.  %^  is-dojo-output  her  who  :-  uf
-            "+ /{(scow %p her)}/base/2/web/testing/udon"
+            "clay: committed initial filesystem (all)"
           ~
         [%test-done &]~
       ::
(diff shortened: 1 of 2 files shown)


philipcmonk:
I think this is a good change.  lib/ph.hoon uses these printfs to guess when the ship has finished booting, which is why CI is failing.  You should be able to just switch the line that searches for "/base/2/web/"... to the new printf.

joemfb:
Thanks.

It looks like the `%child-sync` test hung again, but this time in the "push" build instead of the "pr" build. I don't think that's related to my changes, but it's hard to be sure. Any tips on debugging these?

PR #1141, opened 2019-04-12
Fix app behn error handling

belisarius222:
I hadn't changed the type signature for +wake in all the apps after changing Behn's %wake +gift type.


belisarius222:
@ixv Thanks for the fix

PR #1143, opened 2019-04-13
no-op on bad ames packet mugs

ixv:
This is mainly just to alleviate some pain while doing devops on zod, which continually gets spammed from ships in past eras, with the correct ames protocol number but bad packet mug hashes.

Normally I just edit the ames on zod to include this logic but I see no reason why it shouldn't go into the repo.


belisarius222:
This would be cleaner if we changed +bite to produce a (unit cake) instead of inlining it. If it's only gonna be here for a few days, then this is fine, but otherwise we shouldn't duplicate the code.

ixv:
Fair enough, I assumed +bite was called elsewhere so I didn't want to disturb it, but I was mistaken.

PR #1145, opened 2019-04-14
Do notation

philipcmonk:
I propose a new synthetic rune.  I'm open to suggestions on the name; as a placeholder I'm using `;#`.

I've tagged some of the Haskell people for review, but I'm also particularly interested in @frodwith, @belisarius222, and @joemfb's opinion.

### Motivation

Suppose we have a `ph` monad (as I do in the philip/individual-breaches branch), where "M a" is `(ph-data a)` and we have operations `ph-return` and `ph-bind`.  These have signatures:

```
++  ph-return
  |*  a=mold
  |=  arg=a
  ^-  (ph-data a)
  !!
::
++  ph-bind
  |*  a=mold
  |*  b=mold
  |=  [m-a=(ph-data a) fun=_|~(a *(ph-data b))]
  ^-  (ph-data b)
  !!
```

Then, we can chain together operations on this monad to form larger state machines.  In the ph example, we use this to create integration tests that are fully composable and wrappable.  This sort of composability of state machines has hardly ever been attained in Hoon code in the past.

Here's how we can chain together two operations, each of which produce values that any later operation can use.

```
%+  ((ph-bind ,[%booted who=@p]) ,~)
  (boot ~bud ~)
|=  [%booted who=@p]
%+  ((ph-bind ,[%boot-done whom=@p]) ,~)
  (check-ship-booted who)
|=  [%boot-done whom=@p]
~&  [who=who whom=whom]
((ph-return ,~) ~)
```

This is wonderfully composable and horribly ugly.  In my opinion, it's worth using even this because excessive boilerplate is better than unrolling the state machine.

However, even though reading this code is hard, writing it is fairly mechanical.  This suggests that a synthetic rune could clarify the code.

### Existing runes

Let's look at why our current runes don't handle what we need.

`;~` is the Kleisli arrow.  While this is useful for some things, it's not as generally useful as a convenient syntax writing in monadic style.  Consider:

```
%.  (boot ~bud ~)
;~  ph-bind
  |=  [%booted who=@p]
  (check-ship-booted ~bud)
::
  |=  [%boot-done whom=@p]
  ~&  [who=who whom=whom]
  ((ph-return ,~) ~)
==
```

This doesn't compile for a couple of reasons.  First, this approach only works if you can keep ph-bind sufficiently wet (as in `+biff`).  Working with wet gates in this manner in extremely finicky and requires many explicit casts to get right.

Secondly, the `~&` fails because `who` is not in scope.  This is killer because it means that to keep stuff in your subject you need to explicitly pass it from one operation to the next, which is extremely cumbersome.

`;:` is a left-associative rune that turns `;(add a b c)` into `(add (add a b) c)`.  This is useful for functions that are not generic, but it can't be used for generic functions, since each invocation requires a new version of the function.  Again, with sufficiently wet gates you may be able to get it working for simple operations, but this doesn't scale.

### Just do it

Thus, I propose `;#`.  It transforms the code above into:

```
;#  ph-bind
  [%booted who=@p]      <--  (boot ~bud ~)
  [%boot-done whom=@p]  <--  (check-ship-booted ~bud)
  ~                     <--  ~&  [who=who whom=whom]
                             (return:(ph ,~) ~)
==
```

This is dramatically easier to read and follow what's happening.  

This works nearly exactly like Haskell's do notation.  The main difference is that we specify the type rather than just a name for each operation result.  This doesn't feel overly burdensome, and it is in keeping with Hoon's general principle of being more explicit about types.  One of the most frustrating parts of Haskell for me is how the types are very non-obvious but aren't specified in the code.

The implementation is pretty straightforward; it's just a translation of the one to the other.

### One more example

A complete example with a simpler monad:

```
=>  |%
    ++  unit-return
      |*  a=mold
      |=  arg=a
      ^-  (unit a)
      `a
    ::
    ++  unit-bind
      |*  a=mold
      |*  b=mold
      |=  [m-a=(unit a) fun=$-(a (unit b))]
      ^-  (unit b)
      ?~  m-a
        ~
      (fun u.m-a)
    --
|=  a=@
;#  unit-bind
  b=@     <--  `a
  c=@     <--  ?:(=(0 b) ~ `(div 120 b))
  d=tape  <--  ?:  =(c 0)
                 ~  ::  don't disrespect ~zod
               `(scow %p c)
  f=tape  <--  `"120 divided by {(scow %ud a)} is {d} in @p"
==
```

Put this in `/gen/unit.hoon` and try `+unit 5`, `+unit 121`, and `+unit 0`.

### Open questions

- What glyphs to give this rune.

- Introducing `<--` is questionable.  We could omit it, in which case it would have the same shape as `?-`, and could be formatted kingside or queenside accordingly.

- If you `=+` a non-monadic variable in the the body of an operation, later operations don't see it.  We could either not support that or include a specific syntax for it, like `===` instead of `<--`.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -6687,6 +6687,7 @@
   ::                                            ::::::  miscellaneous
     {$mcts p/marl:hoot}                                 ::  ;=  list templating
     {$mccl p/hoon q/(list hoon)}                        ::  ;:  binary to nary
+    {$mchx p/hoon q/(list (pair spec hoon))}            ::  ;:  do notation
     {$mcnt p/hoon}                                      ::  ;/  [%$ [%$ p ~] ~]
     {$mcsg p/hoon q/(list hoon)}                        ::  ;~  kleisli arrow
     {$mcmc p/hoon q/hoon}                               ::  ;;  normalize
@@ -8763,6 +8764,28 @@
           ~      !!
         ==
       ==
+    ::
+        {$mchx *}
+      ?~  q.gen
+        ~|("open-mchx" !!)
+      =/  last-spec
+        |-  ^-  spec
+        ?~  t.q.gen
+          p.i.q.gen
+        $(q.gen t.q.gen)
+      |-  ^-  hoon
+      ?~  t.q.gen
+        q.i.q.gen
+      :^    %cnls
+          :+  %cnhp
+            :+  %cnhp
+              p.gen
+            [%ktcl p.i.q.gen]
+          [%ktcl last-spec]
+        q.i.q.gen
+      :+  %brts
+        p.i.q.gen
+      $(q.gen t.q.gen)
     ::
         {$mcnt *}  =+(zoy=[%rock %ta %$] [%clsg [zoy [%clsg [zoy p.gen] ~]] ~])
         {$mcsg *}                                       ::                  ;~
@@ -14030,6 +14053,7 @@
         [%sgzp *]  (rune '~!' ~ ~ (hoons ~[p q]:x))
         [%mcts *]  %ast-node-mcts
         [%mccl *]  (rune ';:' `'==' `[':(' spc ')'] (hoons [p q]:x))
+        [%mchx *]  (rune ';#' `'==' `[';#(' spc ')'] (hn p.x) (matches q.x))
         [%mcnt *]  (rune ';/' ~ ~ (hoons ~[p]:x))
         [%mcsg *]  (rune ';~' `'==' ~ (hoons [p q]:x))
         [%mcmc *]  (rune ';;' ~ ~ (hoons ~[p q]:x))
@@ -16638,6 +16662,7 @@
               %-  stew
               ^.  stet  ^.  limo
               :~  [':' (rune col %mccl expi)]
+                  ['#' (rune hax %mchx exp0)]
                   ['/' (rune net %mcnt expa)]
                   ['~' (rune sig %mcsg expi)]
                   [';' (rune mic %mcmc expb)]
@@ -16832,6 +16857,7 @@
     ::
     ++  glop  ~+((glue mash))                           ::  separated by space
     ++  gunk  ~+((glue muck))                           ::  separated list
+    ++  gink  ~+((glue mick))                           ::  separated <-- list
     ++  butt  |*  zor/rule                              ::  closing == if tall
               ?:(tol ;~(sfix zor ;~(plug gap duz)) zor)
     ++  ulva  |*  zor/rule                              ::  closing -- and tall
@@ -16844,6 +16870,9 @@
     ++  lomp  ;~(plug sym (punt ;~(pfix tis wyde)))     ::  typeable name
     ++  mash  ?:(tol gap ;~(plug com ace))              ::  list separator
     ++  muck  ?:(tol gap ace)                           ::  general separator
+    ++  mick  ?:  tol                                   ::  <-- separator
+                ;~(plug gap gal hep hep gap)
+              ace
     ++  teak  %+  knee  *tiki  |.  ~+                   ::  wing or hoon
               =+  ^=  gub
                   |=  {a/term b/$%({%& p/wing} {%| p/hoon})}
@@ -16877,6 +16906,7 @@
               ==
     ++  rack  (most mash ;~(gunk loaf loaf))            ::  list [hoon hoon]
     ++  ruck  (most mash ;~(gunk loan loaf))            ::  list [spec hoon]
+    ++  rock  (most mash ;~(gink loan loaf))            ::  list [spec <-- hoon]
     ++  rick  (most mash ;~(gunk rope loaf))            ::  list [wing hoon]
     ::
     ::    hoon contents
@@ -16909,6 +16939,7 @@
     ++  expx  |.(;~(gunk loaf wisp))                    ::  hoon and core tail
     ++  expy  |.(;~(gunk ropa loaf loaf))               ::  wings and two hoons
     ++  expz  |.(loaf(bug &))                           ::  hoon with tracing
+    ++  exp0  |.((butt ;~(gunk loaf rock)))             ::  ;#
     ::    spec contents
     ::
     ++  exqa  |.(loan)                                  ::  one hoon


ohAitch:
Definite -1 on weird trigraph runes; what are the essential difficulties
with replicating all of this in explicit "ford" monad-call style? (In
particular ph-bind being unable to wetly extract the sample type from  fun
 seems unexpectedly verbose)

ohAitch:
I guess, more concrete syntax counterproposal, addressing both  
- introducing `<--` is questionable  
- a monolithic `;#` parser mode doesn't compose with other runes like `=+`  
while still resolving  
- just using explicit application syntax is too verbose  
- "type followed by definition" ordering is desirable  
is a flipped `%+ |=` rune ";|", such that: 

    ;|  [%booted who=@p]      an-bind  (boot ~bud ~)
    ;|  [%boot-done whom=@p]  an-bind  (check-ship-booted ~bud)
    ~&  [who=who whom=whom]
    (return:an ~)

expands to

    %+  an-bind  (boot ~bud ~)
    |=  [%booted who=@p]    
    %+  an-bind  (check-ship-booted ~bud)
    |=  [%boot-done whom=@p]
    ~&  [who=who whom=whom]
    (return:an ~)

and similarly

    ;|  b=@     biff  `a
    ;|  c=@     biff  ?:(=(0 b) ~ `(div 120 b))
    ;|  d=tape  biff  ?:  =(c 0)
                        ~  ::  don't disrespect ~zod
                      `(scow %p c)
    `"120 divided by {(scow %ud a)} is {d} in @p"

denotes

    %+  biff  `a
    |=  b=@
    %+  biff  ?:(=(0 b) ~ `(div 120 b))
    |=  c=@
    %+ biff  ?:  =(c 0)
               ~  ::  don't disrespect ~zod
             `(scow %p c)
    |=  d=tape
    `"120 divided by {(scow %ud a)} is {d} in @p"

This does drop the "ph-bind gets passed in argument and result molds explicitly" design, though the argument type could obviously just be duplicated at the cost of a more complex expansion. What are the ways in which that gives clearer errors than effectively `|*([a=* b=gate] (b a))` would?

(Similarly this allows sprinkling `^-  (an-data ,~)` cast/s where appropriate)

philipcmonk:
That's an interesting proposal.  Aesthetically I like putting bind in the middle like that -- it communicates that the bind function is what's making the assignment.  I like that it composes better with `=+`.  I have two concerns, though:

- Repeating `biff` like that suggests that you could interleaves other "bind"s, which is not at all the case.  It's generally an antipattern to require the same code to be repeated multiple times when it can't reasonably be any other.

- I don't trust wet gates any further than I can throw them.  This may be an irrational fear, but it's certainly a common one.  I can generally explain why they do what they do, but there's enough annoying times when you have to add explicit casts to silence misleading error messages that they feel like the opposite of composable.  Maybe it's possible to use wet "bind"s with specific patterns such that everything just works, but I'd have to see it in action.

ohAitch:
- can it not be? I may be misunderstanding how monads work, but having interleaved functions `M a -> N a` and `N a -> M a` seems not completely out of the question
- that's reasonable, but given you are using wet gates either way, having a thicker layer above them seems likely to just give you different misleading error messages

ohAitch:
In particular, "what rune could be useful for calls following the `;~` type signature and only that type signature" is probably not the goal here - consider

```
;|  a=elem  murn  (foo bar)
;|  b=sthg  biff  (baz a)
;|  c=char  bind  (bam b)
?:((lth c ' ') 'x' c)
```

"bind" and "return" being privileged is mostly useful for higher-order monad composition, the result of which could be used as a `;|` composer just fine. Or in the case of do notation, `bind` is a "biblical" name referenced by `<-`: note that `;#` as described is a `;|` on every clause except the last one, and a hoisted `=/  <-- |*(a=mold ((ph-bind a) ,~)` for head/last-type subtwigs `ph-bind`/`,~`.

belisarius222:
@philipcmonk For an example of fully wet ("drenched", vs. "damp", perhaps) bind chaining, see +cope in old Ford:
https://github.com/urbit/arvo/blob/beaa8ec4c5d1bd9144d54c6fffc1be7ccfe0836f/sys/vane/ford.hoon

It is possible -- or at least it was with the type system at that time. @ohAitch's proposal would work better for that structure, since the sequencer frequently varied between links in the call chain. I also think keeping locally defined variables in scope later down the chain is very important.

I like repeating a single rune and specifying the bind each time. It feels both more flexible and lighter weight. I think it's also a simpler desugaring, which means my mind's compiler doesn't have to work as hard.

If we don't go that route, though, I think we could regain scope propagation by requiring the <-- rune. That way normal hoon in between <--s shoild be able to mutate the subject passed to the next gate definition.

I feel like this should be a % rune, especially if we go with something closer to @ohAitch's version, since it's basically just a special way to call a function that's implicitly defined by the subexpressions. `%|`, `%;` and `%/` come
to mind.

Although I guess if we use a ; rune we could claim we've literally made a programmable semicolon ...

ohAitch:
Ah, have the monads been removed from ford? That is what I was referring to
yes, surprised to find out the style is gone: it (and its annoyances) is
definitely what I was modeling the rune definition after.

%| was indeed my first thought, but that's already cubical .n, and "a call
that also defines a gate" seems like an amount of complexity we should
contain to ; runes

On the style side, both a one-off trigraph and an infix rune in tall form
have no precedent, and consistency is another way to prevent people's mind
compilers from working too hard :3

philipcmonk:
The `<--` trigraph is syntactically unnecessary -- it could be omitted entirely.

If we decide to go with Anton's version, then I suggest `=%`.  It has the form of a variable binding, like `=+` or `=/`.  It can be thought of as "`=+` except helped along by a function" (and `%` invokes ideas of functions).

> can it not be? I may be misunderstanding how monads work, but having interleaved functions `M a -> N a` and `N a -> M a` seems not completely out of the question

The functions would be `a -> M a` and `a -> N a`, and no, I don't think you can generally interchange them.  That would require your "bind" function to have type `M a -> (a -> N b) -> N b`.  You could certainly write such a function, but should you?

Your example of `murn`/`biff` is interesting.  I suppose `murn` has type "M a -> (a -> N b) -> M b".  With a type like that, I'd give up on trying to interpret it as anything particularly monadic.  However, your rune is simple enough to mind-decompile that you don't really have to think of it as a monadic rune.

> moist gates

I don't think of `|*  a=mold  |=` as a true "wet gate", but rather as a sort of type-parameterized dry gate.  Your `murn`/`biff` example is a good example.  Can you quickly tell which explicit casts must be added to make this work?

```
%+  murn  ~[1 2 3 4 5 13]
|=  a=@
%+  biff  ?:(=(0 (mod a 2)) ~ `(scow %ux a))
|=  b=tape
?.  ?=([%'0' %x *] b)
  ~
`t.t.b
```

Or, in Anton's version:

```
=%  a=@     murn  ~[1 2 3 4 5 13]
=%  b=tape  biff  ?:(=(0 (mod a 2)) ~ `(scow @ux a))
?.  ?=([%'0' %x *] b)
  ~
`t.t.b
```

Turns out you need two:

```
=%  a=@     murn  ^-  (list @)
                  ~[1 2 3 4 5 13]
^-  (unit tape)
=%  b=tape  biff  ?:(=(0 (mod a 2)) ~ `(mul a a))
?.  ?=([%'0' %x *] b)
  ~
`t.t.b
```

If your last function happens to return a well-faced unit, then you don't need the second one.  But the point is that's a requirement that is exclusive to wet gates.  If you pass a function into a dry gate, you never have to make sure you produce stuff with the right faces on it.

Similarly, you don't need to cast the argument to a dry gate -- that's something I should never have to think about.

Now, there's two ways to deal with this.  One is to use parameterized dry gates, as in my original proposal.  The advantage is that everything is fairly straightforward, and you never need to explicitly put faces on things.

The other is to make sure your "bind" functions explicitly extract the type.  We don't talk much about the "hypothetical assertions to extract types" programming style, but I suppose it isn't that complicated.  I haven't used it enough to know if there are gotchas.  Many functions in the standard library don't use this style, but I think you could make them do it.  For example, you can change this:

```
++  murn
  |*  {a/(list) b/$-(* (unit))}
  |-
  ...
```

Into:

```
++  murn-t
  |*  {a/(list) b/$-(* (unit))}
  =>  .(a (homo a))
  =>  .(b `$-(_?>(?=(^ a) i.a) (unit _?>(?=(^ $:b) +:$:b)))`b)
  |-  ^-  (list _=+(o=$:b ?>(?=(^ o) u.o)))
  ...
```

This is cryptic; arguably clearer, this is equivalent:

```
++  murn-t-2
  |*  {a/(list) b/$-(* (unit))}
  =/  i-type
    |.
    ?>  ?=(^ a)
    -.a
  =/  o-type
    |.
    =/  o  $:b
    ?>  ?=(^ b)
    +.o
  =>  .(a `(list _*i-type)`a)
  =>  .(b `$-(_*i-type (unit _*o-type))`b)
  |-  ^-  (list _*o-type)
  ...
```

Then, you can do:

```
=%  a=@     murn-t  ~[1 2 3 4 5 13]
=%  b=tape  biff    ?:(=(0 (mod a 2)) ~ `(scow @ux a))
?.  ?=([%'0' %x *] b)
  ~
`t.t.b
```

With no problem.  It even produces a real list this time, unlike the previous examples.  In fact, this gives you better error messages because it'll nest-fail on exactly the argument you're trying to cast.  If the function is not defined within a `!:` (like the standard library)`, you can just add `!:` and `!.` immediately around the `=>` lines.

If this style works as consistently as it seems to, and you make all "bind"s use this style, then Anton's proposal has the following advantages relative to mine:

- Since it's done one step at a time, it feels more hoon-like.  The repeated form is clean because of backstep.

- Doesn't have to use either an awkward disconnected syntax or a spurious arrow.

- Can use heterogeneous types, which adds flexibility.

- Is simpler to mind-compile.

And the following disadvantages:

- "bind" is a little harder to write because you have to use this backwards "extract the type" programming style.  However, we should generally prefer this sort of magic in the definition of "bind" rather than the call site, so it's not too bad.

- Can use heterogeneous types, which might encourage inscrutable programming.  Perhaps we should only support more strict "bind" chaining.

- Uses wet gates, which are "scary".  Unless someone can identify a more specific criticism, we should just ignore this.

I guess my open question is whether the "where's waldo" programming style easy to extend to more complex monads?  If it simply follows the rule of "assert whatever you need to find an example of the type, as long as you don't ever evaluate it (i.e. do it in a `$_` or `|.`), then it should be doable.  I'll try to use that style for my ph monads and report back whether I run into any gotchas.

If that works, then I think you've convinced me, Anton.

ohAitch:
I had been vaguely hoping for =% to be a ford rune for locally scoped
changes to % i.e. cd; but I've been hoping that for a while and it
certainly isn't the worst of options.

> can it not be? I may be misunderstanding how monads work, but having
> interleaved functions M a -> N a and N a -> M a seems not completely out
> of the question
>
> The functions would be a -> M a and a -> N a, and no, I don't think you
> can generally interchange them. That would require your "bind" function to
> have type M a -> (a -> N b) -> N b. You could certainly write such a
> function, but should you?
>
Right, not generally, when I said "interleaved" I was thinking of specific
cases like

=% =x turn-and-zing input
%-  drop
=% =y biff (some (unit (functions x)))
%-  prod :: |*(a=(list) `(unit _a)`?~(a ~ (some a)))
=% more list stuff
etc

More abstractly, a -> M a wrapped by a -> N M a might ever come up?

> Your example of murn/biff is interesting. I suppose murn has type "M a ->
> (a -> N b) -> M b". With a type like that, I'd give up on trying to
> interpret it as anything particularly monadic. However, your rune is simple
> enough to mind-decompile that you don't really have to think of it as a
> monadic rune.
>
> moist gates
>
> I don't think of |* a=mold |= as a true "wet gate", but rather as a sort
> of type-parameterized dry gate. Your murn/biff example is a good example.
> Can you quickly tell which explicit casts must be added to make this work?
>
> %+  murn  ~[1 2 3 4 5 13]
> |=  a=@
> %+  biff  ?:(=(0 (mod a 2)) ~ `(scow %ux a))
> |=  b=tape
> ?.  ?=([%'0' %x *] b)
>   ~
> `t.t.b
>
> Or, in Anton's version:
>
> =%  a=@     murn  ~[1 2 3 4 5 13]
> =%  b=tape  biff  ?:(=(0 (mod a 2)) ~ `(scow @ux a))
> ?.  ?=([%'0' %x *] b)
>   ~
> `t.t.b
>
> Turns out you need two:
>
> =%  a=@     murn  ^-  (list @)
>                   ~[1 2 3 4 5 13]
> ^-  (unit tape)
> =%  b=tape  biff  ?:(=(0 (mod a 2)) ~ `(mul a a))
> ?.  ?=([%'0' %x *] b)
>   ~
> `t.t.b
>
> If your last function happens to return a well-faced unit, then you don't
> need the second one. But the point is that's a requirement that is
> exclusive to wet gates. If you pass a function into a dry gate, you never
> have to make sure you produce stuff with the right faces on it.
>
Huh, I thought the recent type system changes have face unification to
solve this particular problem? Agree that murn should do ?-(res ~ ~, [~
u=@] (fun u.res)) instead of just ?~ for various reasons (hopefully there's
not extensive list usage?)

I will note the canonical list builder ++ly, which would not force a weird
overhand line — but also my understanding of current ph-bind is you
/always/ have a (list @) as formal argument somewhere? Which is not in my
book "never have to think about casting the argument", you're just hiding
the ^-

> For example, you can change this:
>
> ++  murn
>   |*  {a/(list) b/$-(* (unit))}
>   |-
>   ...
>
> Into:
>
> ++  murn-t
>   |*  {a/(list) b/$-(* (unit))}
>   =>  .(a (homo a))
>   =>  .(b `$-(_?>(?=(^ a) i.a) (unit _?>(?=(^ $:b) +:$:b)))`b)
>   |-  ^-  (list _=+(o=$:b ?>(?=(^ o) u.o)))
>   ...
>
> I remember some discussion relatively recently about a way to say => .(a
`foo`a) less awkwardly! (iirc just extending =. to accept =. a=foo a, when
that nests)

What are the cases in which the second cast is necessary? The a cast covers
input literals, the output ^- (list _(need (b)) or rather ^- (list _(need
(b (snag *@ a)))) is helpful so the gate is "moist" instead of wet, but the
function itself is taking up by far the most space despite already being
type checked when you call it.

(Also, the ph-bind equivalent targeted at only the rune in this PR would
probably be => .(a `(list ,.+<.b)` a), which is more dry; you might want
the option of passing a wet gate in directly however, in which case its
output type depends on the correct input type)

> It even produces a real list this time, unlike the previous examples. In
> fact, this gives you better error messages because it'll nest-fail on
> exactly the argument you're trying to cast.
>
A lot of my motivation here is "passed in a gate that doesn't match the
data" errors are going to be a lot harder to decode than "passed in a datum
that doesn't match the gate" ones. Especially if the type printer has been
cleaned up enough that the -need. -have. mismatch diagnostics have been
turned back on, don't suppose that's the case?

>  [advantages]
>
>    -
>
>    Is simpler to mind-compile.
>
> This is the most important point imo: more than just /mind/ compile, it
lets you manually expand a single line in isolation, stick ~!s into places
that are doing weird things, and generally poke at the offending call while
leaving the rest of the chain intact. (Or ~|s and ~&s, but I don't think
I've seen a criticism of wet gates that _once they compile_ they do
particularly unexpected things, the whole point of the design is limited
polymorphism while allowing dry-equivalent consistent runtime behavior)

> And the following disadvantages:
>
>    -
>
>    "bind" is a little harder to write because you have to use this
>    backwards "extract the type" programming style. However, we should
>    generally prefer this sort of magic in the definition of "bind" rather than
>    the call site, so it's not too bad.
>
> Yup, fundamentally you should not be defining many monads! Which comes
back to the Curtis Maxim "higher-order functional programming is DSL
design; DSL design should be a lot less frequent and more thoroughly
documented/debugged than the rest of your code base, Here There Be Dragons"

>
>    -
>
>    Can use heterogeneous types, which might encourage inscrutable
>    programming. Perhaps we should only support more strict "bind" chaining.
>
> Heterogeneous type types? I think the old ford code shows the ways in
which this is useful, though it only had 3 or 4 and they followed the same
general structure; agree that this whole PR adds more complexity to what
can be ergonomically expressed at any rate, which is not an unalloyed
positive.

>
>    -
>
>    Uses wet gates, which are "scary". Unless someone can identify a more
>    specific criticism, we should just ignore this.
>
> You had a more specific criticism, inscrutable and misleading error
messages :P

I just think that the source of these is doing things with monads in the
first place: monads exist to let you locally ignore more complicated data
types, which comes at the cost of needing to gear-switch back to the
complicated types if you get an error on that layer. As discussed, the
solution is a mix of more explicit checking in wet gates, and generally
improving type debugging across the system. (if nest need-have is still
disabled for being too verbose, it's the obvious thing to fix, but I'm sure
there are other things)

> I guess my open question is whether the "where's waldo" programming style
> easy to extend to more complex monads? If it simply follows the rule of
> "assert whatever you need to find an example of the type, as long as you
> don't ever evaluate it (i.e. do it in a $_ or |.), then it should be
> doable. I'll try to use that style for my ph monads and report back whether
> I run into any gotchas.
>
 Probably do it in a ^- primarily? Looking forward to hearing how this
works out!

> If that works, then I think you've convinced me, Anton.
>
^-^

ohAitch:
…that email formatting didn't come through great, apologies. I think it is fairly straightforward to follow, some of my replies are just incorrectly quoted; though I did actually bungle one bit:

> Agree that murn should do `?-(res ~ ~, [~ u=@] (fun u.res))` instead of just ?~ for various reasons (hopefully there's not extensive list usage?)

This is a mid-edit version that mashes together several points:
1. _biff_ should fish on the input, so that if it's not actually a unit this gives a non-cryptic "mint-lost" error. (This might not be an issue in practice, esp if ``%.(`foo |*(a=(unit) ... u.a))`` unifies the face in the way I recall being added).
2. biff should also check that the output is a unit: while I think as currently written ``(biff foo |=(a=foo-u-type `(list bar)`...)`` will \~work\~, that probably shouldn't actually compile
3. murn should do "if it's unit shaped put a unit face on it" type regularization so that ``(murn "abc" |=(char `baz))`` doesn't unexpectedly fail, as covered in your post

philipcmonk:
Okay, I'm back to hating wet gates.  The easiest way to hate wet gates is to read them, write them. or use them.

I spent some time trying to well-type `++murn`.  You're correct the `=>  .(b ...` line is bad, because wet gates don't nest in anything but themselves.  Here's the best version of murn I've come up with:

```
++  extract-type-from-list
  |*  a=(list)
  ?>  ?=(^ a)
  ,.-.a
::
++  extract-type-from-unit
  |*  a=(unit)
  ?>  ?=(^ a)
  ,.+.a
::
++  murn-wet
  ~/  %murn
  |*  {a/(list) b/$-(* (unit))}
  !:
  =/  i-type
    |.  (extract-type-from-list a)
  =/  o-type
    |.  (extract-type-from-unit (b $:i-type))
  =>  .(a `(list _*i-type)`a)
  !.
  |-  ^-  (list _*o-type)
  ?~  a  ~
  !:
  =+  c=`(unit _*o-type)`(b i.a)
  !.
  ?~  c
    $(a t.a)
  [i=u.c t=$(a t.a)]
```

This solves a lot of problems, but I can't for the life of me get this to compile:

```
(murn-wet ~[`1 ~ `3] same)
```

Compare an explicitly-parameterized version:

```
++  murn-param
  |*  [i=mold o=mold]
  |=  [a=(list i) b=$-(i (unit o))]
  |-  ^-  (list o)
  ?~  a  ~
  =+  c=(b i.a)
  ?~  c
    $(a t.a)
  [i=u.c t=$(a t.a)]
```

Called as:

```
((murn-param (unit ,@ud) ,@ud) ~[`1 ~ `3] |=((unit @) +<))
```

Which runs just fine.

My point is that the first version took hours to write, is harder to read, and still requires explicit casts.  The second version took less than 5 minutes to write, is much shorter, is much easier to read, and works exactly like you would expect any dry gate to work.

You might argue that there's explicit casts in the second, they're just written as arguments instead of casts.  That is, of course, why it works -- there's no free lunch.  I contend:

(1) Not knowing when you might need to put in a cast is far worse than having to write a few extra types.  The former easily costs hours of debugging while the latter costs a couple seconds of typing at the moment you write the code -- which is exactly when you know the answer.

(2) Boilerplate can be reduced significantly with a rune such as `;#`, so that call-time looks significantly better.

(3) The reason explicit casts are the worst is that I generally think that if we have a function that accepts `a`, and `b` nests within `a`, then I can give `b` to the function.  This is just bog-standard contravariant subtyping.  *Wet gates break this!*  I have to find another type `c` where `b` nests within `c` nests within `a` and give the compiler that explicit nesting path.  It's hard to overstate how much this makes me feel like I'm programming in a hall of mirrors.

The worst part `++murn-param` is that it doesn't accept wet gates, because again wet gates don't nest in anything.

I think the answer is to "just say no" to wet gates.  Fortunately, exit is easy:

```
::  Leave the terrors of the marine world behind, come back to the light!
::
++  wring
  |*  [fun=gate arg=mold]
  |=  a=arg
  (fun a)
```

If you want, you can use this to write `++murn-param` in terms of `++murn-wet`.

```
++  murn-wrung
  |*  [i=mold o=mold]
  (wring murn-wet ,[(list i) $-(i (unit o))])
```

Which works as you would expect any dry gate to work.

```
((murn-wrung (unit ,@ud) ,@ud) ~[`1 ~ `3] (wring same (unit ,@ud)))
```

Magic is only good if it works perfectly.  Flawed magic is far worse than no magic.

philipcmonk:
> but also my understanding of current ph-bind is you /always/ have a (list @) as formal argument somewhere?

I'm not sure how you think ph-bind works, but here's the data type for it.  `(ph-data a)` is "M a".

```
++  ph-data
  |*  a=mold
  $-(ph-input (ph-output a))
::
+$  ph-input
  [who=ship uf=unix-effect]
::
++  ph-output
  |*  a=mold
  $~  [& ~ %done & *a]
  $:  thru=?
      events=(list ph-event)
      $=  next
      $%  [%wait ~]
          [%cont self=ph-data]
          [%done success=? value=a]
      ==
  ==
```

> Yup, fundamentally you should not be defining many monads! Which comes back to the Curtis Maxim "higher-order functional programming is DSL design; DSL design should be a lot less frequent and more thoroughly documented/debugged than the rest of your code base, Here There Be Dragons"

Unfortunately, decomposed state machines are probably the worst part about programming in arvo, and they're *everywhere*.  We desperately need to be able to write IO in a linear fashion without spraying code all over a vane/app.  This is a major source of bugs and limiter as to how big of stuff you can reasonably build on Urbit.  Properly designed monads can let us write state machines linearly, and they're composable.  Composable state machines will make a *huge* difference. 

> Especially if the type printer has been cleaned up enough that the -need. -have. mismatch diagnostics have been turned back on, don't suppose that's the case?

They're not by default.  PSA: you can turn them on trivially by grepping for "have" in urbit/jets/ and uncommenting the need/have lines.

ohAitch:
Can confirm, you cannot cast a wet gate to a dry gate, and specializing one actually requires explicitly calling it. This could maybe be solved in the compiler, I think there were soundness concerns?

What happens with ``(murn-wet ~[`1 ~ `3] |=((unit @) +<)``? I still don't think I'm following the chain of reasoning
- if you put no casts anywhere sometimes wet gates break
- if you put casts everywhere and wrap every wet gate with a dry gate things generally work
- it is possible to define wet gate APIs in a way that forces you to explicitly specify casts and wrappers everywhere
- if you are not forced to do so, you will be unable to remember that you can solve "too-wet" code issues by adding casts and wrappers??

ohAitch:
Ah, so `ph-data` is not actually _data_, but a gate? The naive IO monad seems difficult to do without effectively-permanently leaking historical versions of the code / breaking the ability to change steps in the state machine(after leaving the first state); but ford did a similar thing with wires and build step specifications, which presumably should have run into the same problems but didn't.

Or I guess did if that's the "explicit state-tracking code all over the vane" objection?

philipcmonk:
> What happens with `(murn-wet ~[`1 ~ `3] |=((unit @) +<))`?

Doesn't work.  It doesn't get around to touching the gate, it chokes on the list.  Even `+homo` crashes on that list.

> - if you put no casts anywhere sometimes wet gates break

Often

> - if you put casts everywhere and wrap every wet gate with a dry gate things generally work

Don't need casts if you parameterize (ie `+wring`) every wet gate.

> - it is possible to define wet gate APIs in a way that forces you to explicitly specify casts and wrappers everywhere

It's many hours of work to define a wet gate API that *doesn't* force you to do so.  I don't believe it's always possible

> - if you are not forced to do so, you will be unable to remember that you can solve "too-wet" code issues by adding casts and wrappers??

They can't all be solved in the wet gates themselves, which means they must be solved at call time.  And a world without Liskov substitution isn't worth living in.

ohAitch:
I would like to draw a distinction between between "force you to", i.e. `murn-wrung` which is impossible to call without explicitly specifying the type parameters of the wrapper it generates for you, and "requires in some edge cases", i.e. vanilla `murn` where you can apply explicit types if the derived types don't line up. (And has no casts whatsoever, relying on forward inference / caller wrapping where appropriate)

I guess I just disagree that ``(murn `(list (unit @ud))`~[`1 ~ `3 ~]) same)`` is more onerous than ``((murn-param (unit ,@ud) ,@ud) ~[`1 ~ `3] |=((unit @) +<))``? In particular, the main consistency "benefit" from the latter seems to be saying `(unit @ud)` instead of `(list (unit @ud))`, leaving the fact that ``~[`1 ~ `3]`` is a list implicit in `murn-wrung` being a list function. (I'm also not sure where liskov is being violated: by all means, something like `wring` should be in the standard library, and makes sense to use when you have a `|*` and need a `$~`)

And to reiterate,
> It's generally an antipattern to require the same code to be repeated multiple times when it can't reasonably be any other.

philipcmonk:
``~[`1 ~ `3]`` is obviously a list of units.  More precisely, I'm using "is" in the usual sense of "has a type that nests within list of units".  `+murn` is supposed to take a list of `a` and a gate from `a` to `(unit b)` and produce a list of `b`.  Liskov says if a value "is" of a type, then you can use it for anything that accepts that type.  I gave it a list of `a=(unit)` and it *broke*.  The compiler is gaslighting me by *pretending* to take a list of `a` and doing no such thing.

I agree that the parameterized version is longer and subjectively looks worse in its current form.  But at least it doesn't violate Liskov.  To me, that matters far more.

ohAitch:
I do not recall ever spending hours to make the types of a wet construction line up, and I'm trying to figure out why - I think the conservative version of my "sometimes add casts and wrappers" heuristic is something like "don't put wet types into wet types", i.e.
- have a cast on any wet result being consumed by a wet gate
- and a wrapper on any composed gate being itself passed to a gate composer. 

This is still half as much repetition as "explicitly specify the input and output type of every higher-order gate, and redundantly wring any wet gate you put into it": most times you `turn` this is assigned to an explicitly-typed variable / returned through a `^-` / passed to a dry gate, and most times you `cork` this is immediately invoked on a concretely typed variable and not used as a `turn` body.

You also mention that it took hours to _write `murn-wet`_, and the result is hard to read and still requires casts; while this sounds frustrating, I think it is mostly a point in favor of hoon.hoon `murn`, which also takes 5 minutes to write(I think it's one line different from `murn-param`) and works fine if you specify input types.

ohAitch:
But murn doesn't take a `(list a)` and a gate `$~(a (unit b))`! I agree that if hoon actually claimed to have type variable unification, this would be a serious transgression; but the only thing the wet gate claims is "I take `a=(list *)` and `b=$-(* (unit))`, or things sufficiently similar"

I think the main disagreement here is whether "in an explicit cast odors and faces are adjusted for if there's structural equivalence" means "structural equivalence is hoon's definition of Liskov subtyping": the answer appears to be "it depends", which is obviously more annoying than "yes" or "no". I don't think the view "`murn` accepts a `(list)`, i.e. conservatively, the output of a `^-  (list sth)` expression" is /inconsistent/ however; that's certainly how things work in C or Go or w/e. (A possible view on this: things have real concrete types, which frequently come from type constructors; most hoon runes have a limited form of duck typing, and wet gate calls have a different, more powerful but incompatible, form)

philipcmonk:
> I think the main disagreement here is whether "in an explicit cast odors and faces are adjusted for if there's structural equivalence" means "structural equivalence is hoon's definition of Liskov subtyping": the answer appears to be "it depends", which is obviously more annoying than "yes" or "no".

Hoon's definition of Liskov subtyping is nest, not structural equivalence.  This works perfectly for dry gates.  Under nest, ``~[`1 2 `3]`` is a `(list *)`, but neither `+murn` nor `+homo` can handle it.  Nest is easy for people to think about, it feels very natural.

Whatever Liskov subtyping applies for wet gates is poorly defined and not very useful.  It's something like "does the same thing if you run the same code on it" (as opposed to normal duck typing, which is "does something if you run the code on it"), which is extremely unwieldy and non-composable.  It's useful for a few small things and counterproductive in everything else.

ohAitch:
Super confused about `homo` vs `ly` now actually: it looks like `homo` doesn't like the list, `ly` i.e `le:nl` does, and they're defined identically?

Normal duck typing is arguably "does something if you run the code on it, that something might be an error": see NPEs like `undefined is not a function` :P (But the way I would define wet typing in practice is "has the same faces in the same places", where "spaces" encompasses fork probability-space and axes within the specific cases; there exist wet gate samples that contain faces not mentioned in the gate itself, but they're not like, common/desirable)

philipcmonk:
`le:nl` has the same text as `homo`, but `snag` refers to `snag:nl`, which has slightly more type info than the regular one.

jtobin:
This generally looks great.  A major problem with using monads in (seemingly) every non-Haskell language is that there is rarely decent syntactic support for them.  Both of the broad schemes suggested above appear to be fairly pleasant; either way we could get this into the language feels like a win.

I think I'm agreed that @ohAitch's proposed style (i.e., repeating the single rune on binds) might be a little nicer, and is probably more idiomatic Hoon as-is.  Using a tis-prefixed rune also seems to make sense, and is also in the flavour of the existing/similar `=^` rune.

ohAitch:
Gah, I can see that `^+` breaking basically _all type-derivation uses of snag_, who added that? (I guess more directly, who was writing `++ly`, noticed the bug, and decided the solution was to **shadow a standard library function** with a slightly different version, instead of fixing it or even coming up with a new name for "get homogenized element type")

This looks like `homo` is basically ``|*(a=(list) `(list _,.-.a)`a)`` now, which is super not what it's supposed to do; and if it is, it should not call out to snag as a false affordance. (I still have a wider disagreement about it being gaslighty of the compiler to fail to pull together `(unit @)` when presented with a list of `~` and `[~ @]`, but I agree it would be \~helpful\~ for the standard library to do so manually)

ohAitch:
On further review, "all other type-derivation uses of snag" still in the repo are two lines both of which look like my code, so possibly this is on me not knowing this particular pitfall / except the usage in homo which is just a bug.

I do think "by `x.a`, do you mean `p.a`, as in the type of the head of the tuple, or `i.a`, as in the type of any &1 &2 &3… axis in the list" is a problem any polymorphic code has to address, and the compiler can only give you a reasonable default(currently the former which seems wise). Monomorphic i.e. dry code unsurprisingly doesn't run into it, because you already tested that the one type it will ever run on works; forcing monomorphization by requiring 3-5x the number of explicitly wrung type parameters is tedious and hard to read, a Haskell or Rust like `[a=(list ?a)) b=$-(?a (unit ?b))]` constraint system would be interesting but hardly "less magic". The benefit of forward-inference wet gates is you can just swap in the sample type and read down the code following what it does, same as a dry gate. (For both, diagnostic tools to e.g. annotate each line with all types introduced/changed in that line certainly wouldn't hurt - to be usable this might require `!.` only affecting generated nock so that the compiler _knows_ what line it's on)

ohAitch:
A note on

>  A major problem with using monads in (seemingly) every non-Haskell
language is that there is rarely decent syntactic support for them.

*using* them certainly has various syntactic support: the ? operator,
throw/catch, await, yield…

It's adding new ones that's the trouble :)

philipcmonk:
For those just tuning in, here's a summary of the discussion so far:

I propose a rune that looks like this:

```
;#  ph-bind
  [%booted who=@p]      <--  (boot ~bud ~)
  [%boot-done whom=@p]  <--  (check-ship-booted ~bud)
  ~                     <--  ~&  [who=who whom=whom]
                             (return:(ph ,~) ~)
==
```

Anton proposes a rune that looks like this:

```
;|  [%booted who=@p]      an-bind  (boot ~bud ~)
;|  [%boot-done whom=@p]  an-bind  (check-ship-booted ~bud)
~&  [who=who whom=whom]
(return:an ~)
```

Anton's syntax is better.  The problem is that it only works if your bind is wet.  I think wet gates are almost always a mistake, and certainly so for bind.  Wet gates violate Liskov substitution.  Anton claims that adding explicit casts is sufficient.  I claim that explicit casts just to paper over the fact that Liskov substitution is broken is a very bad programming experience.

Unfortunately, Anton's rune syntax doesn't work with parameterized binds, only wet ones.  Or rather, you can explicitly monomorhpize your binds, but it's repetitive.

So, I stand by my original proposal and paramterized dry gates as superior to wet gates.

ohAitch:
Oh sorry thought I mentioned this, you can totally still parametrize your binds!

```
=/  use  (ph-bind ~)
;|  [%booted who=@p]      use  (boot ~bud ~)
;|  [%boot-done whom=@p]  use  (check-ship-booted ~bud)
~&  [who=who whom=whom]
(return:an ~)
```

This is still repetitive wrt `(return ~)` and `(ph-bind ~)`, but like, that's the general argument against parametrized wet gates. It also moves them apart from each other, which is probably fine for a synthetic `^-`; it's definitely easier to follow the control flow, than in the second OP example where `(ph-bind ... f=tape)` is split by 7 lines.

(you can even define `;|` with "your" expansion of
```
=/(sam-t ,{p.gen} %+(%-({q.gen} sam-t) {r.gen} |=(sam-t {s.gen})))
```
instead of just `%+({q.gen} {r.gen} |=({p.gen} {s.gen}))` to regain sample parametricity, which is compatible-ish with normal wet gates by the expedient of `;|(a=@ _biff (slaw %b c) ...)`. I was just arguing this was not worth the complexity - esp after you add the `=> v=.` hygiene stuff)

cgyarvin:
Wet gates are best thought of as Liskov substitution on the code, not the data.

I also don’t like the trigraphs.  But I agree that this general direction is good and useful.

philipcmonk:
> Oh sorry thought I mentioned this, you can totally still parametrize your binds!

Ah, I see.  Since the `b` argument to bind is the same for all of them, you can swap the order of bind's type parameters so that it has signature

```
|*  b=mold
|*  a=mold
|=  [m-a=(data a) fun=$-(a (data b))]
^-  (data b)
!!
```

I like that solution, I think it's optimal for parametrized binds, and if you really want to use wet binds, you can just add `_`.  It'll just serve as a marker to remind you that you'll need to be careful about possibly adding explicit casts.

> is compatible-ish with normal wet gates by the expedient of `;|(a=@ _biff (slaw %b c) ...)`

Using `_a` as shorthand for "create a function that throws away its argument and just produces `a` seems like a highly suspect use of the mold system, but I dig it.

> esp after you add the `=> v=.` hygiene stuff

I don't think you need that for this expansion?

> Wet gates are best thought of as Liskov substitution on the code, not the data.

I guess subtyping on code is just equivalence, but I'm not sure I can come up with a statement of Liskov that makes any sense for that.

ohAitch:
> Using `_a` as shorthand for "create a function that throws away its argument and just produces `a` seems like a highly suspect use of the mold system, but I dig it.

Is ``,_`.`` still frequently used as an app reset invocation? :P



> > esp after you add the `=> v=.` hygiene stuff
> 
> I don't think you need that for this expansion?
> 

You do if you don't want the `sam-t` name leaking into the `s.gen` subject! (I guess you could just compute the mold twice, but that seems lame, `;|  a=((slog >%ran< ~) @) b c d` shouldn't spuriously duplicate the printf)

philipcmonk:
> Is ``,_`.`` still frequently used as an app reset invocation? :P

Ha, that's a blast from the past.  I haven't seen it in a long time.

> You do if you don't want the sam-t name leaking into the s.gen subject! (I guess you could just compute the mold twice, but that seems lame, `;| a=((slog >%ran< ~) @) b c d` shouldn't spuriously duplicate the printf)

I was thinking to compute the mold twice, but you're right of course.  I'll implement and see how it feels with my ph monad.  If that goes well, I'll PR it.

ohAitch:
I think it popped up on @fang-'s livestream? Agree this is not a _common_ trick, esp in production code, at any rate.

philipcmonk:
As far as I can tell, even the `v=` trick doesn't stop the printf from happening.  For example:

```
=/  m  ,a=((slog >%run< ~) @)  [*m *m]
```

Outputs

```
%run
%run
%run
%run
0
```

It appears that two of these happen at compile time and two at run time.

If I do the naive expansion

```
%+  (q.gen p.gen)
  r.gen
|=  p.gen
s.gen
```

Then it prints twice at compile time and once at run time, same as if I do

```
=>  v=.
=+  sam==>(v ,p.gen)
%+  (=>(v q.gen) sam)
  =>(v r.gen)
|=  sam
=>  .(+> v)
s.gen
```

I guess I'll just do the naive way?

cgyarvin:
The question +mull is asking: consider a dry version of this core, in which it was actually compiled to the type currently in the sample.

At runtime, we don’t have the battery for that core.  We do have the generic battery.  Can we substitute our generic battery for the “real” battery we should be using?

That’s what I mean by Liskov substitution on the code — you are literally describing the effective equivalence of Nock batteries.

philipcmonk:
Ah, so we can say, given a wet core `f` with formal sample `a`, calculate the battery `f{a}`.  Given actual sample `b`, calculate the battery `f{a}`.  Then, if `f{a}` is a subtype of `f{b}`, we can substitute `f{a}` for `f{b}`, so the type checking succeeds.  For a battery to be a subtype of another battery means that if you pull an arm from the subtype it gives the same thing as if you pulled the arm from the supertype.  This is clearly true if the batteries are equal; it is hard to prove otherwise, and I'm not sure we try.

This explains when you can use a wet gate, but it suggests that wet gates are sufficiently different from dry gates that they probably shouldn't use the same calling syntax.  A dry gate is pretty much an earth function.  A wet gate definitely isn't.

In particular, we can say "given a dry gate with formal sample `a`, if `b` nests in `a`, then you can call the gate with `b`", which is how earth functions work.  Asking "given a wet gate with formal sample `a`, what is the subtyping relationship such that if `b` is a subtype of `a`, then we can call that wet gate with `b`" is a really hard question because it's trying to treat a wet gate like an earth function.

philipcmonk:
Closing this in favor of #1150.

Thanks Anton for working through this with me!

ohAitch:
I will note that's not the complicated bit, and gives a fairly
unambiguous subtyping relationship I described as "has the same faces
in the same places". The complicated bit is "repainting", which is the
mechanism that messes with "actually compiled to the type currently in
the sample" such that passing literals without faces into wet gates
that use faces nonetheless sometimes works; ("electroplating" and
"vulcanization" were the previous less-ambitious algorithms which
aimed for a similar goal, but iirc didn't declare recursive types in
scope)

On 4/16/19, Philip Monk <notifications@github.com> wrote:
> Closed #1145.
>
> --
> You are receiving this because you were mentioned.
> https://github.com/urbit/arvo/pull/1145#event-2281282419


-- 
Here lies a toppled god —
His fall was not a small one.
A narrow and tall one.

PR #1148, opened 2019-04-16
add +bake

philipcmonk:
As described in #1145, wet gates are often more tricky to use than dry gates.  `+bake` takes a wet gate and a type and monomorphizes the wet gate into a dry gate with the given argument type.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -615,7 +615,7 @@
   ::                                                    ::
   ::                                                    ::
 ::
-::  +snoc Append an element to the end of a list.
+::  +snoc: append an element to the end of a list
 ::
 ++  snoc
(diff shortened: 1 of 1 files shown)


pilfer-pandex:
Sounds good to me! Feel like we could bake the name a bit more though.

sys/hoon.hoon
```diff
@@ -674,6 +674,13 @@
     --
   a
 ::
+::  +bake Convert wet gate to dry gate by specifying argument mold
```
belisarius222: `+bake: convert wet gate to dry gate by specifying argument mold`

philipcmonk: ah, I just used a 1-nearest neighbor algorithm for determining comment style.

PR #1151, opened 2019-04-19
%acme: use cert from clay files

Fang-:
@joemfb as mentioned over email!

Does what it says on the can. Successfully tested on ~paldev.


ohAitch:
(Do we have non-world-readable clay directories yet)

joemfb:
yes, we do @ohAitch.

Fang-:
Thanks @joemfb! I agree with you on both counts. I'll add a TODO for factoring it out eventually, to more strongly signal the temporary intention of this code, and then merge this in.

PR #1154, opened 2019-04-25
adds Arvo %warn error notification event handler

joemfb:
And slightly tweaks %dill's printing of %crud errors. This is the dual of urbit/urbit#1245

### DIFF
--- a/sys/vane/dill.hoon
+++ b/sys/vane/dill.hoon
@@ -166,7 +166,7 @@
       ++  crud
         |=  {err/@tas tac/(list tank)}
         =+  ^=  wol  ^-  wall
-            :-  (trip err)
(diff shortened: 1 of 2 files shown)


joemfb:
That's fine with me. It's worth noting that we'll need to do the same branch dance in the urbit/urbit cc-release arvo submodule.

sys/vane/dill.hoon
```diff
@@ -166,7 +166,7 @@
       ++  crud
         |=  {err/@tas tac/(list tank)}
         =+  ^=  wol  ^-  wall
-            :-  (trip err)
+            :-  :(weld "%" (trip err) " event failed:")
```
belisarius222: LGTM

PR #1155, opened 2019-04-26
Landscape load fixes

ixv:
For OTA; I'm not going to bother targeting this at hotfix, just to then PR that into master.

This PR reduces the upfront cost of loading inbox and chat pages on landscape. It also contains a few other landscape bugfixes.


galenwp:
V excited about this.

PR #1165, opened 2019-05-09
add .DS_Store to .gitignore

jfranklin9000:
yep, i'm a mac user


Fang-:
Pretty sure you want to add it to your global `.gitignore` instead? I don't think repo-specific `.gitignore`s should concern themselves with OS or editor files (unless the repo is only usable by one specific editor).

jtobin:
I'm inclined to agree with @Fang- here.  I don't have terribly strong opinions on what ought go in a project's `.gitignore`, but arvo's is pretty minimal as-is, so for hygienic purposes we might as well keep it that way and restrict entries to be project- -- rather than contributor- -- specific.

(By that logic of course we can just gut the whole `.gitignore` file as-is, since I believe the only project-specific entry in there is no longer relevant.)

jfranklin9000:
OK, I added it to my global .gitignore.

ohAitch:
+1: .gitignore is for files reader by tools included/referenced in the repo
(*.o node_modules etc), .git/info/exclude and ~/.gitignore are for
file-browser editor backup etc artefacts.

PR #1166, opened 2019-05-09
remove +hard, replace with new ;;

pilfer-pandex:
The previous commit changed the semantics of ;; to just call the mold with the result of the hoon, rather than having it macroexpand to something like hard. By "hard" is meant applying the mold and then checking whether the result is unchanged. This second commit removes the +hard function itself, with the effect that there are no remaining mold "idempotence" tests in the system.

#1055

### DIFF
--- a/lib/hall-legacy.hoon
+++ b/lib/hall-legacy.hoon
@@ -144,7 +144,7 @@
     %-  of  :~
       name+(ot nom+so mon+tors ~)
       text+(cu to-wain:format so)
-      tank+(ot dat+(cu ;;((list tank) blob)) ~)
+      tank+(ot dat+(cu (list tank) blob) ~)
     ==
   ::
   ++  blob  (cu cue (su fel:de-base64:mimes:html))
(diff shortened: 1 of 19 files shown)


cgyarvin:
I believe this to be right and proper.   

While the equation of type, schema and validator is extremely useful, Postel’s law is much less so — you rarely want to normalize incoming network data, though schema evolution remains a thing.  And what the “idempotence” (really fixpoint) test is testing is not the correctness of the data but the correctness of the schema, which really no one needs...

pilfer-pandex:
You're right. My

lib/hall-legacy.hoon
```diff
@@ -144,7 +144,7 @@
     %-  of  :~
       name+(ot nom+so mon+tors ~)
       text+(cu to-wain:format so)
-      tank+(ot dat+(cu (hard (list tank)) blob) ~)
+      tank+(ot dat+(cu ;;((list tank) blob)) ~)
```
pilfer-pandex: Not a call to hard, but still wrong.

PR #1170, opened 2019-05-14
lte: restore lens control plane

joemfb:
This PR is the dual of urbit/urbit#1277. %rver implements the new %request-local task (thanks @ixv!), and there's an incredibly basic, one-request-at-a-time lens app. Just enough to get the CI and pill staging back to a working state (on urbit/urbit).

### DIFF
--- a/sys/vane/rver.hoon
+++ b/sys/vane/rver.hoon
@@ -675,6 +675,31 @@
   ::
   |=  [[our=@p eny=@ =duct now=@da scry=sley] state=server-state]
   |%
+  ::  +request-local: bypass authentication for local lens connections
+  ::
+  ++  request-local
+    |=  [secure=? =address =request:http]
+    ^-  [(list move) server-state]
+    ::
+    =/  act  [%app app=%lens]
+    =/  connection=outstanding-connection
+      [act [& secure address request] ~ 0]
+    ::
+    =.  connections.state
+      (~(put by connections.state) duct connection)
+    ::
+    :_  state
+    :_  ~
+    :^  duct  %pass  /run-app/[app.act]
+    ^-  note
+    :^  %g  %deal  [our our]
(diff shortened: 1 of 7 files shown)


sys/vane/rver.hoon
```diff
@@ -675,6 +675,31 @@
...
+      [act [& secure address request] ~ 0]
+    ::
+    =.  connections.state
+      (~(put by connections.state) duct connection)
+    ::
+    :_  state
+    :_  ~
+    :^  duct  %pass  /run-app/[app.act]
+    ^-  note
+    :^  %g  %deal  [our our]
```
eglaysher: I worry a bit about how this appears to be hardcoding everything to the %lens app, but I suppose that's fine and maintaining backwards compatibility.

joemfb: Right - this PR is just restoring the old lens functionality. I think the correct approach would be to make lens (and herb) make and handle generic http requests, using dynamically-bound routes, generalized across local and remote servers. Then the only difference between `%request` and `%request-local` would be that the latter is pre-authorized. But all that will take some work.

ohAitch:
It seems a lot more robust to use the standard eyre app authentication
mechanism, perhaps with a lens token written out into .urb for shell
access. I think a lot of the lens port separation was there to quarantine
code shoehorning "%poke with empty ack, console %peer with several spurious
responses" into a single JSON rpc call-response, which looks like it's now
directly supported / entirely in userspace?

joemfb:
Full stateful http-request handling is now in userspace (with lighter-than-eyre), Lens/herb/urb.py should be changed to work that way, and to support the standard cookie-based authentication. But I still think it's worth keeping around the lens server as a automatically-authorized local-only channel -- it's handy for scripting, CI, etc.

PR #1171, opened 2019-05-15
Use spec in lib/collections expression.

jtobin:
4b2a3d877 replaced

```
  =/  c=[term cord]  ((hard ,[term cord]) b)
```

with

```
  =/  c=[term cord]  ;;([term cord] b)
```

but without the comma, `[term cord]` is not parsed as a spec.  As a result, collections errors on boot.  This commit replaces the cell with a pair, yielding a proper spec.


belisarius222:
Seems like the real problem here is that `;;` doesn't properly switch the parser into pattern mode when in flat form. The comma on the first sub-hoon shouldn't be necessary.

ohAitch:
It having that parser mode is the reason to keep it despite being redundant
with %- even!

joemfb:
As of #1164, the `%mcmc` parser on master does use pattern mode . @jtobin, I'm not seeing this collections failure on boot, how can I reproduce it?

joemfb:
I suspect you're booting into a mismatch between compiler and userspace - a compiler that doesn't have `%mcmc` changes and a userspace that expects them. And I further suspect that this is happening in the new cc-release pill staging scripts. The fake-ship provisioning was recently simplified to make CI work better, but it makes these mismatches more likely. Long story short, the pill update scripts need to boot ships without `-A`. This is my bad ....

jtobin:
Story checks out -- I observed this when using `-A`on an Arvo directory that included the `;;` changes, but the pill used was indeed the standard `0.7.4` one grabbed from `bootstrap.urbit.org`.  I'll nuke this one.

PR #1172, opened 2019-05-15
remove ~&pure-mint from mint (prelude to jet decapitation)

pilfer-pandex:
This message printed on each recursion of `mint:ut` radically slows down the hoon code, and must be disabled if the hoon implementation is to take the place of the C.


jtobin:
Minor quibble, but is there any reason not to just delete the thing instead of commenting it out?  Generally I think it's good practice to be more aggressive than less when it comes to gutting dead code.

PR #1174, opened 2019-05-16
Clad improvements

philipcmonk:
Two changes:

- When sending a note with the clad monad, you may include a path which gets appended on the wire.  This is used exclusively for documentation purposes, to counteract the negative effects of continuation-based programming on inspectability by `|verb`.
- I extracted the essential character of the clad monad's evaluator into a pure function in `+clad`.  Now, `+take:eval:clad` deals with the specifics of the monad (eg accumulating effects, implementing the `%cont` handling that ensures the associativity of `+bind`, etc).  Thus, `+take-commit` and `+take-merge` only have to deal with the imperative parts (fetching the transaction from our global state and writing it back) and the results when the transaction finishes.

Additional generalizations that I expect to happen at some point:

- Generalize `+clad` to be completely non-specific to Clay, and move it into zuse.  I think all that's necessary is parameterizing the `move`, `sign`, `note`, and `input` types.

- Make a core that manages all running transactions within one opaque piece of state, so you don't have to duplicate `+take-merge` and `+take-commit`.

These will happen as soon as I have a good reason to do them, because I don't believe in generalizing without an external reason to do so.

Note: this is targeted against #1169 because I didn't want to push new features directly to #1169 because it's already been reviewed.  I also don't want to interfere with #1169 hitting master -- that shouldn't be blocked on this PR.

### DIFF
--- a/sys/vane/clay.hoon
+++ b/sys/vane/clay.hoon
@@ -153,10 +153,9 @@
   %-  unit
   $:  hen=duct
       req=task:able
-      mos=(list move)
-      $=  cad
-      $%  [%commit com=form:commit-clad]
-          [%merge mer=form:merge-clad]
+      $=  eval-data
+      $%  [%commit commit=eval-form:eval:commit-clad]
+          [%merge merge=eval-form:eval:merge-clad]
       ==
   ==
 ::
@@ -306,7 +305,8 @@
 ::
 ::  notes:   notes to send immediately.  These will go out even if a
 ::           later stage of the process fails, so they shouldn't have any
-::           semantic effect on the rest of the system.
+::           semantic effect on the rest of the system.  Path is
+::           included exclusively for documentation and |verb.
 ::  effects: moves to send after the process ends.
 ::  wait:    don't move on, stay here.  The next sign should come back
 ::           to this same callback.
@@ -342,7 +342,7 @@
   ++  output  (clad-output-raw a)
   ++  form  (clad-form-raw a)
   ++  pure
-    |=  [arg=a]
+    |=  arg=a
     ^-  form
     |=  clad-input
     [~ ~ %done arg]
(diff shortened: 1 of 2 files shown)


belisarius222:
The code itself looks good, though, so I'll approve this once I understand the move flopping.

joemfb:
The flopping outside of `+clad` happens twice: once in `+emil`, again in `+abet`. It looks like the `+clad` interpreter is just emitting moves in the order it expects them to be run.

philipcmonk:
Yes, the interpreter produces a list of moves in "natural order".  A list of moves is in natural order if we expect the head of the list to be the first move evaluated.

It's common in cores to internally keep moves in reverse order and flop it at the end because that makes it easier to add a single event to the end of the list.  If you're doing that, you should be very careful and should perhaps document it.  But if we see something accept or produce a list of moves, we assume that it's in natural order.

If you're storing moves in natural order (uncommon in cores), then `+emit` (which should add a move to be evaluated after all the previously-emitted moves) is `=.  moves  (flop [new-move (flop moves)])`.  But `+emil` (which should emit a list of moves in natural order after previous moves) is simple: `=.  moves  (weld moves new-moves)`.  

If you're storing moves in reverse order, as in `+de`, then `+emit` is very simple: `=.  reverse-moves 
 [new-move reverse-moves]`.  `+emil` is slightly more complex, and I think there's a couple places where we get this wrong: `=.  reverse-moves  (weld (flop new-moves) reverse-moves)`.

In both cases, you could define `+emil` in terms of `+emit`:

```
|-  ^-  +>.^$
?~  moves
  +>.^$
=.  +>.^$  (emit i.new-moves)
$(moves t.moves)
```

We usually store moves in reverse order, but this should be entirely abstracted away by `+emit` and `+emil`.  Unless you're writing those functions themselves, any time you see a list of moves, you should expect it to be in natural order.

jtobin:
> Note: this is targeted against #1169 because I didn't want to push new features directly to #1169 because it's already been reviewed. I also don't want to interfere with #1169 hitting master -- that shouldn't be blocked on this PR.

Feel free to merge this one into #1169, given the fellas' reviews -- I can merge the whole clay & clad goodness into master tomorrow.  Or since you know I plan on doing it anyway, you can. 😄

philipcmonk:
Will do so once @belisarius222 is satisfied with the move order.

PR #1176, opened 2019-05-24
fixes the macro expansion of $> and $< specs

joemfb:
The expansions of `$>` and `$<` were infinitely recursive. These are new specs for building molds from parts of other `$%` molds. They are `?>`-style assertions against the head of the noun being cast or nest-checked. For example:

```
> `task:able:ames`[%west ~zod /foo 1]
[%west p=~zod q=/foo r=1]
>
> `$>(%west task:able:ames)`[%west ~zod /foo 1]
[%west p=~zod q=/foo r=1]
> 
> `$<(%west task:able:ames)`[%west ~zod /foo 1]
nest-fail
```

I think it would be very helpful to use these specs when building the `+note` and `+sign` molds for vanes, which are obviously subtypes of the `+task` and `+gift` types of the other vanes in the system. If there's interest, I'd be happy to do so now.

### DIFF
--- a/sys/hoon.hoon
+++ b/sys/hoon.hoon
@@ -8319,7 +8319,7 @@
       ::
           {$bsld *}
         :+  %tsls
-          relative
+          relative:clear(mod q.mod)
         :+  %wtld
           [%wtts [%over ~[&/3] p.mod] ~[&/4]]
         $/2
@@ -8328,7 +8328,7 @@
       ::
           {$bsbn *}
         :+  %tsls
-          relative
+          relative:clear(mod q.mod)
         :+  %wtbn
           [%wtts [%over ~[&/3] p.mod] ~[&/4]]
         $/2


joemfb:
These specs have some limitations that should be called out explicitly. Most notably, some of them are not directly bunt-able:

```
> *$<(%west task:able:ames)
[%want p=~zod q=/ r=0]
> *$>(%west task:able:ames)
```

This "intuitively" follows from that fact that specs are not available at runtime, but the user-experience is less than intuitive. The ability-to-bunt can be restored (in a frustratingly-explicit nanner) with `$~`:

```
> *$~([%west ~zod / 0] $>(%west task:able:ames))
[%west p=~zod q=/ r=0]
```

As I mentioned, I think the right way to use these specs is in building up a `$%` that references other specs, like `+note` and `+sign` in the vanes. The whole `$%` should be wrapped in a single `$~` to ensure that it bunts, but the individual "pages" of the tagged-union "book" need not be bunt-able.

cgyarvin:
Ha not surprised that I f*ed this up.  

Yes these are very useful for dynamic checks of static properties we can’t statically check.  

They should be applied to containers as well — more urgent.  Also to a layer of managed string types.  No one should be able to send you a misordered set or a dodgy string over the network.

joemfb:
I think `$|` is the key for dynamic checks. `+map` should be something like:

```
+*  map  [key value]
  $|  (tree (pair @ @))
  |=(a=(tree (pair @ @)) ~(apt by a))
```

For example:

> ($|((tree (pair @ @)) |=((tree (pair @ @)) ~(apt by +<))) [[1 2] ~ [[3 4] ~ ~]])
{[p=1 q=2] [p=3 q=4]}
> ($|((tree (pair @ @)) |=((tree (pair @ @)) ~(apt by +<))) [[1 2] [[3 4] ~ ~] ~])
```

cgyarvin:
There is also a way to fix this...

cgyarvin:
Ha ha of course, wrong rune pair.

PR #1178, opened 2019-05-27
deduplicates vane interfaces (+note/+sign)

joemfb:
This PR uses `$>` (fixed in #1176) to deduplicate the `+note` (outgoing "request") and `+sign` (incoming "response") interfaces of all the vanes, defining these interfaces in terms of the `+task` (incoming "request") and `+gift` (outgoing "response") interfaces of other vanes wherever possible. 

I've also moved all the `+note`/`+sign` definitions into the vanes themselves  -- some were previously in %zuse, but there's no reason for anything but the vanes (and their tests) to reference these.

### DIFF
--- a/sys/vane/jael.hoon
+++ b/sys/vane/jael.hoon
@@ -117,8 +117,8 @@
           $>(%look task)                                ::  set ethereum source
       ==                                                ::
       $:  @tas                                          ::
-  $%  [%init p=ship]                                    ::  report install
-      [%sunk p=ship q=rift]                             ::  report death
+  $%  $>(%init vane-task)                               ::  report install
+      $>(%sunk vane-task)                               ::  report death
   ==  ==  ==                                            ::
 ::                                                      ::
 +$  sign                                                ::  in result $<-
--- a/sys/vane/jael.hoon
+++ b/sys/vane/jael.hoon
@@ -94,10 +94,40 @@
       [%vent-result p=vent-result]                      ::  tmp workaround
   ==                                                    ::
 ++  card                                                ::  i/o action
-  (wind note:able gift)                                 ::
+  (wind note gift)                                      ::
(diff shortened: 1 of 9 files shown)


sys/vane/jael.hoon
```diff
@@ -94,10 +94,40 @@
...
+          $>(%hiss task:able:eyre)                      ::  http request
+      ==                                                ::
+      $:  %a                                            ::    to %ames
+          $>(%want task:able:ames)                      ::  send message
+      ==                                                ::
+      $:  %j                                            ::    to self
+          $>(%look task)                                ::  set ethereum source
+      ==                                                ::
+      $:  @tas                                          ::
+  $%  $>(%init vane-task)                               ::  report install
```
belisarius222: this formatting is inconsistent with the rest

sys/zuse.hoon
```diff
@@ -7105,20 +7066,55 @@
...
-      [%crud tag=@tas =tang]                            ::  any vane: error report
-      {$hail ~}                                         ::  dill: refresh
-      {$hear p/lane:ames q/@}                           ::  ames: input packet
-      {$hook ~}                                         ::  dill: hangup
-      {$into p/desk q/? r/mode:clay}                    ::  clay: external edit
-      {$they p/@ud q/httr:eyre}                         ::  eyre: in response
-      {$this p/? q/clip:eyre r/httq:eyre}               ::  eyre: in request
-      {$thud ~}                                         ::  eyre: in cancel
-      {$wake ~}                                         ::  behn: wakeup
++$  unix-task                                           ::  input from unix
```
belisarius222: It might be nice to define vane interfaces in terms of this, rather than the other way around. That way you could look at a vane interface and immediately see which moves come from Unix. Could be done in a different PR, though.

PR #1179, opened 2019-05-28
More useful +diff and +sigh in gall apps

philipcmonk:
Suppose `my-mark` has type `my-type`.

In gall apps, `+diff-my-mark` handles all `%diff` cards with mark `my-mark` and takes an argument of type `[wire my-type]`.

However, `+diff` handles all `%diff` cards not otherwise handled, and takes an argument of type `[wire cage]`.  A `cage` includes `vase`, which is very unhelpful (so it was never used).

This is obviously wrong.  This PR makes it so `+diff` takes an argument of type `[wire %my-mark my-type]`.  This gives us a well-typed value, so we can actually do something with this.

The same applies to `+sigh`, so I changed that too.

The motivation for this is in the `philip/trad` branch, where I would like to be able to handle all subscription results in a single arm because the library that creates this arm doesn't know until compile time what are the possible subscription results (it gets the types via `|*  mold` parameterization).  See https://github.com/urbit/arvo/blob/40c4ea45d22ce02b36033736e580555f6da83aab/lib/tapp.hoon#L149-L159 and the definition of `in-peer-data` https://github.com/urbit/arvo/blob/40c4ea45d22ce02b36033736e580555f6da83aab/lib/tapp.hoon#L3-L8

### DIFF
--- a/sys/vane/gall.hoon
+++ b/sys/vane/gall.hoon
@@ -682,12 +682,17 @@
       =+  ^=  arg  ^-  vase
           %-  slop
           ?:  =(0 p.u.cug)
-            [!>(`path`+.pax) !>(cag)]
+            [!>(`path`+.pax) (ap-cage cag)]
           [!>((slag (dec p.u.cug) `path`+.pax)) q.cag]
       =^  cam  +>.$  (ap-call q.u.cug arg)
       ?^  cam
         (ap-pump:(ap-lame q.u.cug u.cam) | her pax)
       (ap-pump & her pax)
+    ::
+    ++  ap-cage                                         ::  cage to tagged vase
+      |=  cag/cage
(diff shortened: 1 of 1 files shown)


belisarius222:
Am I correct in interpreting this as a shift from `[mark vase]` to `!>([mark data])`? That was a change I was pushing for in Ford-Fiesta, and I suspect there are several places we should make that transformation.

philipcmonk:
No, from gall's perspective it's a shift from `!>([mark !>(data)])` to `!>([mark data])`.  From the app's perspective, you take off the outer `!>`, so it's a shift from `[mark !>(data)]` to `[mark data]`.

cgyarvin:
!>(mark) part surely not that useful...

PR #311, opened 2016-12-08
applies map/set invariants inside the molds

joemfb:
also renames `+-ept:by` to `+-apt:by`, and simplifies `++qeu`.

my branch was created on top of cgyarvin/cc-merge. Should I reparent these commits somewhere else?

PR #1180, opened 2019-05-29
validate map/set treap order with $|

joemfb:
after:

```
> ((map) [[1 2] ~ [[3 4] ~ ~]])
{[p=1 q=2] [p=3 q=4]}
> ((map) [[1 2] [[3 4] ~ ~] ~])
```

before:

```
> ((map) [[1 2] ~ [[3 4] ~ ~]])
{[p=1 q=2] [p=3 q=4]}
> ((map) [[1 2] [[3 4] ~ ~] ~])
{[p=3 q=4] [p=1 q=2]}
```

much before:

#311

I had to move these mold/mold-builder definitions so that the `+in` and `+by` cores are available to them. Note that the section labels are still wrong, although somewhat less so. I think that's out of scope for this PR, as the labels are wrong in many parts of hoon.hoon.


belisarius222:
^- and ^+ casting do not clam, so the menace of malicious map misordering remains. At least now the validator will get called on network data though, so you have to be on the inside (e.g. an app) to cause a problem.

—
~rovnys-ricfer
https://urbit.org

joemfb:
Right, molds (specs) only exist at compile time. `^-(spec hoon)` desugars to `^+(*spec hoon)`, so only the bunt of the mold is used for casting -- the validation function is not involved (or even available).

Some more examples:

```
> ((set) [2 [1 ~ ~] ~])
{1 2}
> ((set) [2 ~ [1 ~ ~]])
```

Note also that the validation is applied recursively (`(jug)` is `(map * (set))`):

```
> `*`(my [1 (sy 2 3 ~)] [4 (sy 5 6 ~)] ~)
[[4 6 [5 0 0] 0] [[1 2 0 3 0 0] 0 0] 0]
> ((jug) [[4 [6 [5 ~ ~] ~]] [[1 [2 ~ [3 ~ ~]]] ~ ~] ~])
{[p=1 q={2 3}] [p=4 q={5 6}]}
> ((jug) [[4 [6 [5 ~ ~] ~]] ~ [[1 [2 ~ [3 ~ ~]]] ~ ~]])
> ((jug) [[4 [6 ~ [5 ~ ~]]] [[1 [2 ~ [3 ~ ~]]] ~ ~] ~])
```

philipcmonk:
Makes sense.

Is there a reason why we don't verify the vertical ordering of a qeu?

joemfb:
I guess nobody has taken the time to write `+apt:to`. I think it would be something like this:

```
++  apt
  ^-  ?
  ?~  a  &
  |-  ^-  ?
  ?~  l.a  &
  ?~  r.a  &
  &((mor n.l.a n.r.a) $(a l.a) $(a r.a))
```

jtobin:
(extra :+1: for the tests)

PR #1183, opened 2019-05-30
Transaction apps

philipcmonk:
This includes three additional libraries:

- `lib/tapp` -- A library to create imperative-looking apps via the transaction monad.
- `lib/trad` -- The asynchronous transaction monad.
- `lib/stdio` -- A set of useful asynchronous functions (ie functions that produce values of the asynchronous transaction monad) for various standard requests (HTTP, timers, app pokes and peers, etc)

There are also two example apps to demonstrate how these apps can be created.  When we have real apps that use these libraries, maybe we should delete these, but for now I think this is the right place for them.

Notice how something like "fetch a web page" happens inline, with no need to cut your app into many functions to handle the different responses.  This is the main idea behind these libraries.  You can still cut your app by producing a move as an effect rather than in-line (which you want to do if you want to be able to upgrade or process other input before receiving a response), but this makes it easy to not do that when you don't want to.

This also provides a straightforward upgrade approach for situations where you're in the middle of an asynchronous computation.  In explicit state machines, you need to write specific code to be able to upgrade at any point regardless of whether you're in the middle of a computation.  In practice, we usually ignore this, risking breakage for any user in the middle of a computation.  With these imperative-style apps, if you upgrade in the middle of a computation, you can simply cancel the computation and restart it, since computations are transactions that span multiple events.

I'm not super happy with the names here, particularly lib/tapp and lib/trad.  For lib/trad, perhaps simply "async" would be better?  Then, for example, `form:(async ,@)` would be read as "the type of an asynchronous computation that produces an atom".  This name should be short because it's written pretty often.  Particularly, any code that uses it will have many functions that start with:

```
|=  ...
=/  m  (trad ,return-type)
^-  form:m
```

A note on style:  I use the `=/  m  (trad ,return-type)` pattern religiously at the top of each arm that will produce an asynchronous computation.  I think this is good because you'll find that you'll call `form:m`, `pure:m`, and `bind:m` often in the middle of the computation, and it's helpful when reading the code because it makes it immediately obvious what the computation produces.  This style should be enforced in `lib/stdio` and probably any apps that use it.

### DIFF
--- a/app/example-tapp-fetch.hoon
+++ b/app/example-tapp-fetch.hoon
@@ -41,7 +41,7 @@
     :*  top-stories=(weld base "topstories.json")
         item=|=(item=@ud `tape`:(welp base "item/" +>:(scow %ui item) ".json"))
     ==
---
+  --
 =,  trad=trad:tapp
 =,  tapp-trad=tapp-trad:tapp
 =,  stdio
@@ -89,7 +89,7 @@
   |-  ^-  form:m
   =*  loop  $
   ::
-  ::  If done, tell subscriers and print the results
+  ::  If done, tell subscribers and print the results
   ::
   ?~  top-stories
     ;<  ~  bind:m  (give-result /comments %comments top-comments)
@@ -139,5 +139,4 @@
   ;<  =state  bind:m  (handle-poke %noun 'fetch')
   =.  top-comments  state
   (pure:m top-comments)
-  ::  (handle-poke %noun 'poll')
 --
--- a/app/example-tapp-fetch.hoon
+++ b/app/example-tapp-fetch.hoon
@@ -129,7 +129,7 @@
   |=  =path
   =/  m  tapp-trad
   ^-  form:m
-  ~&  [%baby-take-peer path]
+  ~&  [%tapp-fetch-take-peer path]
   (pure:m top-comments)
 ::
 ++  handle-take
--- a/app/example-tapp-subscribe.hoon
+++ b/app/example-tapp-subscribe.hoon
@@ -32,9 +32,9 @@
       (trad-fail %no-subscription ~)
     ;<  ~  bind:m  (pull-app [target path]:u.subscription)
     (pure:m ~)
-  ;<  ~  bind:m  (poke-app [our.bowl %baby] %noun 'print')
-  ;<  ~  bind:m  (peer-app [our.bowl %baby] /comments)
-  =.  subscription  `[[our.bowl %baby] /comments]
+  ;<  ~  bind:m  (poke-app [our.bowl %example-tapp-fetch] %noun 'print')
+  ;<  ~  bind:m  (peer-app [our.bowl %example-tapp-fetch] /comments)
+  =.  subscription  `[[our.bowl %example-tapp-fetch] /comments]
   ;<  ~  bind:m  (wait (add now.bowl ~s3))
   (pure:m subscription)
 ::
@@ -43,6 +43,6 @@
   =/  m  tapp-trad
   ^-  form:m
   ?>  ?=(%comments -.data)
-  ~&  sitter-got-data=(lent comments.data)
+  ~&  subscriber-got-data=(lent comments.data)
   (pure:m subscription)
 --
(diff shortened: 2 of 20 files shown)


vvisigoth:
@philipcmonk I was thinking of building somethign small to try this out. Do you think something like [pastebinit](https://help.ubuntu.com/community/Pastebinit) would work?

philipcmonk:
Yeah, that would be a good project.

lib/tapp.hoon
```diff
@@ -0,0 +1,337 @@
...
+::
+++  create-tapp-poke-peer-take
+  |=  handler=tapp-core-poke-peer-take
+  %-  create-tapp-all
+  |_  [=bowl:gall state=state-type]
+  ++  handle-poke  ~(handle-poke handler bowl state)
+  ++  handle-peer  ~(handle-peer handler bowl state)
+  ++  handle-diff  |=(* (trad-fail:trad-lib %no-diff-handler >path< ~))
+  ++  handle-take  ~(handle-take handler bowl state)
+  --
```
Fang-: The repetition of all of the above is a bit meh, but I don't think we have any fancy affordances that could make this better? It's fine as-is, but already shows that it scales pretty poorly. Worth taking note of.

app/example-tapp-subscribe.hoon
```diff
@@ -0,0 +1,48 @@
...
+  =/  m  tapp-trad
+  ^-  form:m
+  ?:  =(cord.in-poke-data 'pull')
+    ?~  subscription
+      (trad-fail %no-subscription ~)
+    ;<  ~  bind:m  (pull-app [target path]:u.subscription)
+    (pure:m ~)
+  ;<  ~  bind:m  (poke-app [our.bowl %example-tapp-fetch] %noun 'print')
+  ;<  ~  bind:m  (peer-app [our.bowl %example-tapp-fetch] /comments)
+  =.  subscription  `[[our.bowl %example-tapp-fetch] /comments]
```
Fang-: Probably want a  `++  target  [our.bowl %example-tapp-fetch]` to make these lines a little less noisy.

app/example-tapp-fetch.hoon
```diff
@@ -0,0 +1,143 @@
...
+  (pure:m top-comments)
+::
+++  handle-take
+  |=  sign:tapp
+  =/  m  tapp-trad
+  ^-  form:m
+  ;<  =state  bind:m  (handle-poke %noun 'fetch')
+  =.  top-comments  state
+  (pure:m top-comments)
+  ::  (handle-poke %noun 'poll')
```
Fang-: Can this go?

app/example-tapp-fetch.hoon
```diff
@@ -0,0 +1,143 @@
...
+  =/  top-stories=(list @ud)
+    ((ar ni):dejs:format top-stories-json)
+  ::
+  ::  Loop through the first 5 stories
+  ::
+  =.  top-stories  (scag 5 top-stories)
+  |-  ^-  form:m
+  =*  loop  $
+  ::
+  ::  If done, tell subscriers and print the results
```
Fang-:
```suggestion
  ::  If done, tell subscribers and print the results
```

app/example-tapp-fetch.hoon
```diff
@@ -0,0 +1,143 @@
...
+  ::
+  ?~  top-stories
+    ;<  ~  bind:m  (give-result /comments %comments top-comments)
+    (handle-poke %noun 'print')
+  ::
+  ::  Else, fetch the story info
+  ::
+  ~&  "fetching item #{+>:(scow %ui i.top-stories)}"
```
Fang-: Can't believe we still don't have a known-to-all, simple way to print "normal" numbers. I'd use `((d-co:co 1) i.top-stories)` here, which feels a little less hacky, but is also more obscure. Hard problems!

lib/tapp.hoon
philipcmonk: We don't have any easy way to say "call this if it exists", and I don't think the `core` type is flexible enough to make that work.  The other option would be to say that everyone has to implement all the arms in `tapp-core-all`.  You could say that each arm is either the gate or null, so not handling `diff`, `take`, and peer would look like this at the bottom of the app:

```
++  handle-diff  ~
++  handle-take  ~
++  handle-peer  ~
```

PR #1184, opened 2019-05-31
Merge remaining contributions on 'next' branch

jtobin:
@joemfb pointed out that there were still a few things in `next` that hadn't made it into master -- this just merges them in before we retire that branch for good.

*pours out a fifth of bourbon*


joemfb (approved):
LGTM

PR #1186, opened 2019-06-01
adds explicit initialization and peek (namespace) handlers to lib/tapp

joemfb:
And a couple default namespace paths for inspecting the state of a tapp.

### DIFF
--- a/lib/tapp.hoon
+++ b/lib/tapp.hoon
@@ -48,7 +48,11 @@
   %-  create-tapp-poke-peer
   |_  [=bowl:gall state=state-type]
   ++  handle-poke  ~(handle-poke handler bowl state)
-  ++  handle-peer  |=(* (async-fail:async-lib %no-peer-handler >path< ~))
+  ++  handle-peer
+    |=  *
+    ~|  %default-tapp-no-sole
+    ?<  ?=([%sole *] +<)
+    (async-fail:async-lib %no-peer-handler >path< ~)
   --
 ::
 ::  The form of a tapp that only handles pokes and peers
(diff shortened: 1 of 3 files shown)


lib/tapp.hoon
```diff
@@ -28,6 +30,8 @@
       active=(unit eval-form:eval:tapp-async)
       app-state=state-type
   ==
++$  tapp-peek
+  [%noun ?(? (set contract))]
```
jtobin: My mental parser paused for a second here to verify that `?(?` wasn't some strange irregular syntax it hadn't seen before.

philipcmonk:
I don't have ny github password on my phone so I can't mark it approved
right now, but this looks good to me.

lib/tapp.hoon
```diff
@@ -44,7 +48,11 @@
   %-  create-tapp-poke-peer
   |_  [=bowl:gall state=state-type]
   ++  handle-poke  ~(handle-poke handler bowl state)
-  ++  handle-peer  |=(* (async-fail:async-lib %no-peer-handler >path< ~))
+  ++  handle-peer
+    |=  *
+    ~|  %default-tapp-no-sole
+    ?<  ?=([%sole *] +<)
```
jtobin: As an anti-larkite, I would give the sample a face and kill the lark here:

```
++  handle-peer
   |=  =noun
    ~|  %default-tapp-no-sole
    ?<  ?=([%sole *] noun)
   (async-fail:async-lib %no-peer-handler >path< ~)
```

jtobin:
🚢

PR #1189, opened 2019-06-06
fixes :acme retry on "bad-nonce" error

joemfb:
This fixes a (dumb) bug I introduced when refactoring the ACME (LetsEncrypt) client.

Signed ACME requests include an anti-replay nonce, which implies server state. Previously-good nonces retrieved from ACME can become "bad" if the server-state in question goes away for some reason. Since any request can produce a "bad-nonce" response, robust clients must be able to intercept this error condition, retrieve a fresh nonce, and then "retry" their request (this makes for some pretty complicated control flow in our explicit state machines). The bug in this case was simply not passing the name of the action to take after retrieving a nonce. The symptom is an :acme app crash with `%bad-nonce-next` in the stack trace.

This is a good candidate for an OTA hotfix release.

/cc @galenwp

### DIFF
--- a/app/acme.hoon
+++ b/app/acme.hoon
@@ -1194,7 +1194,7 @@
   ::  request nonce if expired-invalid
   ::
   ?:  (bad-nonce rep)
-    (nonce:effect spur)
+    (nonce:effect [act spur])
   ::  XX replace with :hall notification
   ::
   ~|  [%sigh-fail wire]


galenwp:
Still not getting https to `~ravmel-ropdyl` — perhaps something on `~samzod` needs to get kicked?

Fang-:
I didn't get a sync on ~palfun, under ~wanzod. I'll try `|bonk`ing the stars and see what happens.

joemfb:
I think ~samzod didn't get a certificate -- if you're on the stars, can you `:acme %retry` it?

Fang-:
image

Looks like that worked!

galenwp:
Yep, that worked. Thanks!

PR #1190, opened 2019-06-12
add runes for 'commenting out' subhoons

pilfer-pandex:
By analogy to clojure https://clojuredocs.org/clojure.core/comment or `#_`, we add runes for commenting out code.

The effect of `=_` is to discard its first subhoon and return the second. That is,

```
=_  p
    q
```

is equivalent to

```
q
```

The only restriction imposed on `p` is that it parse. It need not, for example, type check.

For convenience, we also provide a `!_` rune such that

```
!_  p
```

is equivalent to

```
=_  p
    !!
```


pilfer-pandex:
The clojure equivalent of my `=_` actually deletes its children from the ast, leaving nothing in their place. The code that comes to replace the removed code in clojure is the subsequent sibling of the comment node, not its child. It wasn't immediately (~2m) obvious to me how to do this for hoon, nor did it strike me as particularly necessary. I could look into this more if people have a stronk preference for it though.

ohAitch:
Minor suggestion: how hard would it be for !_ to print p in a ~|? I'm
thinking in particular of usages like !_("Can't implement this case until
the eschaton merge") or !_(%bad-width)

belisarius222:
Why do we want this? I detest commented-out code, and I worry this rune would legitimize and encourage it.

If this rune did require the commented-out code to typecheck, I would be less vehemently opposed to it, since at least the comment would be forced to stay up-to-date as the code around it changes.

pilfer-pandex:
Oftentime in the course of debugging code, you need to comment out parts. But using :: to do this can be quite cumbersome. This is a direct need I came upon in doing work on test, where I want to progressively strip out sections of hall to get to a minimal repro.

pilfer-pandex:
It seems useful to have this as an affordance, and just reject code that has commented out parts in review, if that's your taste.

ohAitch:
Add it to CI, even!

Fang-:
> Oftentime in the course of debugging code, you need to comment out parts. But using :: to do this can be quite cumbersome.

Sounds like you want block comments (boo!), or just to git gud at editing .hoon files?  
It seems _very_ weird to me to solve this using language, especially if we're gonna say "code that contains this is not space-rated/approvable".

belisarius222:
> Oftentime in the course of debugging code, you need to comment out parts. But using :: to do this can be quite cumbersome. This is a direct need I came upon in doing work on test, where I want to progressively strip out sections of hall to get to a minimal repro.

This is better solved by Vim than by Hoon. Here's how I comment out a block of code, which, as you correctly point out, is frequently needed for debugging:

`0 <Ctrl-V> <bunch of down arrow> I :: <Esc>`

`0` goes to the beginning of the line.
`Ctrl-V` starts a rectangular selection.
`Down` extends the selection to the next line.
`I` enters insert at the beginning of each selected line.
`::` inserts the comment.
`Esc` exits insert mode, completing the command.

Works every time! Takes a second or two, and there are probably Vim wizards who know faster incantations.

jtobin:
> Works every time! Takes a second or two, and there are probably Vim wizards who know faster incantations.

No wizardry here, but one time saver is to insert an exact or rough count of the number of lines to descend before hitting the down button (by which I mean `j`, of course 😝).

e.g., for twenty lines:

```
0 <Ctrl-V> 20 j I :: <Esc>
```

(Often you may just want to ballpark a number like 20, and then adjust manually with `j` and `k`.)

jtobin:
This one has been sitting without any action for over a week, so I'm going to close it for the time being.  I suspect it's a bit too contentious a feature to get merged as-is (aside from the complaints above, I'm also not keen on it myself).  Maybe worth taking to the mailing list?

cgyarvin:
Conditional compilation is actually already in there... perhaps you just have to find it!

philipcmonk:
FWIW, after using it with @pilfer-pandex, I'm in favor of `!_`.  It's quite ergonomic for debugging and refactoring.  I think `=_` is a bridge too far, and certainly it shouldn't use a `=` rune slot since it's only for debugging.

PR #1192, opened 2019-06-14
Rver cancels subscriptions on correct gall ducts and cancels behn timers correctly as well

tacryt-socryp:


eglaysher (approved):

sys/vane/rver.hoon
```diff
@@ -60,7 +60,7 @@
       $:  %b
           ::
           ::
-          $%  [%wake ~]
+          $%  [%wake error=(unit tang)]
```
joemfb: This is just half of the necessary timer fix. A `%wake` where `error` is `~` is a normal timer, but when `error` is present it's a report of a previously-failed timer event. The fully correct way to handle this is to unwind whatever state was associated with that timer (maybe canceling subscriptions, builds, producing an HTTP 500 response etc.). The less correct approach is to just print the error via `%dill` if present (this is what most vanes are doing right now). But it's not correct to treat such an event as a normal timer wakeup.

joemfb (changes requested):
The timer handling needs some tweaks, details below:

joemfb:
contained-in/superceded-by #1202

loganallenc:
Closing this as it is superseded by the Modulo merge PR.

PR #1194, opened 2019-06-17
rewrites dynamic dns, disabling indirection, proxying, and automation

joemfb:


joemfb:
@philipcmonk, I'd appreciate your feedback on the changes to `lib/{async,stdio,tapp}.hoon`.

lib/tapp.hoon
```diff
@@ -33,107 +39,171 @@
...
+  ::  Receive syscall result
+  ::
+  ++  handle-take
+    |~  sign
+    *form:tapp-async
+  --
+::
+::  Default handlers for all comands
+::
+++  default-tapp
```
philipcmonk: I like this pattern much better than what I had.

lib/tapp.hoon
```diff
@@ -390,12 +488,19 @@
...
       %-  from-form:eval:tapp-async
       ^-  form:tapp-async
-      ?-  -.u.next
-        %init  ~(handle-init handler bowl app-state)
-        %poke  (~(handle-poke handler bowl app-state) +.u.next)
-        %peer  (~(handle-peer handler bowl app-state) +.u.next)
-        %diff  (~(handle-diff handler bowl app-state) +.u.next)
-        %take  (~(handle-take handler bowl app-state) +.u.next)
+      =/  out
+        %-  mule  |.
```
philipcmonk: This is very good.  We still need to think about non-deterministic errors at some point.

lib/tapp.hoon
```diff
@@ -225,24 +249,66 @@
...
       ~&  [%waiting-until-current-async-finishes waiting]
       `this-tapp
     start-async
   ::
+  ::  Receive acknowledgement of outgoing poke
+  ::
+  ::    XX these can be distinguished by dock, but, without a wire or
+  ::    some alternative, that's not very useful if you poke the same
+  ::    dock multiple times. %poke wires are not currently exposed ...
+  ::    Wat do?
```
philipcmonk: Wires are definitely the right way to handle this.  I guess we should let the user specify a wire when poking?

ohAitch: Like, so it ends up in the bone?

philipcmonk: Not in the bone (at least, not from this app's perspective).  This is a `%coup` response to a poke, so whatever wire we sent on the poke comes back to us as our first argument.  We can then switch on that if we want.

joemfb:
These changes have now been ported to lighter-than-eyre (and a portion of them already merged in #1197)

joemfb:
Merging in the interest of monorepo integration, will solicit further review as necessary.

app/dns-bind.hoon
```diff
@@ -0,0 +1,895 @@
...
+      ::
+      ++  local-uri
+        |=  [our=ship =path]
+        ^-  @t
+        ::  XX can't scry in +mule
+        ::
+        :: =/  =hart:eyre  .^(hart:eyre %r /(scot %p our)/host/real)
+        (crip (en-purl:html [hart [~ path] ~]))
+      ::
+      ::  XX can't scry in +mule
```
philipcmonk: `+mule` calls `+mute`, which calls `+mock` with a trivial scry handler.  Would it be possible to call `+mock` more directly and supply a scry handler that calls `.^`?

joemfb: We could do that, but it's kind of like a double-sided chainsaw from a determinism standpoint. I assume that's why it's not already in the stdlib. To support scry via, we'll either need to do that or operate in a vase context as %gall does.

PR #1195, opened 2019-06-19
Don't crash on life mismatch for incoming packets

Fang-:
Ames crashes are scary, don't tell you anything, and eat up your entire screen.

I replace them with slightly more informative printfs and silent ignoring. These will still eat up your screen, but do so more slowly, and at least give you someone to direct complaints to.


belisarius222:
I sympathize with the inconvenience of big crash messages making the terminal unusable, but I'm not sure we should move away from crashing. Crashing prevents the event from being written to the event log. If we're getting spammed by messages to the wrong life, I suspect we don't want to fill up our disk with them.

Could we have a silent crash somehow?

belisarius222:
It also strikes me that it would be diagnostically useful to log crashed packets to a logging endpoint (not the event log), as long as the log message doesn't interfere with the user-facing terminal.

ohAitch:
The solution here is probably in the space of "log low-SNR errors to a
file, not directly to the console"; perhaps with a rotating log system to
dodge the "fill up our disk" issue.

belisarius222:
Yeah, that's what I was thinking. Not sure how far we are from being able to implement that.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
> Crashing prevents the event from being written to the event log.

I agree, but it seems that's a theoretical problem compared to the actual pain of repeated giant stack traces getting yelled at you.  I would support the idea of a rune that's similar to `!!` but either completely silent or prints just a single line.  Could be a variation on `~|` that truncates the stack before it.

belisarius222:
I wonder if we could solve this by just having Ames refrain from sending its crash traces to Dill to be printed.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
Ah, yes, that's it.  Depending on how many of these there are, we could print a single line.  Or maybe not even that.  Could be nice to have a `|verb %ames` command.

"1 threat quarantined by Urbit Enterprise Antivirus"

joemfb (changes requested):
This one is tough. The constant crash reports when you get into a state of receiving bad packets makes the console unusable. But the crashes are semantically meaningful.

The runtime "catches" crashes and ames and attempts to turn them into negative acknowledgements. The crash reports you're seeing are actually failed attempts to send said nack. I'll defer to @belisarius222 as to exactly which error conditions should be turned into nacks, but there's no question that this functionality is necessary for at least some of them.

The output handling is a problem for several reasons. I think the ideal case would be an entirely separate output channel, observable and inspectable but non-interrupting. The right way to accomplish that is probably running in "daemon mode" with a separate console process, but that remains a ways off.

We could suppress the errors entirely, but I don't think that's acceptable either. It's important to know these failures are taking place. Which leaves either the option of turning these printfs off (@philipcmonk's `|verb %ames`), or limiting the output by truncating the stack traces before sending to %dill. Implementing both is probably the best we can accomplish in the short term.

Thoughts?

belisarius222:
I have an alternative proposal: add a non-interrupting logging service inside Arvo. It's frankly a little strange that we don't already have something like this. What operating system doesn't have a system logger?

Ideally the logs are queryable, tailable (Ctrl-X into the %logs app and tail the messages, ideally with configurable filters), and flush themselves after some (also configurable) amount of time or maximum number of messages. We might also want the service to log a message to the user-facing console when some heuristically-determined condition is met, such as a hundred crashed nack attempts within the past ten minutes.

Production ships might flush logs to rotating Unix log files on a timer. We might also want |verb logging to end up here.

We could use simple string-based logging, but I often wished for structured logging in my previous life as an Earth software developer, and statically typed log messages are obviously the Urbit way. An app should be able to subscribe to a logging stream to calculate statistics, perform anomaly detection, present dashboards, tune system configurations, etc.

An example would be a DDoS protection app that subscribes to a filtered Ames packet log stream, probably identified by path /a/hear, syncs its blacklists with other ships, and dynamically reconfigures Ames to update the blacklist and maybe even tune its congestion control.

Dill comes to mind as candidate for where to house this logging service, possibly backed by Clay for storing the log messages themselves. This is also a candidate for presenting an external scry interface to the runtime, so the runtime could tail or mirror logs.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
I don't want this to stall, so I'd say the easy solution is: keep the printfs (they're useful because they tell you who's giving you bad packets and what's their life), go back to crashing (so we don't record them in our event log), and make the `%crud` handling in ames a no-op instead of slipping to dill.

Fang-:
I'll see if I can do that over the weekend.

joemfb:
As we are through the monorepo looking-glass, please re-open with your changes on urbit/urbit against `v0.8.0.rc`

PR #1200, opened 2019-06-25
Minor clarity improvement for ++fold

baudtack:


baudtack:
Fine fine, lemme change it...

Fang-:
Actually, scratch the "better names" request, this is a small enough scope that it's probably fine to keep short names like this. I don't feel strongly about renaming these, and can't give good concrete suggestions anyway.

baudtack:
Yeah me either... I don't really know if it really needs to be tall formed here either... I'll do it if you really want @Fang- but eh

Fang-:
I don't feel super strongly either. It's not like anyone ever reads `hoon.hoon` anyway. (^:

This PR is an improvement as-is. Merge away!

baudtack:
Honestly the only reason I care is I'm  using fold as an example of wet and iron polymorphism in the docs. and I want the code to be reasonable heh.

PR #1201, opened 2019-06-27
:claz, poweruser Azimuth transaction generation

Fang-:
The _actual_ commits of this PR start at cba368d, everything else is just history from the Azimuth ceremony branches which I pulled in (for git log history purposes), and of which only `:send-txs` and small additions to zuse remain.  
@philipcmonk I just straight-up removed `/lib/ceremony` and most tools in `/app` and `/gen`. Only `:send-txs` is really relevant in this day and age still, right?

`:claz` isn't _actually finished_ yet, but we've already used it for transaction generation internally, and I want this to make it across the monorepo transition.

For context: in the short term, this is intended as an internal tool for generating `.eth-txs` files, to be signed and submitted using `:send-txs`. This'll get fancier over time, and should eventually accommodate powerusers who have no need for Bridge's UI.


joemfb:
One potential point of concern: the integration plan is to merge this (and other outstanding PRs) into master,  merge master into lighter-than-eyre, and monorepo that. This app will then not work, having been written against current eyre. It should still compile, so it should pass CI and need not block the integration. But it will come out the other side in a non-working state.

Porting this to lighter-than-eyre could be accomplished incrementally by sending `%request` (instead of `%hiss`, with `+hiss-to-request:html`) and accepting `+http-response` (instead of `+sigh-*`), or less incrementally by porting the `:claz` to use `lib/tapp` and `lib/stdio` (#1194 includes some examples, as does `:example-tapp-fetch` in on the lighter-than-eyre branch). 
at least one of these will need to be done rather soon, but that can be deferred to the monorepo.

/cc @jtobin

Fang-:
> or less incrementally by porting the :claz to use lib/tapp and lib/stdio

This was already in the cards (handrolled tapp was done for educational purposes), so I'll definitely get to that. Examples are helpful, thanks!

PR #1202, opened 2019-06-27
Modulo

ixv:


app/chat.hoon
```diff
@@ -0,0 +1,606 @@
...
+      ^-  [circle:hall @ud]
+      ?~  lis
+        [cir 0]
+      =/  last  (snag (dec (lent lis)) `(list envelope:hall)`lis)
+      [cir (add num.last 1)]
+  =/  maptjson  *(map @t json)
+  =.  maptjson
+    (~(put by maptjson) 'config' (config-to-json str))
+  =.  maptjson
+    (~(put by maptjson) 'numbers' (numbers-to-json numbers))
```
joemfb: This is not idiomatic hoon -- it'd be more typical to use `+gas:by`, or at least collapse the `=/` and the first `=.`, or to use `=|` instead of the `=/` ...

app/chat.hoon
```diff
@@ -0,0 +1,606 @@
...
+::  +peer-messages: subscribe to subset of messages and updates
+::
+::
+++  peer-primary
+  |=  wir=wire
+  ^-  (quip move _this)
+  ~&  (lent (prey:pubsub:userlib /primary bol))
+  =*  messages  messages.str.sta
+  =/  lisunitmov=(list (unit move))
+    %+  turn  ~(tap by messages)
```
joemfb: `+turn` over a list producing `(unit)` and filter out the `~` items is in the stdlib as `+murn`:

```
> =((murn "foo" some) (turn "foo" same))
%.y
```

app/chat.hoon
```diff
@@ -0,0 +1,606 @@
...
+        ?:  add.rum
+          (~(put in circles.str.sta) cir.rum)
+        (~(del in circles.str.sta) cir.rum)
+      =/  str  %=  str.sta
+          circles  cis
+          peers
+            ?:  add.rum
+              (~(put by peers.str.sta) [our.bol cir.rum] ~)
+            (~(del by peers.str.sta) [our.bol cir.rum])
+      ==
```
joemfb: This `=/` and its inner `%=` makes confusing use of whitespace, please indent/backstep idiomatically.

app/chat.hoon
```diff
@@ -0,0 +1,606 @@
...
+    ::  %gram:
+    ::
+        %gram
+    ?>  ?=(%gram -.sto)
+    =*  messages  messages.str.sta
+    =/  circle/circle:hall  [`@p`(slav %p &2:wir) &3:wir]
+    =/  nes/(list envelope:hall)  (~(got by messages) circle)
+    =/  str  %=  str.sta
+      messages  (~(put by messages) circle (snoc nes nev.sto))
+    ==
```
joemfb: Here too

app/chat.hoon
```diff
@@ -0,0 +1,606 @@
...
+      %circle
+      =/  shp/@p  (slav %p &2:wir)
+      =/  pat  /circle/[&3:wir]/config/group
+      ?:  =(&3:wir 'inbox')
+        :_  this
+        [ost.bol %peer wir [shp %hall] pat]~
+      ?:  (~(has in src.inbox.str.sta) [[shp &3:wir] ~])
+        :_  this
+        [ost.bol %peer wir [shp %hall] pat]~
+      [~ this]
```
joemfb: This entire expression (after `%circle`) should be de-dented.

app/chat.hoon
```diff
@@ -0,0 +1,606 @@
...
```
joemfb: This one too

app/clock.hoon
```diff
@@ -0,0 +1,78 @@
...
+  |=  [wir=wire success=? binding=binding:http-server]
+  ^-  (quip move _this)
+  [~ this]
+::
+++  prep
+  |=  old=(unit ~)
+  ^-  (quip move _this)
+  =/  lismov/(list move)  %+  weld
+    `(list move)`[ost.bol %connect / [~ /'~clock'] %clock]~
+    `(list move)`[ost.bol %poke /clock [our.bol %launch] [%noun [%clock /tile '/~clock/js/tile.js']]]~
```
joemfb: 80 chars

app/clock.hoon
```diff
@@ -0,0 +1,78 @@
...
+::
+++  bound
+  |=  [wir=wire success=? binding=binding:http-server]
+  ^-  (quip move _this)
+  [~ this]
+::
+++  prep
+  |=  old=(unit ~)
+  ^-  (quip move _this)
+  =/  lismov/(list move)  %+  weld
```
joemfb: `%+  weld` should start a new line, with appropriate backstep thereafter.

app/clock.hoon
```diff
@@ -0,0 +1,78 @@
...
+  %+  turn  (prey:pubsub:userlib /tile bol)
+  |=  [=bone ^]
+  [bone %diff %json jon]
+::
+++  poke-handle-http-request
+  %-  (require-authorization:app ost.bol move this)
+  |=  =inbound-request:http-server
+  ^-  (quip move _this)
+  =+  request-line=(parse-request-line url.request.inbound-request)
+  =+  back-path=(flop site.request-line)
```
joemfb: Prefer `=/  face  value` over `=+  face=value`

app/hall.hoon
```diff
@@ -2535,14 +2535,16 @@
...
-        (slag (sub count.s 100) grams.s)
-      [shape.s mirrors.s]
-    [locals.s remotes.s]
+    ::?.  ?=($channel sec.con.shape.s)  ~
+    :::+  ~  n
+    ::::  share no more than the last 100, for performance reasons.
+    :::+  ?:  (lte count.s 100)  grams.s
+    ::    (slag (sub count.s 100) grams.s)
+    ::  [shape.s mirrors.s]
+    ::[locals.s remotes.s]
```
joemfb: Worth discussing further (elsewhere). /cc @Fang-

app/launch.hoon
```diff
@@ -0,0 +1,149 @@
...
+  =/  data/(unit [json url=@t])  (~(get by data.sta) name)
+  ?~  data
+    [~ this]
+  :-
+  %+  turn  (prey:pubsub:userlib /main bol)
+    |=  [=bone *]
+    [bone %diff %json (frond:enjs:format name jon)]
+  %=  this
+    data.sta  (~(put by data.sta) name [jon url.u.data])
+  ==
```
joemfb: Unusual whitespace usage from `:-`

app/launch.hoon
```diff
@@ -0,0 +1,149 @@
...
+  :_  this
+  [ost.bol %diff %json data]~
+::
+++  generate-script-marl
+  |=  data=tile-data
+  ^-  marl
+  %+  turn  ~(tap by data)
+    |=  [key=@tas [jon=json url=@t]]
+    ^-  manx
+    ;script@"{(trip url)}";
```
joemfb: This gate should be de-dented.

app/timer.hoon
```diff
@@ -0,0 +1,124 @@
...
+  |=  [wir=wire success=? binding=binding:http-server]
+  ^-  (quip move _this)
+  [~ this]
+::
+++  prep
+  |=  old=(unit tim=@da)
+  ^-  (quip move _this)
+  =/  lismov/(list move)  %+  weld
+    `(list move)`[ost.bol %connect / [~ /'~timer'] %timer]~
+    `(list move)`[ost.bol %poke /timer [our.bol %launch] [%noun [%timer /tile '/~timer/js/tile.js']]]~
```
joemfb: 80 chars

app/timer.hoon
```diff
@@ -0,0 +1,124 @@
...
+  |=  jon=json
+  ^-  (quip move _this)
+  ?.  ?=(%s -.jon)
+    [~ this]
+  =/  str/@t  +.jon
+  ?:  =(str 'start')
+    =/  data/@da  (add now.bol ~s10)
+    :-  %+  weld
+      `(list move)`(send-tile-diff [%s (scot %da data)])
+      `(list move)`[ost.bol %wait /timer data]~
```
joemfb: Backstep. Also, unless the order is important, just cons the `%wait` onto the head of the list produced by `+send-tile-diff`.

sys/vane/rver.hoon
```diff
@@ -60,7 +60,7 @@
       $:  %b
           ::
           ::
-          $%  [%wake ~]
+          $%  [%wake error=(unit tang)]
```
joemfb: `+take` needs to actually handle this error case. See https://github.com/urbit/arvo/pull/1192#discussion_r293957948

joemfb (approved):
I've noted some style issues -- since we're integrating branches there's unlikely to be another PR which presents an opportunity to do so. I haven't marked all the instances of the non-idiomatic patterns identified below. These can be discussed further if there isn't time to resolve before integrating.

It's also worth noting that this PR includes/subsumes #1192, and therefore inherits an issue therein (see https://github.com/urbit/arvo/pull/1192#discussion_r293957948).

Approving now in the interest of integration. We can circle back on these issues if there isn't time to resolve them now.

/cc @jtobin

ixv:
cc-ing @loganallenc

joemfb:
@ixv or @loganallenc, please resolve the conflicts in this PR

joemfb:
Also, there are some minor differences between this PR (f86d76d56cd269dc187559aabc644f8989c4292e) and #1192 (479b2189b176090a0944e1fb6f27a9ae6ad50d7a). Please confirm that #1192 can be closed, or merge it into this PR if it cannot.

joemfb:
Bump.

app/timer.hoon
loganallenc: fixed

app/timer.hoon
loganallenc: fixed

app/chat.hoon
loganallenc: fixed

app/chat.hoon
loganallenc: fixed

app/chat.hoon
loganallenc: fixed

app/chat.hoon
loganallenc: fixed

app/chat.hoon
loganallenc: fixed

app/clock.hoon
loganallenc: fixed

app/clock.hoon
loganallenc: fixed

app/clock.hoon
loganallenc: fixed

app/launch.hoon
loganallenc: fixed. I hate this convention.

app/launch.hoon
loganallenc: fixed.

sys/vane/rver.hoon
```diff
@@ -1985,6 +1997,9 @@
         ~|([%bad-channel-wire wire] !!)
     ::
         %timeout
+      ?>  ?=([%b %wake *] sign)
+      =?  moves  ?=(^ error.sign)
+        [[duct %d %flog %crud %wake error.sign] moves]
```
ixv: Handled this case as per @eglaysher's guidance. Does this seem correct to you @joemfb

app/chat.hoon
loganallenc: fixed.

sys/vane/rver.hoon
joemfb: No, this isn't going to be sufficient. The semantics of getting a `%wake` with an error is that you already got a `%wake` without the error, and something you did crashed. You can't just continue doing the same thing you would normally do (and in fact already did), because it'll just crash again.

The bare minimum is to simply print the error and do nothing else. The objectively correct thing to do is to release/undo any leases/commitments associated with that timer (cancel builds, requests, clean up state, etc.), but I'm not sure what that specifically means in this case.

joemfb:
This branch still has a conflict with `lighter-than-eyre` (in `%gall`), please address that.

sys/vane/rver.hoon
eglaysher: The only thing this handler does is undo leases. If we crashed during this handling, we crashed while trying to free the timer resources on timeout.

So I guess I was wrong, and you should just print, @ixv ?

joemfb: That looks right to me.

joemfb: To be clear, I think @eglaysher is correct and you should only print, not continue with the evaluation in the error case.

joemfb (approved):
LGTM

ixv:
Thanks Joe, sorry for the long delay.

PR #1203, opened 2019-06-27
Alef (unused)

belisarius222:
New Ames as-is, with passing unit tests but not yet hooked up to anything.  Making the PR so this unfinished work makes it into the monorepo.


sys/vane/alef.hoon
```diff
@@ -0,0 +1,2588 @@
...
+::    Ames passes a %memo to itself to trigger a heartbeat message to
+::    our sponsor.
+::
+::    Ames passes a %private-keys to Jael to request our private keys.
+::    Ames passes a %public-keys to Jael to request a peer's public
+::    keys.
+::
++$  note
+  $%  $:  %a
+      $%  [%memo sponsor=ship message=_[/a/ping ~]]
```
joemfb: I think these `+note` types are unnecessarily over-constrained (some with `+sign` below). The interface ames exposes generalizes without any change to every vane in the system. I'd be inclined to prefix `%memo` here (and `%done` in `+sign`) with just `@tas`.

sys/vane/alef.hoon
```diff
@@ -0,0 +1,2588 @@
...
+    |=  [=message-num =fragment-num]
+    ^+  packet-pump
+    ::
+    =-  ::  if no sent packet matches the ack, don't apply mutations or effects
+        ::
+        ?.  found.-
+          packet-pump
+        ::
+        =.  metrics.state  metrics.-
+        =.  live
```
joemfb: Any particular reason not to use `+sign:as:crub` and `+sure:as:crub` here?

joemfb: Barring some other anticipated need, I'm be inclined to use these bits to support more protocol-versions for safer breach cycling.

sys/vane/alef.hoon
```diff
@@ -0,0 +1,2588 @@
+::  protocol-version: current version of the ames wire protocol
+::
+=/  protocol-version=?(%0 %1 %2 %3 %4 %5 %6 %7)  %0
+::
+|%
++|  %generics
+::  $mk-item: constructor for +ordered-map item type
+::
++*  mk-item  [key val]  [key=key val=val]
```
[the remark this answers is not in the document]
belisarius222: Nope, doesn't work. I agree, I'll make a guthub issue for that.

sys/vane/alef.hoon
```diff
@@ -0,0 +1,2588 @@
...
++$  address        @uxaddress
++$  blob           @uxblob
++$  bone           @udbone
++$  fragment       @uwfragment
++$  fragment-num   @udfragmentnum
++$  message-num    @udmessagenum
++$  private-key    @uwprivatekey
++$  public-key     @uwpublickey
++$  signature      @uwsignature
++$  symmetric-key  @uwsymmetrickey
```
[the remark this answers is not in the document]
belisarius222: Couldn't agree more

sys/vane/alef.hoon
```diff
@@ -0,0 +1,2588 @@
...
+::    %send: packet to unix
+::    %done: notify vane that peer (n)acked our message
+::
+::    %mass: memory usage report
+::    %turf: domain report, relayed from jael
+::
++$  gift
+  $%  [%memo =message]
+      [%send =lane =blob]
+      [%done error=(unit error)]
```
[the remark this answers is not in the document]
belisarius222: I wrote it as %aver first. Then I had a long discussion with @philipcmonk, and we both found it hard to say and hard to map onto the right concept.

So I switched it to %done, which seems fine so far.

sys/vane/alef.hoon
belisarius222: Yeah, I think you're right

sys/vane/alef.hoon
belisarius222: I wanted more manual control while I was designing this. In particular, the order in which we validate life, public key, and signature could affect DDoS susceptibility. So I decided to do these things more manually until they settle.

I wouldn't be surprised if we go back to +sign and +sure after thinking a bit harder.

belisarius222: Which bits? GitHub is pointing this comment at a seemingly unrelated piece of code

joemfb: My bad -- this is about the unused packet header bits.

sys/vane/alef.hoon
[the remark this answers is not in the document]
belisarius222: Hm, basic usage works in dojo. Maybe this isn't broken after all ... I'll loop back around on this in Alef once I get some basic integration stuff compiling.

sys/vane/alef.hoon
belisarius222: Oh, yeah. We could reserve the bits for extra protocol versions, but I think it might be better to just leave them officially unused until we start to need more protocol versions. If we do need some extra bits for other purposes between now and then, this approach avoids potential conflict between using a bit for versioning or something else.

joemfb:
/me apologizes for the noise. extra merge commits never hurt anybody, right?

PR #1205, opened 2019-06-27
Jael rewrite

philipcmonk:
For monorepo migration.  Compiles, but has no tests and has never been run.  Won't break the build because it's added as a new vane "Kale".


joemfb (approved):
LGTM

PR #1207, opened 2019-06-28
Correct indentation of ++render-atom (whitespace-only change)

jfranklin9000:


jtobin:
We're actually in the process of deprecating this repo in favour of
monorepo-ing urbit/urbit, but I've applied the patch for you over there
(on the v0.8.0.rc branch).

